PR #755, opened 2025-04-08
Fetch from all installations when none is selected

gusevda:
### What does this PR do?

In this PR, the behavior of the installation selectors was changed. Now, if no installations are selected, resources will be fetched from all installations.

Towards https://github.com/giantswarm/roadmap/issues/3956.


marians (approved):
Great!

PR #782, opened 2025-04-17
Display aggregated status in the deployments table

gusevda:
### What does this PR do?

This PR changes how deployment statuses are being displayed in the deployments table. Statuses are being aggregated into three groups - `Successfull` (reconciled, deployed), `Pending` (reconciling, pending-install, pending-upgrade, pending-rollback, uninstalling) and `Failed` (all other statuses).

### How does it look like?

Screenshot 2025-04-17 at 15 02 46
Screenshot 2025-04-17 at 15 04 03

Towards https://github.com/giantswarm/roadmap/issues/3975.

### DIFF
--- a/plugins/gs/src/components/deployments/DeploymentsDataProvider/utils.ts
+++ b/plugins/gs/src/components/deployments/DeploymentsDataProvider/utils.ts
@@ -44,7 +44,7 @@ export type DeploymentData = {
 };
 
 function getStatus(deploymentStatus: string) {
-  const successfullStatuses = ['reconciled', 'deployed'];
+  const successfulStatuses = ['reconciled', 'deployed'];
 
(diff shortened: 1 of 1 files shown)


plugins/gs/src/components/deployments/DeploymentsDataProvider/utils.ts
```diff
@@ -43,6 +43,28 @@ export type DeploymentData = {
   app?: string;
 };
 
+function getStatus(deploymentStatus: string) {
+  const successfullStatuses = ['reconciled', 'deployed'];
```
marians, line 47: Since "successful" is written with one l only, I think this should be named `successfulStatuses`

marians (approved):
Works like advertised!

PR #815, opened 2025-04-28
Delegate unimplemented Kubernetes client methods

gusevda:
### What does this PR do?

Our GS plugins leverage a custom Kubernetes client that enables direct frontend connections to Kubernetes clusters. While this client implements the standard Kubernetes API interface (making it interchangeable with the default Kubernetes backend client), it previously only implemented the proxy method.

This limitation caused the standard Kubernetes plugin to fail when used alongside our plugins. With this PR, we've enhanced the custom Kubernetes client to delegate all methods except proxy to the standard backend client. This improvement ensures compatibility with the standard Kubernetes plugin while maintaining our custom functionality.

### How does it look like?
Before:
Screenshot 2025-04-29 at 11 31 48

After:
Screenshot 2025-04-29 at 11 30 13


marians (approved):
Awesome!

PR #862, opened 2025-06-02
Hide installations selector when only one installation is connected

gusevda:
### What does this PR do?

When there is only one installation available, we don't need to show installations selector component. In this PR, the state with only one installation is being handled.

### How does it look like?

Only one installation connected:
Screenshot 2025-06-02 at 10 26 58
Screenshot 2025-06-02 at 10 25 50

More than one installation connected:
Screenshot 2025-06-02 at 10 27 44
Screenshot 2025-06-02 at 10 27 26

Towards https://github.com/giantswarm/roadmap/issues/3814.

### DIFF
--- a/.changeset/huge-women-spend.md
+++ b/.changeset/huge-women-spend.md
@@ -1,0 +1,5 @@
+---
+'@giantswarm/backstage-plugin-gs': minor
+---
+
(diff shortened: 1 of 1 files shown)


marians (approved):
👍

PR #1489, opened 2026-04-07
Support multiple values sources in app-deployment template

gusevda:
### What does this PR do?

Adds multi-source value editing to the scaffolder deployment workflow and a set of related improvements to the edit flow.

**Core feature:**
- **`GSValueSourcesEditor`** — new scaffolder field extension with a dynamic list of `valuesFrom` sources (ConfigMaps/Secrets), per-source YAML editors, name validation, drag-to-reorder, and merged Helm schema validation
- **`MultiSourceDeploymentPicker`** — new DeploymentPicker variant that fetches HelmRelease `valuesFrom` entries and resolves their ConfigMap/Secret data for edit mode
- `valuesMode` is passed via URL formData so edit templates can switch between single- and multi-source flows

**Other changes:**
- Custom review step with YAML preview for scaffolder templates
- Warn when an app deployment is managed through GitOps (read-only notice in the edit flow)
- Disable Edit Deployment button when the deployment structure is incompatible with the edit template (non-OCIRepository source, mismatched names/namespaces, digest pinning)
- `fromJson` Nunjucks filter in the scaffolder backend for deserializing JSON strings in templates
- `height` / `maxHeight` ui:options for `GSValueSourcesEditor`, `GSYamlValuesEditor` and `GSSecretYamlValuesEditor`;
- Fix YAML block-style autocompletion in CodeMirror editors

### How does it look like?

Multiple value sources:
screencapture-localhost-3000-create-templates-default-edit-app-deployment-v2-2026-04-14-10_17_55

Custom "Review" step:
screencapture-localhost-3000-create-templates-default-app-deployment-v2-2026-04-10-12_02_52

Towards https://github.com/giantswarm/roadmap/issues/4274, https://github.com/giantswarm/roadmap/issues/4132.


marians:
Some comments

## 1. The UI per ConfigMap/Secret is getting a little heavy

image

This is a lot of UI per item. I'd like to make this appear lighter, require less space, and draw less attention. Can we reduce the font size for this line?

Do we really need customization for the data key? I think we can remove it. This should be a rare case where we want the user to edit this in the resulting manifest, not in the UI directly.

Our trash can icon isn't nice. Let's try `DeleteOutline` instead of `Delete`. That's not as heavy.

## 2. I'd like to replace the modality of "Value source references (ConfigMap/Secret)" vs. "Inline values"

image

Instead, we could add an option "Inline values" to the type drop-down with ConfigMap and Secret. This option should only be applicable to the last item in the list, as the inline values always override all values from CMs and Secrets.

## 3. Removing an item should require confirmation

Since there is no undo, users should confirm before removing a section. At least if there is any content in the section.

## 4. Info line placement and spacing

The line `Add ConfigMap or Secret value sources. Order determines merge priority (last wins).` is very close to the upper container. There should be more spacing. Or, even better, we might place it next to the add button.

Also we can increase the spacing between that button and the validation results.

marians (approved):

PR #1522, opened 2026-04-20
fix(ai-chat-backend): make tool-using chat actually return answers

teemow:
## Summary

The AI chat plugin currently produces an empty reply (the spinner just
disappears) whenever the model decides to invoke a tool. Two independent
problems are involved; both are fixed here.

### 1. `ai` v6 caps the agent at one step

Since the upgrade to `ai` v6 (#1160), `streamText` defaults to
`stopWhen: stepCountIs(1)`. As a result, every chat turn where the
model decides to invoke a tool terminates **immediately after the first
tool call** -- before the model is ever invited to consume the tool
result and produce a user-facing answer.

To the user this looks like the chat spinner just disappears: no error
is logged anywhere (`onError` is never called, the stream completes
cleanly with `finishReason: "tool-calls"`), and the assistant message
ends up empty.

Set `stopWhen: stepCountIs(20)` so the agent loop runs to completion,
and expose the cap as a new optional `aiChat.maxSteps` config
(default `20`, matching the cap that `ai`'s built-in `Agent` class
uses internally) so operators can tune it per deployment.

### 2. Some OpenAI-compatible servers can't speak the Responses API

Once the agent loop is allowed to continue, the SDK posts the
`function_call_output` items back to whatever endpoint the OpenAI
provider points at. `@ai-sdk/openai` v2 defaults to the new Responses
API (`/v1/responses`).

That's fine for real OpenAI, but OpenAI-compatible servers haven't
all caught up. In particular, vLLM crashes the API server with
`KeyError: 'role'` while parsing the tool-result message, the AI SDK
retries 3 times, and finally surfaces
`"Failed after 3 attempts. Last error: 'role'"` to the chat client.

Add an optional `aiChat.openai.api` config to choose the endpoint:

  - `"responses"` (default): keeps the current `@ai-sdk/openai`
    behavior of talking to `/v1/responses`. Recommended for real
    OpenAI.
  - `"chat"`: pins the model client to `openai.chat(modelName)`,
    talking to `/v1/chat/completions`. Use this for vLLM and any
    other OpenAI-compatible server that doesn't yet implement the
    Responses API correctly.

Azure and Anthropic paths are unchanged.

## Reproduction

In an environment that has `aiChat.mcp` configured (so the model has
tools to call), send any prompt that triggers a tool call (for
example `"List the available kubernetes clusters"` against a muster
MCP server). The HTTP response stream from `POST /api/ai-chat/chat`
ends with:

```
data: {"type":"tool-output-available", ...}
data: {"type":"finish-step"}
data: {"type":"finish","finishReason":"tool-calls"}
data: [DONE]
```

i.e. no `text-start` / `text-delta` events ever follow the tool
output, so the UI shows nothing.

After this change, the same prompt continues into a second step, the
model receives the tool result, and emits the expected text reply.
On vLLM-backed deployments the operator additionally needs to set
`aiChat.openai.api: chat` in `app-config.yaml` to side-step the
Responses-API bug.

## Test plan

- [ ] CI green
- [ ] Manual: in an environment with at least one MCP server
      configured, ask a question that requires a tool call and
      confirm the assistant produces a text reply (not just an empty
      bubble).
- [ ] Manual: with `aiChat.maxSteps` set explicitly in
      `app-config.yaml`, confirm the configured cap is honoured.
- [ ] Manual (vLLM/BWI): with `aiChat.openai.api: chat` set, confirm
      multi-step tool-using chats complete end to end. Confirm that
      the default (`responses`) still works against real OpenAI.

## Follow-up

- BWI's Backstage `app-config.yaml` should set `aiChat.openai.api: chat`
  once a release containing this PR is rolled out, so the workaround
  becomes part of the GitOps config instead of a live patch.
- Upstream vLLM bug (Responses API `KeyError: 'role'` on
  `function_call_output` items) should be filed and tracked
  separately.


marians:
Please also verify this PR with the following test: Ask a question like this (after having asked/instructed something else before in the conversation):

> How is my current context usage? Use the getCurrentContext tool.

I suspect that the current `stopWhen: stepCountIs(1)` implementation has been a too general fix for the problem that the model always tried to add more text after this tool call, which was very confusing.

gusevda (commented):
I think `.github/workflows/zz_generated.pre-commit.yaml` and `.pre-commit-config.yaml` should not be edited here since the changes will be overwritten by the align files PRs.

teemow:
Rebased onto current main and addressed the review feedback:

@gusevda Reverted the changes to `.github/workflows/zz_generated.pre-commit.yaml` and `.pre-commit-config.yaml` — those will land via the next align-files PR, as you suggested. The remaining pre-commit cleanup commits now only touch local files (`certificate.yaml.example` PEM placeholders, the trailing-newline fixes, the `.claude` skill script executable bits, and the regenerated `helm/backstage/values.schema.json`). Updated their commit messages so they reflect the reduced scope and call out that the workflow/config bits are deliberately deferred.

@marians Verified your scenario against a Backstage pod running this branch's behavior (multi-step + `aiChat.openai.api: chat`). After warming the conversation up with an unrelated turn, asking *"How is my current context usage? Use the getContextUsage tool."* produces:

```
Events: {"start-step":2,"text-delta":63,"tool-input-available":1,"tool-output-available":1,"finish":1}
finishReason: stop

Your current context usage is:
- Input tokens: 3,722
- Output tokens: 52
- Model: bwi-active

The context window is still well within limits. Let me know if you'd like to continue with any specific questions or tasks!
```

So in this run the model takes exactly 2 steps (one `getContextUsage` tool call + one text reply that quotes the actual numbers from the tool), with `finishReason: "stop"`. No invented stats, no extra tool calls, no chatty hallucination after the tool — i.e. the regression you were worried about (the model padding nonsense onto a tool result) does not reproduce here. If you can think of a more aggressive prompt that used to misbehave under the old `stopWhen(1)` workaround, happy to retry.

marians (approved):
tested, works for me.

the suggested test looks like this for me:

Image

gusevda (approved):
LGTM

PR #1535, opened 2026-04-21
feat(ai-chat): make welcome screen configurable

marians:
### What does this PR do?

Adds three optional config keys — `aiChat.welcome.title`, `aiChat.welcome.subtitle`, and `aiChat.welcome.suggestions` — that let operators customize the welcome screen shown before the first message in the AI chat. Each key independently falls back to its built-in default, so there is no behavior change unless a key is explicitly set. Up to 3 suggestions are still picked at random per mount.

Operators can tailor the greeting and suggested questions to their portal's audience (translated copy, domain-specific prompts, org-specific examples). End users see no change unless the operator overrides one or more keys.

### How does it look like?

No visual change by default.

Example config:

```yaml
aiChat:
  welcome:
    title: 'How can I help you today?'
    subtitle: 'Not sure what to try? Pick one of the suggestions below.'
    suggestions:
      - 'Which clusters exist in the platform?'
      - 'What can I do here?'
      - 'Which MCP tools are available?'
      - 'Tell me a joke about DevOps engineers'
```

With the above, title, subtitle, and cards reflect the configured values.

Extends the pattern established by the recently-landed `aiChat.systemPrompt` override, continuing to move hardcoded AI-chat strings into app config.

No — the commented block added to `app-config.yaml` documents the shape, and `plugins/ai-chat/config.d.ts` is the schema source of truth.

### DIFF
--- a/.changeset/calm-rainbow-nestle.md
+++ b/.changeset/calm-rainbow-nestle.md
@@ -1,0 +1,5 @@
+---
+'@giantswarm/backstage-plugin-ai-chat': patch
+---
+
+Add optional `aiChat.welcome.{title,subtitle,suggestions}` config to customize the AI chat welcome screen. Falls back to the built-in defaults when not set.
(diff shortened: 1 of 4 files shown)


gusevda (approved):
LGTM

PR #1579, opened 2026-04-27
Make sidebar logo height configurable

gusevda:
### What does this PR do?

Adds a new optional config value `app.branding.logo.height` that overrides the height (in pixels) of the custom branding logo image rendered in the sidebar's `LogoFull` and `LogoIcon` components.

The inline SVG fallbacks (Giant Swarm wordmark/icon) keep their existing hardcoded heights — the override applies only when a custom branding asset (`logo-full.svg/png`, `logo-icon.svg/png`) is present.

Downstream deployments that ship custom branding assets via the branding module can now resize the sidebar logo without forking the components. With no config value set, behavior is identical to before.

Example:

```yaml
app:
  branding:
    logo:
      height: 48
```

### DIFF
--- a/.changeset/zesty-maple-shine.md
+++ b/.changeset/zesty-maple-shine.md
@@ -1,0 +1,5 @@
+---
+'app': patch
+---
+
+Make sidebar logo image height configurable via `app.branding.logo.height`.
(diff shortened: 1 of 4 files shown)


marians (approved):
Very good!

PR #1608, opened 2026-04-30
Fix AI chat input freeze on dead-key composition

gusevda:
### What does this PR do?

Fixes the AI chat composer freezing after typing a dead-key character (most reproducibly the backtick on a German keyboard, but also affects U.S. International, French AZERTY, Spanish ISO, etc.).

Upstream `@assistant-ui/react`'s `ComposerPrimitive.Input` tracks IME composition with a ref and short-circuits `onChange` while composing. On dead-key layouts Chrome occasionally drops the matching `compositionend`, the ref stays stuck at `true`, and every subsequent keystroke is silently ignored — the input still has focus but no longer accepts text.

This PR replaces the primitive with a controlled MUI `TextField` bound directly to the assistant runtime via `useAssistantApi()` / `useAssistantState()`. Composition is only consulted at the `Enter`→send keybinding (a per-event check on `e.nativeEvent.isComposing`), so CJK users keep correct submit behavior and there is no longer any stateful flag that can get stuck. `setText` is wrapped in `flushResourcesSync` so the controlled input stays in sync with fast keystrokes.

Upstream issue - https://github.com/assistant-ui/assistant-ui/issues/3923

### DIFF
--- a/.changeset/gentle-apple-travel.md
+++ b/.changeset/gentle-apple-travel.md
@@ -1,0 +1,5 @@
+---
+'@giantswarm/backstage-plugin-ai-chat': patch
+---
+
+Fix AI chat input freezing after typing dead keys (e.g., backtick on German keyboard) by replacing the upstream `ComposerPrimitive.Input` with a controlled `TextField` bound directly to the assistant runtime.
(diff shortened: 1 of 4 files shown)


marians (approved):
Tested, works for me

PR #1617, opened 2026-05-04
Update ai to v6.0.174 and bump CI buildjob to large resource class

marians:
## Summary

Re-applies the `ai` v6.0.174 lockfile bump from #1615, which was closed after CI failed. The original PR's typecheck step OOMed at 4 GB; raising `--max-old-space-size` to 8192 alone wasn't enough because the default `medium` Docker executor only has 4 GB RAM, so the kernel killed Node before V8 could use the bigger heap.

- Cherry-pick `Update dependency ai to v6.0.174` from #1615
- Set `resource_class: large` (8 GB / 4 vCPU) on `buildjob`
- Raise `NODE_OPTIONS --max-old-space-size` to 8192

## Test plan

- [ ] CircleCI `buildjob` typecheck step passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ workflows:
 jobs:
   buildjob:
     executor: architect/architect
-    resource_class: large
+    resource_class: xlarge
     working_directory: ~/backstage
     docker:
       - image: cimg/node:24.15.0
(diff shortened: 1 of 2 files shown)


marians:
Closing — the underlying issue is a type regression introduced in `ai@6.0.171`, not a CI memory shortage. tsc reproducibly OOMs locally with any version in (6.0.170, 6.0.174]. Staying on 6.0.170 until upstream ships a fix.

PR #1644, opened 2026-05-09
chart: make replicas, livenessProbe, and readinessProbe configurable

teemow:
The chart hardcoded `replicas: 1` and the liveness/readiness probe parameters in the Deployment template, with no values knob to override them.

That worked when Backstage installs were small and a 10s `initialDelaySeconds` liveness was generous, but plugin-heavy installs (BWI portal, etc.) take 60-120s to register all plugins on a cold start and get killed mid-init by the default liveness. Operators working around this had no choice but to hand-patch the live Deployment after every chart upgrade -- which is reverted on the next HelmRelease reconcile.

## Changes

- Add `replicas` (integer, default 1). Same default as before; only safe to bump when backing the install with an RWX volume or external Postgres database. Documented on the value.
- Add `readinessProbe` (object, default `{}`). Merged over the chart's default readiness probe via `mergeOverwrite`, so callers can override individual fields (e.g. just `initialDelaySeconds`) without restating the entire probe shape.
- Add `livenessProbe` (object, default `{}`). Same merge semantics.

Defaults are unchanged so existing installs render identically.

## Verification

- `helm lint helm/backstage` clean
- `helm template ... --set replicas=2 --set livenessProbe.initialDelaySeconds=120 --set livenessProbe.failureThreshold=10 --set readinessProbe.initialDelaySeconds=60` renders the expected Deployment with `replicas: 2`, the chart's default `httpGet` probe paths/ports preserved, and the overridden timing fields applied.

## Why now

The BWI bundle (`giantswarm/bwi` `bwi-backstage`) ships an SQLite-PVC-backed Backstage and reconciles via Flux. Every bundle release reverts hand-patched probe overrides on the live Deployment, and Backstage's plugin-heavy cold start exceeds the default 10s liveness, leaving the new pod in `CrashLoopBackOff`. With these knobs the bundle can set `livenessProbe.initialDelaySeconds: 120, failureThreshold: 10` and `readinessProbe.initialDelaySeconds: 60, failureThreshold: 20` once and have it survive reconciliation.


marians (approved):
Looks good to me

teemow:
Superseded on main. Both knobs this PR introduced exist in the chart under a different shape and are released:

- Probe timing: #1658 (v0.130.0) added `probes.liveness.{initialDelaySeconds,periodSeconds,timeoutSeconds,failureThreshold}` and the same four fields under `probes.readiness`, and raised `timeoutSeconds` to 5. The probe path and port stay fixed by design, so a full `Probe` object override was not taken.
- Replicas: #2105 (v0.188.0) added `replicas` (integer, minimum 1, default 1).

Rebased onto main this PR resolves to nothing. `replicas` is already templated, and the schema on main rejects the keys proposed here as additional properties:

```
$ helm template backstage helm/backstage --set livenessProbe.initialDelaySeconds=120
Error: values don't meet the specifications of the schema(s) in the following chart(s):
- at '': additional properties 'livenessProbe' not allowed
```

while the shape on main renders the intended Deployment:

```
$ helm template backstage helm/backstage \
    --set probes.liveness.initialDelaySeconds=120 --set probes.liveness.failureThreshold=10 \
    --set probes.readiness.initialDelaySeconds=60 --set probes.readiness.failureThreshold=20
```

gives `livenessProbe: {httpGet: {path: /.backstage/health/v1/liveness, port: 7007}, initialDelaySeconds: 120, periodSeconds: 15, timeoutSeconds: 5, failureThreshold: 10}` and the readiness counterpart with 60/10/5/20. The defaults on main render byte-identical to the probes live on every current installation (liveness 10/15/5/3, readiness 5/10/5/3; `helm lint` clean).

Closing without merge. The values for slow cold starts follow in the next comment.

teemow:
Proposed values for an install whose cold start exceeds the default 10 s liveness delay. Not applied anywhere; the numbers are the ones from the description above, mapped to the chart's shape on main:

```yaml
probes:
  liveness:
    initialDelaySeconds: 120
    failureThreshold: 10
  readiness:
    initialDelaySeconds: 60
    failureThreshold: 20
```

With the default `periodSeconds` (15 and 10) this tolerates a startup of up to 120 + 150 s before the first liveness kill and marks the pod Ready as soon as the readiness endpoint answers after 60 s. `timeoutSeconds` stays at the chart default of 5.

Where the block belongs:

- The BWI bundle, `giantswarm/bwi`, `backstage/release.yaml`, under the `backstage` HelmRelease's `spec.values`. The bundle follows every chart version (`semver: x.x.x`), so the knobs are available to it today; it currently sets none.
- Giant Swarm's own portals: `giantswarm/giantswarm-management-clusters`, `management-clusters/<installation>/extras/backstage/backstage/user-values.yaml` (the `user-values-backstage` ConfigMap). The same path in an installation's `<customer>-management-clusters` repo for a customer portal.

No fleet portal needs it today: every Backstage pod on the installations that run the chart (2.56.6 and 0.245.0) was rolled on 2026-09-22 and came up at the default 10 s liveness and 5 s readiness delay with zero restarts and no `Liveness probe failed` events. Add the block only to an install that shows those events during startup.

PR #1659, opened 2026-05-11
Add PagerDuty integration

gusevda:
### What does this PR do?

Adds PagerDuty integration to Backstage with three main capabilities:

- **"Who is on call" entity card** — displays current on-call responders for services with PagerDuty annotations
- **Catalog processor** — auto-annotates catalog entities with PagerDuty service/integration IDs based on entity metadata
- **MCP action** — resolves PagerDuty IDs from catalog entities, enabling agentic workflows

Users can now see who is on call for a service directly on its catalog entity page, and entities are automatically linked to their PagerDuty services without manual annotation. The PagerDuty backend plugin is optional and only loads when configured.

### How does it look like?

Service entity:
Screenshot 2026-05-11 at 15 40 35

Team entity:
Screenshot 2026-05-11 at 15 40 07

Towards https://github.com/giantswarm/giantswarm/issues/34903.


marians (approved):
Nice!

PR #1664, opened 2026-05-12
Add LatestReleaseProcessor for GitHub release annotations

gusevda:
### What does this PR do?

Adds a new `LatestReleaseProcessor` in `catalog-backend-module-gs`. For any `Component` entity carrying a `github.com/project-slug` annotation, the processor looks up the latest GitHub release and writes two annotations:

- `giantswarm.io/latest-release-tag`
- `giantswarm.io/latest-release-date`

Two modes:

- **Default**: queries GitHub's canonical `/repos/{owner}/{repo}/releases/latest` (newest non-draft, non-prerelease).
- **Monorepo / prefix mode**: when the entity also carries `giantswarm.io/release-tag-prefix` (e.g. `sre/`), the processor pages through `/releases` and picks the first non-draft release whose `tag_name` starts with that prefix. Prereleases are included in this mode (matches the existing monorepo-level annotation behavior).

Results are cached in-memory keyed by `(owner/repo, prefix)` with a configurable TTL. Concurrent cache fills for the same key are deduped via an inflight map so the first refresh after a TTL expiry doesn't fan out to multiple GitHub calls.

Disabled by default. Enable via:

```yaml
catalog:
  processors:
    latestRelease:
      enabled: true
      cacheTtlSeconds: 3600  # optional, defaults to 1h
```

### DIFF
--- a/.changeset/dark-grape-persist.md
+++ b/.changeset/dark-grape-persist.md
@@ -1,0 +1,5 @@
+---
+'@giantswarm/backstage-plugin-catalog-backend-module-gs': minor
+---
+
+Add `LatestReleaseProcessor` that annotates Component entities carrying a `github.com/project-slug` annotation with `giantswarm.io/latest-release-tag` and `giantswarm.io/latest-release-date` by querying GitHub Releases. Defaults to `/releases/latest`, or matches by tag prefix when `giantswarm.io/release-tag-prefix` is set (for monorepos with prefixed per-subproject tags). Results are cached in-memory per `(owner/repo, prefix)` with a configurable TTL. Gated behind `catalog.processors.latestRelease.enabled`.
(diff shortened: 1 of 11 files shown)


marians (commented):
I'm in agreement with the move in general, but this I don't understand:

> Results are cached in-memory keyed by (owner/repo, prefix) with a configurable TTL. Concurrent cache fills for the same key are deduped via an inflight map so the first refresh after a TTL expiry doesn't fan out to multiple GitHub calls.

Why not use the database for this?

gusevda:
> I'm in agreement with the move in general, but this I don't understand:
> 
> > Results are cached in-memory keyed by (owner/repo, prefix) with a configurable TTL. Concurrent cache fills for the same key are deduped via an inflight map so the first refresh after a TTL expiry doesn't fan out to multiple GitHub calls.
> 
> Why not use the database for this?

An in-memory map is a more lightweight implementation. We do the same in the PagerDutyAnnotations processor. We do use the database for caching in the sbomDependencies processor, mainly because we fetch/parse more data there and it won't fit nicely in an in-memory map. I thought we don't need the extra complexity here.

marians (approved):

PR #1680, opened 2026-05-20
Fix muster MCP connection by dropping custom session-aware transport

teemow:
## Problem

The AI chat backend fails to connect to muster MCP servers with the runtime error:

```
Cannot set property protocolVersion of #<StreamableHTTPClientTransport> which has only a getter
```

Backstage logs show `Failed to connect to MCP server 'muster' ...` instead of `Successfully loaded tools from MCP server: muster`, so the AI assistant cannot call any muster tools.

## Root cause

The custom `createSessionAwareTransport` was introduced in #1365 (Mar 2026) specifically because muster used a non-standard session header `X-Muster-Session-ID`. Muster has since switched to the spec-standard `Mcp-Session-Id` header (which is what every current deployment config sets), and `@ai-sdk/mcp`'s built-in `type: 'http'` transport already handles `mcp-session-id` round-tripping case-insensitively.

The custom code is now dead weight, and it is also what triggers the `protocolVersion` runtime error against `@modelcontextprotocol/sdk@1.29.0`.

## Fix

Drop the custom transport entirely and use `@ai-sdk/mcp`'s built-in HTTP transport with `{ type: 'http', url, headers }`. Removes 200+ LoC of glue + tests, removes a direct dependency on `@modelcontextprotocol/sdk`, and fixes the runtime error.

## Changes

- Delete `plugins/ai-chat-backend/src/createSessionAwareTransport.ts` and its test.
- In `getMcpTools.ts`, drop the `createSessionAwareTransport` import, the `sessionHeader` field on `McpServerConfig` and its four call-sites, and replace the conditional transport block with a plain `{ type: 'http', url, headers }` config.
- Remove the `sessionHeader` field from `config.d.ts`. Deployments still setting it will see a config-validation warning, which flags them for the follow-up cleanup PRs.
- Drop the direct `@modelcontextprotocol/sdk` dependency from `plugins/ai-chat-backend/package.json` (it remains as a transitive dep of `@backstage/plugin-mcp-actions-backend`).
- Add a changeset.

## Follow-up

Three deployment configs still set `sessionHeader: Mcp-Session-Id` and need a follow-up PR per repo once the image with the new `config.d.ts` is released:

- `giantswarm-management-clusters` (gazelle + graveler)
- `workload-clusters-fleet` (demotech)

## Validation

- `yarn tsc` passes.
- `yarn workspace @giantswarm/backstage-plugin-ai-chat-backend build` passes.
- After the chart rolls out to gazelle, ask the AI Assistant on `https://devportal.giantswarm.io` to list its MCP tools and to trigger one (e.g. "check the pods in gazelle"). Logs should show `Successfully loaded tools from MCP server: muster` and the chat should successfully invoke a muster tool instead of falling back to "muster is unavailable".


marians:
Tested, worked for me

PR #1697, opened 2026-06-01
chore(ai-chat): add Opus 4.8 context window size

marians:
This PR adds the context window info for Claude Opus 4.8

PR #1698, opened 2026-06-01
feat(ai-chat): model-aware Anthropic thinking config

marians:
### What does this PR do?

Makes the AI chat backend's Anthropic thinking configuration **model-aware**.

Previously `router.ts` hardcoded `thinking: { type: 'enabled', budgetTokens: 10000 }` for every Anthropic model. The adaptive-thinking generation (Opus 4.5+, Sonnet 4.6) rejects that shape with a 400:

> "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.

The provider options are now built once at router setup based on the model ID:

- **Adaptive-thinking models** (Opus 4.5/4.6/4.7/4.8, Sonnet 4.6) → `{ thinking: { type: 'adaptive' }, effort }`, where `effort` is configurable via the new `aiChat.anthropic.effort` option (default `high`).
- **Older Claude models** (Sonnet 4.5, Haiku 4.5, Opus 4.1, Claude 3.x) → unchanged legacy `{ thinking: { type: 'enabled', budgetTokens: 10000 } }`.
- **Non-Claude models** → unchanged (`undefined`).

Adaptive-thinking models also reject `temperature`/`top_p`/`top_k`, so any of those configured under `aiChat.sampling` are dropped (with a warning) for those models to prevent the next 400.

The branching logic lives in a new pure helper (`utils/anthropicProviderOptions.ts`) with unit tests. The same options object is reused for both the `streamText` call and the debug-meta header (removing the duplicated literal).

This PR also folds in the Opus 4.8 context-window addition from #1697 (which can be closed once this lands).

AI chat works when `aiChat.model` is set to Claude Opus 4.8 (or any Opus 4.5+/Sonnet 4.6) instead of failing on every message. Thinking depth on those models is tunable via `aiChat.anthropic.effort` (`low | medium | high | xhigh | max`).

### How does it look like?

No visual change. Behavior fix — chat messages stream normally on Opus 4.8 rather than erroring out.

The `enabled + budgetTokens` thinking interface was removed on the newer Claude generation; they require `thinking.type=adaptive` + `output_config.effort` (exposed by `@ai-sdk/anthropic` as a top-level `effort` field). The prefix-list approach mirrors the manually-maintained model map already in `ContextUsageDisplay.tsx`.

No (the new `aiChat.anthropic.effort` option is documented inline in `config.d.ts`).

### DIFF
--- a/.changeset/nice-loops-allow.md
+++ b/.changeset/nice-loops-allow.md
@@ -1,0 +1,5 @@
+---
+'@giantswarm/backstage-plugin-ai-chat': patch
+---
+
+add Opus 4.8 context window info
--- a/.changeset/quick-pugs-think.md
+++ b/.changeset/quick-pugs-think.md
@@ -1,0 +1,5 @@
+---
+'@giantswarm/backstage-plugin-ai-chat-backend': minor
+---
+
+AI chat: make the Anthropic thinking config model-aware. Adaptive-thinking Claude models (Opus 4.5+, Sonnet 4.6) now use `thinking: { type: 'adaptive' }` plus `effort` (configurable via `aiChat.anthropic.effort`, default `high`) instead of the legacy `thinking: { type: 'enabled', budgetTokens }` shape, which those models reject with a 400 ("thinking.type.enabled is not supported for this model"). Older Claude models keep the legacy budget-based interface, and non-Claude models are unaffected. For adaptive-thinking models, `temperature`/`topP`/`topK` from `aiChat.sampling` are dropped (with a warning) since those models also reject them.
--- a/plugins/ai-chat-backend/config.d.ts
+++ b/plugins/ai-chat-backend/config.d.ts
@@ -12,6 +12,13 @@ export interface Config {
        * @visibility backend
        */
       baseUrl?: string;
+      /**
+       * Thinking effort for adaptive-thinking Claude models (Opus 4.5+,
+       * Sonnet 4.6). Maps to output_config.effort. Ignored by older Claude
+       * models, which use a fixed thinking budget instead. Defaults to 'high'.
+       * @visibility backend
+       */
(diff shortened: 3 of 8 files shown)


marians:
### Suggestion: reuse `DEFAULT_ANTHROPIC_EFFORT` for the router default

Nice change — verified the shapes against `@ai-sdk/anthropic@3.0.81` (the `effort` enum `low|medium|high|xhigh|max` matches `config.d.ts` exactly) and the new tests pass.

One small DRY nit. The default effort `'high'` is now defined in two places:

- `anthropicProviderOptions.ts` exports `DEFAULT_ANTHROPIC_EFFORT = 'high'`
- `router.ts` hardcodes the same literal:

```ts
const anthropicEffort =
  config.getOptionalString('aiChat.anthropic.effort') ?? 'high';
```

Since the router always passes a non-`undefined` `effort` into `buildAnthropicProviderOptions`, the helper's own default is effectively dead in production (only exercised by tests) — the *real* runtime default is the literal in `router.ts`. That means the exported constant and the live default could silently diverge.

Suggest importing and reusing the constant so there's a single source of truth:

```ts
import { DEFAULT_ANTHROPIC_EFFORT } from './utils/anthropicProviderOptions'; // or via ./utils barrel

const anthropicEffort =
  config.getOptionalString('aiChat.anthropic.effort') ?? DEFAULT_ANTHROPIC_EFFORT;
```

Not a blocker — happy to push this if you'd like.

marians:
Good call — fixed in c5535853. The router now falls back to `DEFAULT_ANTHROPIC_EFFORT` (exported via the `./utils` barrel) instead of the duplicated `'high'` literal, so there's a single source of truth and the constant can't silently diverge from the live default. tsc/tests/lint green.

PR #1709, opened 2026-06-03
fix(app): only report untracked page views for registered app routes

marians:
### What does this PR do?

Stops bot/scanner probes from creating "Untracked page view" noise issues in Sentry.

The public Backstage URLs attract bots probing for known CMS/admin paths (`/wp-login.php`, `/phpmyadmin`, …). Every such navigation produced an "Untracked page view" warning in Sentry, even though the path lands on the "Not Found" page.

The fix gates the error report on the navigate event's analytics context: the app's `RouteTracker` resolves each navigation against the registered routes and stores the owning plugin/extension in the context (`pluginId`/`extensionId`). When nothing matches, the context keeps its default values (`'app'`/`'app'`) — those navigations are no longer reported.

Untracked page views on **real** app routes (pages missing from `getTelemetryPageViewPayload`) are still reported, preserving the signal introduced in #1249. TelemetryDeck pageview signals are unchanged.

Fixes https://github.com/giantswarm/giantswarm/issues/36756 (source-side fix, option 1). The already-accumulated Sentry issues will be muted separately as per the issue's cleanup section.

No

### DIFF
--- a/packages/app/src/apis/analytics/TelemetryDeckAnalyticsApi.test.ts
+++ b/packages/app/src/apis/analytics/TelemetryDeckAnalyticsApi.test.ts
@@ -1,0 +1,129 @@
+import {
+  AnalyticsEvent,
+  ConfigApi,
+  IdentityApi,
+} from '@backstage/core-plugin-api';
+import { TelemetryDeckAnalyticsApi } from './TelemetryDeckAnalyticsApi';
+
+const mockSignal = jest.fn();
+
+jest.mock('@telemetrydeck/sdk', () => ({
+  __esModule: true,
+  default: jest.fn().mockImplementation(() => ({ signal: mockSignal })),
+}));
+
+const flushPromises = () => new Promise(resolve => setTimeout(resolve, 0));
+
+/**
+ * Builds a 'navigate' event the way the app's RouteTracker emits it: with
+ * the owning plugin/extension in the context for matched routes, and the
+ * default context values ('app'/'app') for paths that did not match any
+ * registered route.
+ */
+function navigateEvent(
+  subject: string,
+  context: Record<string, string>,
+): AnalyticsEvent {
+  return {
+    action: 'navigate',
+    subject,
+    context: context as AnalyticsEvent['context'],
+  };
+}
+
+describe('TelemetryDeckAnalyticsApi', () => {
+  const configApi = {
+    getOptionalConfig: jest.fn().mockReturnValue(undefined),
+  } as unknown as ConfigApi;
+
+  const identityApi = {
+    getBackstageIdentity: jest
+      .fn()
+      .mockResolvedValue({ userEntityRef: 'user:default/test' }),
+    getProfileInfo: jest.fn().mockResolvedValue({}),
+  } as unknown as IdentityApi;
+
+  const errorReporterApi = { notify: jest.fn() };
+
+  function createApi() {
+    return TelemetryDeckAnalyticsApi.fromConfig({
+      configApi,
+      identityApi,
+      errorReporterApi,
+    });
+  }
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('reports untracked page views for paths that matched a registered route', () => {
+    createApi().captureEvent(
+      navigateEvent('/notifications', {
+        pluginId: 'notifications',
+        extensionId: 'page:notifications',
+      }),
+    );
+
+    expect(errorReporterApi.notify).toHaveBeenCalledWith(
+      'Untracked page view: /notifications',
+      {
+        level: 'warning',
+        type: 'untracked_page_view',
+        path: '/notifications',
+      },
+    );
+  });
+
+  it('does not report untracked page views for paths that did not match any route', () => {
+    createApi().captureEvent(
+      navigateEvent('/wp-login.php', {
+        pluginId: 'app',
+        extensionId: 'app',
+      }),
+    );
+
+    expect(errorReporterApi.notify).not.toHaveBeenCalled();
+  });
+
+  it('does not report tracked page views', () => {
+    createApi().captureEvent(
+      navigateEvent('/clusters', {
+        pluginId: 'gs',
+        extensionId: 'page:gs/clusters',
+      }),
+    );
+
+    expect(errorReporterApi.notify).not.toHaveBeenCalled();
+  });
+
+  it('still sends a pageview signal for unmatched paths', async () => {
+    createApi().captureEvent(
+      navigateEvent('/wp-login.php', {
+        pluginId: 'app',
+        extensionId: 'app',
+      }),
+    );
+    await flushPromises();
+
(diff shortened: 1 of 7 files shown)


packages/app/src/apis/analytics/TelemetryDeckAnalyticsApi.ts
```diff
@@ -81,6 +81,20 @@ export class TelemetryDeckAnalyticsApi implements AnalyticsApi {
...
+   *
+   * The app's RouteTracker resolves each navigation against the registered
+   * routes and stores the owning plugin/extension in the analytics context.
+   * When nothing matches (e.g. internet bots probing for /wp-login.php and
+   * similar paths on the public URL), the context keeps its default values
+   * of pluginId 'app' and extensionId 'app'.
+   */
+  private static isRegisteredRoute(event: AnalyticsEvent): boolean {
+    const { pluginId, extensionId } = event.context;
+    return pluginId !== 'app' || extensionId !== 'app';
```
marians: **The fix appears to be a no-op: unmatched routes don't actually carry `extensionId: 'app'` — they carry `extensionId: 'app/root'`.**

The `{pluginId: 'app', extensionId: 'app'}` values are only the fallback of `useAnalyticsContext()` when there is **no** `AnalyticsContext` provider in the tree. At runtime there always is one above the `RouteTracker`: `AppRoot`'s factory wraps `AppRouter` in an `ExtensionBoundary` ([AppRoot.tsx](https://github.com/backstage/backstage/blob/master/packages/frontend-app-api/src/extensions/AppRoot.tsx)), and `ExtensionBoundary` provides `AnalyticsContext` with `{extensionId: node.spec.id, pluginId}` — i.e. `{extensionId: 'app/root', pluginId: 'app'}`. For unmatched routes the `RouteTracker` contributes empty attributes, so the navigate event inherits those boundary values.

Verified empirically against the installed `@backstage/plugin-app` with a `renderInTestApp` probe and an `analyticsApiRef` spy:

```jsonc
// initialRouteEntries: ['/wp-login.php'], no matching route
{
  "action": "navigate",
  "subject": "/wp-login.php",
  "context": { "pluginId": "app", "extensionId": "app/root" }
}
// matched route for contrast:
{
  "action": "navigate",
  "subject": "/my-test-page",
  "context": { "pluginId": "test", "extensionId": "test-route:test//my-test-page" }
}
```

So for a bot probe, `pluginId !== 'app' || extensionId !== 'app'` → `false || true` → `true`, and the Sentry report still fires. The doc comment above (lines 84–91) documents the same incorrect assumption.

```suggestion
    return pluginId !== 'app' || extensionId !== 'app/root';
```

Two caveats worth addressing beyond the one-liner:

1. `'app/root'` is also an implementation detail (albeit a public attachment-point id), so it would be good to pin the real contract with an integration test using `renderInTestApp` from `@backstage/frontend-test-utils` + an `analyticsApiRef` spy, instead of hand-built contexts (see comment on the test file).
2. Alternatively, a deeper home for this suppression could be the error-reporter layer (e.g. filtering `untracked_page_view` events in `SentryErrorReporter`/`beforeSend` for paths that resolve to the Not Found page), which wouldn't depend on analytics-context default values at all.

packages/app/src/apis/analytics/TelemetryDeckAnalyticsApi.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        level: 'warning',
+        type: 'untracked_page_view',
+        path: '/notifications',
+      },
+    );
+  });
+
+  it('does not report untracked page views for paths that did not match any route', () => {
+    createApi().captureEvent(
+      navigateEvent('/wp-login.php', {
```
marians: These fixtures encode a context shape the real `RouteTracker` never emits: at runtime an unmatched route's navigate event carries `{pluginId: 'app', extensionId: 'app/root'}` (inherited from the `AppRoot` `ExtensionBoundary`'s `AnalyticsContext`), not `{pluginId: 'app', extensionId: 'app'}`. So this test passes while the production path takes the opposite branch and still reports bot probes (see comment on `isRegisteredRoute`).

Since the whole fix hinges on what the framework puts into the context, consider replacing the hand-built contexts with an integration test that exercises the real `RouteTracker`, e.g.:

```tsx
import { renderInTestApp } from '@backstage/frontend-test-utils';
import { analyticsApiRef } from '@backstage/frontend-plugin-api';

const events: AnalyticsEvent[] = [];
renderInTestApp(<div />, {
  initialRouteEntries: ['/wp-login.php'],
  apis: [[analyticsApiRef, { captureEvent: e => events.push(e) }]],
});
// feed the captured events into TelemetryDeckAnalyticsApi and assert on errorReporterApi.notify
```

That pins the actual framework contract and will catch upstream changes to the default context values on Backstage upgrades.

packages/app/src/apis/analytics/TelemetryDeckAnalyticsApi.ts
marians: Fixed in 645cf444 — good catch, the comparison was indeed a no-op.

- The check now compares against `{pluginId: 'app', extensionId: 'app/root'}` and the doc comment describes the real mechanism (inherited from the `AppRoot` `ExtensionBoundary`, not the bare `useAnalyticsContext()` fallback).
- **Caveat 1 (pinning the contract):** done — the test file now uses `renderInTestApp` with the real `RouteTracker` (see the other thread). I verified the new test fails when the old `'app'` comparison is reinstated.
- **Caveat 2 (error-reporter layer instead):** I kept the suppression in the analytics layer: the route-match information only exists in the analytics context of the navigate event, while `SentryErrorReporter` would have to re-derive it from the path (route table not available there) or pattern-match the message, which is the same `ignoreErrors` blunt instrument that would also drop reports for real-but-untracked routes.

Additionally, the API now implements `AnalyticsApi`/`AnalyticsEvent` from `@backstage/frontend-plugin-api` (and is registered via the NFS `analyticsApiRef` in `AppOverrides.tsx`), since the old `core-plugin-api` context type declares `routeRef`/`extension` fields that never arrive under the new frontend system.

packages/app/src/apis/analytics/TelemetryDeckAnalyticsApi.test.ts
marians: Done in 645cf444 — replaced the hand-built contexts with an integration test, going one step further than the sketch: instead of capturing events with a spy and re-feeding them, the test installs the real `TelemetryDeckAnalyticsApi` as the app's `analyticsApiRef` via `renderInTestApp(..., { apis: [[analyticsApiRef, api]] })`. The real `RouteTracker` then emits navigate events straight into the API under test, and assertions run against `errorReporterApi.notify` / the TelemetryDeck signal.

Matched routes are produced with `mountedRoutes`, unmatched ones by just navigating to `/wp-login.php`. I verified the suppression test fails against the previous (broken) comparison, so it genuinely pins the framework contract and will catch upstream changes to the boundary context values.

PR #1712, opened 2026-06-03
feat: migrate pages to the new frontend system page header

marians:
### What does this PR do?

Brings the **Catalog, Clusters, Deployments, Installations, and Flux** pages to the page header rendered automatically by the new frontend system (`PluginHeader` from `@backstage/ui` via the `PageLayout` swappable component) — the same header already used on `/settings`, `/create`, and `/search`. Previously these pages set `noHeader: true` on their page blueprints and rendered the classic `Header`/`PageWithHeader` from `@backstage/core-components` themselves.

Details:

- Removed `noHeader: true` from the page blueprints and stripped the manual `Page`/`Header` wrappers from the page components. The header now comes from each blueprint's existing `title`/`icon` params.
- The Catalog index page gets a static **Catalog** title and the `FolderIcon` used by the sidebar nav item. Portals can override the title via app-config (`page:catalog` → `config.title`), like the existing `page:scaffolder` config.
- The Flux **List view / Tree view** become `SubPageBlueprint` extensions rendered as tabs in the page header (same pattern as `/settings`). `FluxPageLayout` was removed from `flux-react`, and `FluxListFilterBlueprint`/`FluxTreeFilterBlueprint` now attach to `sub-page:flux/list` / `sub-page:flux/tree` (input `filters`).
- The home page (`/`) intentionally stays headerless (landing-page design), `/ai-chat` already renders the new header manually, and catalog entity pages use the upstream entity-header system — all unchanged.

- Consistent page headers across the portal: every main page now shows the slim new-style header with icon and title instead of the old gradient header.
- The Clusters/Deployments subtitles are gone (the new header has no subtitle support), as are the Support buttons on the Catalog and Installations pages.
- The Catalog page title changes from "{org name} Catalog" to "Catalog".
- `/flux` now redirects to `/flux/list`; `/flux/tree` is unchanged. The view switcher tabs moved into the page header.

The app is fully on the new frontend system; these pages were the remaining opt-outs via `noHeader: true`. The migration follows the upstream patterns (catalog/scaffolder alpha pages and `SubPageBlueprint` tabs).

No.


packages/app/src/modules/catalog/IndexPage.tsx
```diff
@@ -1,11 +1,13 @@
...
 import catalogPlugin from '@backstage/plugin-catalog/alpha';
+import FolderIcon from '@material-ui/icons/Folder';
 
-// Catalog index page — render GSCustomCatalogPage directly (full page component)
+// Catalog index page — render GSCustomCatalogPage with the NFS page header
 export const IndexPage = PageBlueprint.makeWithOverrides({
   factory(originalFactory) {
     return originalFactory({
-      noHeader: true,
+      title: 'Catalog',
```
marians, line 9: **Behavior change: catalog title no longer includes the organization name.**

The old `CustomCatalogPage` rendered `` `${orgName} Catalog` `` from `organization.name` config (which is set in `app-config.yaml`, and presumably per customer portal). This hardcodes the title to plain `Catalog`.

If that's intentional, consider mentioning it in the changeset alongside the subtitle note. Otherwise, note that `PageBlueprint`'s config schema allows overriding the title per installation via app-config (`page:catalog/index` → `config.title`), or the title could keep the org name here.

plugins/flux/src/components/FluxResourcesListPage/FluxResourcesListPage.tsx
```diff
@@ -0,0 +1,14 @@
+import { type ReactNode } from 'react';
+import { Content } from '@backstage/core-components';
+import { FluxResourcesListView } from '@giantswarm/backstage-plugin-flux-react';
+import { QueryClientProvider } from '../QueryClientProvider';
+
+export function FluxResourcesListPage({ filters }: { filters: ReactNode }) {
+  return (
+    <QueryClientProvider>
```
marians, line 8: **Query cache is no longer shared between the list and tree views.**

The old `FluxPage` mounted one `QueryClientProvider` above both tabs, so switching views kept the same live `QueryClient`. Now each sub-page mounts its own provider, and `QueryClientProvider` creates a new `QueryClient` per mount — every tab switch discards the live client and re-hydrates from the localStorage persister (async restore + refetch churn for all 8 flux resource types).

The persister softens this, but a shared client would avoid it entirely — e.g. hoist the `QueryClient` to module level inside `QueryClientProvider`, or have both sub-pages render a single shared wrapper (which would also deduplicate `FluxResourcesListPage`/`FluxResourcesTreePage`, which are otherwise identical).

plugins/gs/src/components/catalog/CustomCatalogPage/CustomCatalogPage.tsx
```diff
@@ -53,37 +45,30 @@ export function CustomCatalogPage(props: CustomCatalogPageProps) {
...
-                actions={actions}
-                tableOptions={tableOptions}
-                emptyContent={emptyContent}
-              />
-            </CatalogFilterLayout.Content>
-          </CatalogFilterLayout>
-        </EntityListProvider>
-      </Content>
-    </PageWithHeader>
+    <Content>
```
marians, line 50: **`SupportButton` removed silently (here and on the Installations page).**

The old `ContentHeader` carried a `SupportButton` ("All your software catalog entities" / "Installations"). This PR drops it entirely rather than relocating it, and the changeset doesn't mention it. If dropping it is intentional (the new header has no slot for it), a line in the changeset would help; otherwise it could move into the page content or a `PluginHeaderActionBlueprint` action.

packages/app/src/modules/catalog/IndexPage.tsx
marians: The title change is intentional (matches the upstream NFS catalog page default). Documented it in the changeset now, including the per-installation override path via app-config: `app.extensions` → `page:catalog` → `config.title`. (Note the extension id is `page:catalog` — the override in `IndexPage.tsx` has no `name`, so it replaces the default catalog page extension.) → c551ab86

plugins/flux/src/components/FluxResourcesListPage/FluxResourcesListPage.tsx
marians: Good catch. Fixed by hoisting the `QueryClient` (and persister) to module scope in `plugins/flux/src/components/QueryClientProvider.tsx`, so the list and tree sub-pages share one live client — tab switches no longer discard the cache or re-hydrate from localStorage. This also keeps the cache warm when navigating away from `/flux` and back, which the old per-mount client didn't. Kept `FluxResourcesListPage`/`FluxResourcesTreePage` as separate components since they're now trivial 15-line wrappers and the shared state lives in the provider. → c551ab86

plugins/gs/src/components/catalog/CustomCatalogPage/CustomCatalogPage.tsx
marians: Dropping it is intentional — the new header has no slot for it, and upstream's own NFS pages (`/create`, `/search`, settings) ship without a SupportButton. Added a line to the changeset documenting the removal on both pages. If we want it back later, `PluginHeaderActionBlueprint` is the right hook. → c551ab86

PR #1733, opened 2026-06-10
feat(muster): add workflow visualization plugins

teemow:
## Summary

- New frontend plugin `plugins/muster` (`@giantswarm/backstage-plugin-muster`, pluginId `muster`): renders muster workflow definitions as flow diagrams built on `@xyflow/react`, with an AI-Elements-style node look (header/content/footer) implemented in MUI v4 JSS.
  - `/muster`: workflows list (name, description, availability)
  - `/muster/workflows/:name`: vertical flow canvas (input node + one node per step, dashed side edges for `condition.from_step` dependencies), execution history panel, and a step detail drawer (tool, templated args, condition, resolved input, result/error, timing)
  - Execution overlay: per-step status colors (completed/failed/skipped/in-progress/pending), animated edge into the running step, condition outcome badge on skipped steps; polls every 3s while the selected execution is in progress
- New backend plugin `plugins/muster-backend` (`@giantswarm/backstage-plugin-muster-backend`, pluginId `muster`): thin REST proxy over the muster MCP server's `core_workflow_list/get` and `core_workflow_execution_list/get` tools via `@ai-sdk/mcp`, with a TTL + close-detection cached MCP client (same self-healing approach as ai-chat-backend).
  - Connection config reuses the existing `aiChat.mcp` entry named `muster` (overridable via `muster.serverName`), so the muster endpoint is configured once
- Wiring: app features + nav item, backend registration, commented config example in `app-config.yaml`, changeset

## Test plan

- [ ] CI: tsc, lint, prettier, tests (includes new unit tests for the workflow->graph transform, backend routes, and config/result parsing)
- [ ] Manual: point `aiChat.mcp` entry `muster` at a local muster bridge (`http://localhost:8091/mcp`), open `/muster`, select a workflow, run an execution and watch the live overlay


plugins/muster-backend/src/MusterMcpClient.ts
```diff
@@ -0,0 +1,258 @@
...
+export function readMusterServerFromConfig(
+  config: Config,
+  logger: LoggerService,
+): MusterServerConfig | undefined {
+  const serverName = config.getOptionalString('muster.serverName') ?? 'muster';
+
+  const mcpConfigs = config.getOptionalConfigArray('aiChat.mcp');
+  const mcpConfig = mcpConfigs?.find(
+    mcp => mcp.getString('name') === serverName,
+  );
```
marians, line 58: **Bug: `getString('name')` throws if any `aiChat.mcp` entry lacks a `name`, crashing plugin registration**

`Config.getString()` throws when the key is missing. Since `find()` runs this callback over *every* entry in `aiChat.mcp` — including ones that have nothing to do with muster — a single entry without a `name` field throws before the match is found. That exception propagates out of `readMusterServerFromConfig` → `createRouter` → plugin init, taking down the whole backend plugin instead of degrading to "unconfigured".

```ts
const mcpConfig = mcpConfigs?.find(
  mcp => mcp.getOptionalString('name') === serverName,
);
```

plugins/muster-backend/src/MusterMcpClient.ts
```diff
@@ -0,0 +1,258 @@
...
+
+    const tool = tools[toolName];
+    if (!tool) {
+      throw new NotFoundError(`Unknown muster tool: ${toolName}`);
+    }
+
+    let result;
+    try {
+      result = await tool.execute!(args, {
+        toolCallId: `muster-backend-${toolName}`,
```
marians, line 140: **Bug: `tool.execute!` non-null assertion produces a raw `TypeError` if `execute` is undefined**

`toolsFromDefinitions` returns a `Record<string, Tool>` where `Tool.execute` is typed optional. If the `@ai-sdk/mcp` version returns a definition-only tool with no executor, `tool.execute!` satisfies the compiler but throws `Cannot read properties of undefined (reading 'call')` at runtime. That bare `TypeError` bypasses the structured-error path and surfaces as a 500 with a stack trace rather than a clean error.

```ts
if (!tool.execute) {
  throw new NotFoundError(`Muster tool ${toolName} has no executor`);
}
result = await tool.execute(args, { ... });
```

plugins/muster/src/components/WorkflowDetailPage/WorkflowDetailPage.tsx
```diff
@@ -0,0 +1,146 @@
...
+    [],
+  );
+
+  const selectedStepNode = nodes.find(
+    (node): node is Node<StepNodeData> =>
+      node.data.kind === 'step' && node.id === selectedStepId,
+  );
+
+  if (workflowQuery.error) {
+    return (
```
marians, line 100: **Bug: `executionQuery`/`executionsQuery` errors are silently swallowed**

Only `workflowQuery.error` renders a `ResponseErrorPanel`. If the execution-list fetch fails, `ExecutionHistoryPanel` just gets `executions ?? []` and looks empty — as if the workflow has no runs. If the *selected* execution fetch fails (e.g. 503), `execution` stays `undefined`, the canvas renders the bare definition with no overlay, and the user gets no signal that anything broke. The `QueryClient` even disables retries for `ServiceUnavailableError`/`NotFoundError`, so these surface immediately and then vanish. Consider rendering an error state (or at least an inline alert) for both secondary queries.

marians (commented):
Code review at high effort — 9 findings (6 bugs, 1 altitude, 2 cleanup).

Dropped after verification: the `getClient()` concurrency race (no `await` before `this.entry =`, so it's serialized), the `EXECUTION_STATUSES` skipped omission (executions are never `skipped` per the type doc — only steps are), and the static `toolCallId` (AI SDK ignores it today).

plugins/muster-backend/src/router.ts
```diff
@@ -0,0 +1,154 @@
...
+  config: Config;
+  /** Overridable for tests. */
+  client?: MusterMcpClient;
+}
+
+function parseOptionalInt(value: unknown, name: string): number | undefined {
+  if (value === undefined) {
+    return undefined;
+  }
+  const parsed = Number(value);
```
marians, line 21: **Bug: `parseOptionalInt` silently coerces an empty value to `0`**

`Number('')` is `0`, which passes `Number.isInteger(0) && 0 >= 0`. So `GET /executions?limit=` (present but empty) returns `limit = 0` instead of being rejected — a zero-row page rather than a 400. `parseInt(String(value), 10)` returns `NaN` for `''` and would be caught by the existing guard.

```ts
const parsed = parseInt(String(value), 10);
if (!Number.isInteger(parsed) || parsed < 0) {
  throw new InputError(`${name} must be a non-negative integer`);
}
```

plugins/muster-backend/src/router.ts
```diff
@@ -0,0 +1,154 @@
...
+      name: req.params.name,
+    });
+    res.json(result);
+  });
+
+  router.get('/executions', async (req, res) => {
+    const { workflow_name: workflowName, status } = req.query;
+
+    if (
+      status !== undefined &&
```
marians, line 80: **Minor bug: a repeated `status` query param yields a misleading 400**

`req.query.status` becomes an array when the param is repeated (`?status=completed&status=failed`). The `EXECUTION_STATUSES.includes(status as ...)` cast hides this from the compiler, but at runtime `includes(array)` is `false`, so the handler throws `status must be one of: ...` even though each value is individually valid. Worth normalizing the param (take the first element, or reject arrays with a clearer message).

plugins/muster/src/lib/workflowToGraph.ts
```diff
@@ -0,0 +1,155 @@
...
+  steps.forEach((step, index) => {
+    const executionStep = executionSteps.get(step.id);
+    let status: StepNodeStatus | undefined;
+    if (execution) {
+      status = executionStep?.status ?? 'pending';
+    }
+
+    nodes.push({
+      id: step.id,
+      type: 'workflowStep',
```
marians, line 116: **Minor bug: duplicate step `id`s collide in the React Flow graph**

Node IDs and `seq-*` edge IDs are derived directly from `step.id`. If a workflow defines two steps with the same `id`, xyflow dedupes them and one node/edge silently disappears, breaking the layout. If muster already guarantees unique step IDs this is moot — otherwise consider keying on the array index (`${step.id}-${index}`) or validating uniqueness.

plugins/muster-backend/src/router.ts
```diff
@@ -0,0 +1,154 @@
...
+  const router = Router();
+  router.use(express.json());
+
+  router.get('/health', (_, res) => {
+    res.json({ status: 'ok', configured: Boolean(client) });
+  });
+
+  const requireClient = (): MusterMcpClient => {
+    if (!client) {
+      const error = new Error(
```
marians, line 55: **Altitude: use `ServiceUnavailableError` from `@backstage/errors` instead of a plain `Error` + a hand-rolled error handler**

This throws `new Error(...)` and mutates `error.name = 'ServiceUnavailableError'`, then adds ~20 lines of custom middleware to map that name (and `InputError`) to status codes. `@backstage/errors` exports `ServiceUnavailableError` (and `InputError` is already used here) — both carry the right `statusCode` and are mapped to 503/400 by Backstage's default error middleware. Throwing the real classes lets you delete the custom middleware entirely and keeps the error body shape consistent with every other plugin.

```ts
import { ServiceUnavailableError } from '@backstage/errors';
// ...
throw new ServiceUnavailableError('No muster MCP server is configured. ...');
```

plugins/muster/src/components/WorkflowDetailPage/ExecutionHistoryPanel.tsx
```diff
@@ -0,0 +1,99 @@
...
+  header: {
+    padding: theme.spacing(1.5, 2),
+    borderBottom: `1px solid ${theme.palette.divider}`,
+  },
+  listIcon: {
+    minWidth: 32,
+  },
+}));
+
+function formatDuration(durationMs: number): string {
```
marians, line 32: **Cleanup: `formatDuration` is copy-pasted identically in three files**

The same two-line `formatDuration(durationMs)` appears here, in `StepDetailDrawer.tsx`, and in `flow/WorkflowStepNode.tsx`. Extract it once (e.g. `plugins/muster/src/lib/formatDuration.ts`, or `ui-react` alongside the other formatters) so the format can't drift across the three surfaces.

plugins/muster-backend/src/MusterMcpClient.ts
```diff
@@ -0,0 +1,258 @@
...
+const CLIENT_TTL_MS = 30 * 60 * 1000; // 30 minutes
+
+// Matches @ai-sdk/mcp's MCPClientError thrown from request() once the
+// transport closed. Same self-healing signal as ai-chat-backend's
+// McpClientCache (see plugins/ai-chat-backend/src/McpClientCache.ts).
+const CLOSED_CLIENT_ERROR_FRAGMENT =
+  'Attempted to send a request from a closed client';
+
+export function isClosedClientError(error: unknown): boolean {
+  if (!error) return false;
```
marians, line 30: **Reuse: the cached-MCP-client machinery duplicates `ai-chat-backend`'s `McpClientCache`**

`isClosedClientError` (same constant + body), the TTL + `onclose`-hook cache entry, and `readMusterServerFromConfig`'s header parsing are all re-implementations of code already in `plugins/ai-chat-backend/src/McpClientCache.ts` / `getMcpTools.ts`. The comments acknowledge the overlap, but the two copies will drift: e.g. if the AI SDK changes the closed-client message string, a fix to ai-chat-backend won't self-heal muster, and any new `aiChat.mcp` config field won't be picked up here. Consider extracting the shared cache + config reader into a common backend module both plugins import.

Issue #1829, opened 2026-06-28
Optimize the CircleCI buildjob: add Yarn cache, parallelize, adopt generated Node job


teemow:
## Problem

`.circleci/custom.yml`'s `buildjob` runs `yarn install --immutable` **cold on every run** — it caches only `dist-types/tsconfig.tsbuildinfo`, never the Yarn Berry cache (`.yarn/cache`). It then runs install → tsc → `lint:all` → `prettier:check` → `test:all` → backend build **serially** in one `large` job, and uses the full `lint:all`/`test:all` rather than the change-scoped variants (`lint` already exists as `backstage-cli repo lint --since origin/main`). This is the largest Node codebase in the fleet (~2.8k TS files) and the most expensive CI gate, with the same class of missing-cache problem that giantswarm/giantswarm#36941 / architect-orb#838 fixed for Go.

Part of the Node/TypeScript CI workstream giantswarm/giantswarm#37023, under epic giantswarm/giantswarm#36729.

## Proposed solution

Short term (independent, do now):
- `restore_cache`/`save_cache` on `.yarn/cache` keyed on `yarn.lock` (write-once-per-lockfile), the Node analogue of the `$GOCACHE` persistence in architect-orb#838.
- Use change-scoped selection on PR builds: `backstage-cli repo lint/test --since origin/main` instead of the `*:all` variants; keep full `*:all` on `main`/tag builds.
- Split typecheck / lint+prettier / test into parallel jobs feeding the build.

Medium term:
- Replace the hand-written `buildjob` with the generated Node job from the `giantswarm/devctl` enabler issue, keeping `gen.ci.image.preBuildJob` so the `packages/backend/Dockerfile` image build still overlays the persisted `packages/*/dist/*` workspace.

## Acceptance criteria

- [ ] Yarn cache persisted and restored across runs (warm-cache install is a no-op).
- [ ] PR builds use change-scoped lint/test; `main`/tag builds remain full.
- [ ] Measurable wall-clock reduction on the build pipeline recorded on the issue.
- [ ] (Medium term) `buildjob` replaced by the generated Node job; image preBuildJob handoff preserved.

Refs: epic giantswarm/giantswarm#36729; workstream giantswarm/giantswarm#37023; giantswarm/giantswarm#36941 + architect-orb#838 (Go cache precedent); `giantswarm/devctl` + `giantswarm/github` enabler issues.

---

## Outcome (resolved)

[...]

PR #1839, opened 2026-06-28
ci: cache the node-build output (node_modules) to skip native rebuilds

teemow:
## Summary

- Regenerated `.circleci/workflows.yml` with devctl v8.29.0 (giantswarm/devctl#2063/#2064): the node-build job now caches the **build output** (`node_modules` + `.yarn/install-state.gz`), keyed on the node image version + lockfile (`node-build-yarn-v1-24.18.0-{{ checksum "yarn.lock" }}`), in addition to the `.yarn/cache` tarball cache.
- **Why:** the tarball cache only saved ~14s (fetch 19s->4.7s); the warm install was still ~179s because the Link step recompiles native addons (better-sqlite3, isolated-vm, tree-sitter, …) from source via node-gyp every run. Those `.node` binaries live in `node_modules`, which the tarball cache doesn't hold. Restoring `node_modules` lets the install reconcile instead of recompiling — the Node analogue of go-build's `$GOCACHE`.

Records the warm-install number projected in #1829. Closes the perf gap that devctl#2056 alone could not.

## Test plan

- [ ] cold node-build saves node_modules under the build-cache key
- [ ] warm node-build restores it and skips the node-gyp rebuild (Link step drops sharply)

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -23,6 +23,15 @@ jobs:
         keys:
         - node-deps-yarn-v1-{{ checksum "yarn.lock" }}
         - node-deps-yarn-v1-
+    # Build-output cache: restore the materialized node_modules (with its
+    # compiled native addons) so the install reconciles incrementally instead
+    # of recompiling better-sqlite3/isolated-vm/etc. from source via node-gyp
+    # every run. Keyed on the node image version -- a node bump must not restore
+    # stale-ABI binaries. The Node analogue of go-build's $GOCACHE persist.
+    - restore_cache:
+        keys:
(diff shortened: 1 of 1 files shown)


teemow:
Closing — this was a manual validation vehicle and I shouldn't have hand-edited backstage CI. The change is now delivered the proper way: devctl v8.29.0 (giantswarm/devctl#2064) generates this build-output cache, and the align-files pin is bumped (giantswarm/github#5563), so align-files will regenerate `.circleci/workflows.yml` with it on the next run.

Validated result (from this PR's CI before closing): warm node-build install dropped from ~182s to ~9s (Link step 170s -> 2.3s, no node-gyp recompile) once node_modules + .yarn/install-state.gz are restored from the build cache.

PR #1857, opened 2026-06-30
feat(muster): use shared Table and faceted filters for workflows list

marians:
### What does this PR do?

Replaces the hand-built Material-UI table on the muster Workflows page (`/muster/workflows`) with the shared `@backstage/core-components` Table used by the Deployments and Installations pages, and moves filtering into the faceted sidebar pattern (`FiltersLayout` + `MultiplePicker` + `useFilters`).

- New `WorkflowsDataProvider` maps Workflow CRs to plain rows and applies the facet filters (mirroring `DeploymentsDataProvider`).
- **Status** (Valid / Validation warning), **Namespace**, and **Source** (GitOps / Manually added) facets in the sidebar, replacing the old inline relevance-search bar and status dropdown.
- Built-in Table search, sort, pagination, and column-visibility; default page size 25.
- Quick-search keeps token-boundary matching (so "dex" doesn't match "index"); relevance ranking is dropped (the shared Table filters but can't re-order by score).
- Installation picker moved to the top of the facet column (full width, lined up with the other widgets).
- "Create workflow" moved into the table toolbar (via a `components.Toolbar` override), next to the column-selection button.
- Description shown under the name in the Name column; the name truncates responsively with a CSS ellipsis (full name on wide viewports, ellipsis only when the column is narrow) rather than a fixed character cap. Per-row kebab menu removed.
- Responsive column widths (`tableLayout: 'fixed'`) so Name/description gets the bulk of the space.
- Intro blurb moved above the table with updated copy and a docs link to the muster workflow-authoring tutorial.
- Drops the now-unused `workflowSearch` relevance util; replaced by a trimmed token-boundary matcher with tests.

The Workflows list now looks and behaves consistently with the other list pages (Deployments, Installations): faceted sidebar filters (Status, Namespace, Source), persisted column visibility, built-in search/sort/pagination, and a responsive layout. The per-row kebab menu is gone; "Run workflow" remains reachable from the workflow detail page.

### How does it look like?

image

Follow-up to the muster section modernization. Brings the last hand-built table in the section onto the shared pattern.

No. (The table patterns learned here were captured in the internal `.claude/skills/tables` guide.)

- [ ] A changeset describing the change and affected packages was added. (`@giantswarm/backstage-plugin-muster` is a private package that ships with the app image, so no changeset is required.)

Towards

- https://github.com/giantswarm/giantswarm/issues/37055


plugins/muster/src/components/WorkflowsListPage/WorkflowsTable/columns.tsx
```diff
@@ -0,0 +1,119 @@
...
+      title: 'Name',
+      field: WorkflowColumns.name,
+      highlight: true,
+      defaultSort: 'asc',
+      render: row => <WorkflowNameCell row={row} />,
+      // The description is shown under the name, so keep it searchable here.
+      customFilterAndSearch: (query, row) =>
+        `${row.name} ${row.description}`
+          .toLowerCase()
+          .includes(query.toLowerCase()),
```
marians: **Search regression: reintroduces the “dex” → “index” (F3) bug.**

This `customFilterAndSearch` does plain substring matching over `name + description`. The util this PR deletes (`workflowSearch.ts`) was written specifically to *avoid* this: it matched on token boundaries so `"dex"` would not match a row whose description contains `index_stats`. Its test (`workflowSearch.test.ts`, also deleted) asserted exactly this case returns no match.

Now, searching `"dex"` matches `loki-request-errors` (description `"…index_stats…"`) because `"loki-request-errors index_stats".includes("dex")` is `true` via the substring `in`**`dex`**`_stats`. The relevance ranking (name > description, exact > prefix) is also gone.

If the F3 behavior is still a requirement, the column needs boundary-aware matching (and ideally a replacement test). If the team has consciously accepted plain substring search as the cost of moving to the shared Table, worth saying so in the PR description so the dropped `workflowSearch` coverage is an intentional trade-off rather than a silent regression.

plugins/muster/src/components/WorkflowsListPage/WorkflowsTable/WorkflowsTable.tsx
```diff
@@ -0,0 +1,105 @@
...
+        {
+          // The icon is unused — `components.Action` renders the button instead.
+          icon: () => <></>,
+          isFreeAction: true,
+          onClick: () => {},
+        },
+      ]}
+      components={{
+        Action: ActionComponent,
+      }}
```
marians: **Fragile, non-idiomatic way to put a button in the toolbar.**

This couples two hacks: a dummy free-action whose `icon` is `() => <></>` and an `onClick` that does nothing, plus a `components.Action` override that ignores the `action` prop and always renders `CreateWorkflowButton`. material-table renders `components.Action` *once per entry* in `actions`, so this works only as long as there is exactly one action. If anyone later adds a second action (e.g. a reload button, like `DeploymentsTable`/`ClustersTable`), `ActionComponent` renders twice → two “Create workflow” buttons. It also silently depends on material-table internals, so a Backstage bump can break it with no type error.

No other table in the repo overrides `components.Action`. Cleaner in-repo idioms:
- `components.Toolbar` override (see `NodePoolNodesTable`) — render the standard title/columns button plus the Create button.
- Or just keep the button in the page body where it lived before this PR (the simplest option).

plugins/muster/src/components/WorkflowsListPage/WorkflowsTable/columns.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <AvailabilityBadge available={row.available} />
+          {row.validationWarning && (
+            <StateBadge tone="warning" label="Validation warning" />
+          )}
+        </Box>
+      ),
+    },
+    {
+      title: 'Source',
+      field: WorkflowColumns.source,
```
marians, line 115: **Quick-search on the Source column matches the raw value, not the label.**

The column has a `render` showing “GitOps” / “Manually added”, but no `customFilterAndSearch` and is not `searchable: false`. material-table’s default search ignores `render` and matches the raw `row.source` value (`gitops` / `manual`). So typing “Manually added” or “GitOps” into the search box returns nothing.

Either make it match the displayed label or exclude it from search:

```suggestion
      field: WorkflowColumns.source,
      customFilterAndSearch: (query, row) =>
        (row.source === 'gitops' ? 'gitops' : 'manually added').includes(
          query.toLowerCase(),
        ),
```

plugins/muster/src/components/WorkflowsListPage/WorkflowsTable/columns.tsx
```diff
@@ -0,0 +1,119 @@
...
+      render: row => <WorkflowNameCell row={row} />,
+      // The description is shown under the name, so keep it searchable here.
+      customFilterAndSearch: (query, row) =>
+        `${row.name} ${row.description}`
+          .toLowerCase()
+          .includes(query.toLowerCase()),
+    },
+    {
+      title: 'Namespace',
+      field: WorkflowColumns.namespace,
```
marians, line 88: **Minor: lost the `-` placeholder for an empty namespace.**

The old page rendered `getNamespace() ?? '-'`; here the column binds directly to `row.namespace`, which the data provider sets to `''` when `getNamespace()` is undefined, so a namespace-less workflow now renders a blank cell. Workflow CRs are namespaced in practice, so this is cosmetic/theoretical rather than a live bug — but a one-line `render` restores the old behavior:

```suggestion
      field: WorkflowColumns.namespace,
      render: row => <>{row.namespace || '-'}</>,
```

plugins/muster/src/components/WorkflowsListPage/filters/WorkflowsFilters.tsx
```diff
@@ -0,0 +1,13 @@
+import { InstallationPicker } from '../../InstallationPicker';
+import { StatusPicker } from './StatusPicker';
+import { NamespacePicker } from './NamespacePicker';
+
+export const WorkflowsFilters = () => {
```
marians: **Convention: component not in its own directory.**

Repo `CLAUDE.md` (Code Patterns & Conventions → Component Structure): *“Components must be in their own directory with: `ComponentName.tsx` … `index.ts`”*. Its siblings `StatusPicker` and `NamespacePicker` follow this, but `WorkflowsFilters` sits directly at `filters/WorkflowsFilters.tsx` (re-exported by `filters/index.ts`). Moving it to `filters/WorkflowsFilters/{WorkflowsFilters.tsx,index.ts}` would match the convention and its neighbours. Minor.

PR #1859, opened 2026-06-30
feat(muster): YAML-based Create workflow modal with shared editor

marians:
### What does this PR do?

Revamps the muster **Create workflow** modal (also used for editing ad-hoc workflows) on `/muster/workflows`, and extracts the editor wrapper it uses into a shared package.

- **Shared editor wrapper:** moves `YamlEditorFormField` from `plugins/gs` into `@giantswarm/backstage-plugin-ui-react` (next to the `YamlEditor` it wraps) and exports it. The `gs` plugin now re-exports it from the package, so its existing scaffolder editors (`ValueSourcesEditor`, `YamlValuesEditor`, `SecretYamlValuesEditor`) are unchanged.
- **YAML editing:** the workflow definition is now edited as YAML (seeded with `yaml.dump`, parsed with `yaml.load`) using the shared CodeMirror editor instead of a plain JSON textarea.
- **Close affordance:** the footer "Close" button is replaced with an X icon in the title bar.
- **Save button:** now uses the standard primary color instead of the red `secondary` color.
- **Stable layout:** validation/error output renders in a fixed-height, scrollable region so the modal no longer resizes when a message appears.

The "Create workflow" (and ad-hoc "Edit workflow") modal is easier to use: definitions are authored in YAML with a real code editor (syntax highlighting, line numbers), closing is a clear top-right X, the primary action is no longer styled as a destructive button, and the dialog no longer jumps in size when validation output appears.

### How does it look like?

image

`AdHocWorkflowDialog` backs both the create flow (`CreateWorkflowButton`) and the ad-hoc edit flow (`WorkflowMutationActions`), so all changes apply consistently to both.

No.

Towards

- https://github.com/giantswarm/giantswarm/issues/37055


plugins/muster/src/components/WorkflowsListPage/WorkflowMutationActions.tsx
```diff
@@ -266,11 +276,11 @@ export function AdHocWorkflowDialog({
 
   const parsed = (): Record<string, unknown> | undefined => {
     try {
-      const obj = JSON.parse(value);
+      const obj = yaml.load(value) as Record<string, unknown>;
```
marians: **Empty / non-object YAML now fails silently (feedback regression).**

`yaml.load("")` (or whitespace/comment-only input) returns `undefined` — it does *not* throw like `JSON.parse("")` did. So `parsed()` falls through to the success branch, calls `setError(undefined)` (clearing any prior error), and returns `undefined`. In `validate()`/`save()` the `if (!def) return;` guard then bails with **no message at all**: a user who clears the editor and clicks Validate/Save sees nothing happen, whereas before they got `Invalid JSON: Unexpected end of JSON input`.

Likewise a bare scalar/array parses to a non-`Record` value that the `as Record<string, unknown>` cast hides and forwards to `callTool`.

Consider rejecting non-object results explicitly, e.g.:

```suggestion
      const obj = yaml.load(value);
      if (obj === null || typeof obj !== "object" || Array.isArray(obj)) {
        setError("Workflow definition must be a YAML mapping.");
        return undefined;
      }
      setError(undefined);
      return obj as Record<string, unknown>;
```

marians: Good catch — fixed in c9f266c2. `parsed()` now rejects empty/comment-only input and non-mapping documents (scalar/array) with "Workflow definition must be a YAML mapping." instead of silently returning `undefined`.

PR #1862, opened 2026-07-01
feat(muster): workflow detail page cleanups and shared StackedBarChart

marians:
### What does this PR do?

A batch of UI cleanups on the muster **workflow detail** page, plus a new shared chart component in `ui-react`:

- Removes the breadcrumb from the workflow detail header.
- Renders the workflow description's Markdown paragraphs with the standard MUI `body1` typography so their line height/spacing matches other paragraphs.
- Adds a theme-aware, recharts-based **`StackedBarChart`** to `@giantswarm/backstage-plugin-ui-react` and uses it for the "Runs per day" chart. The chart now spans at least **30 days**, always ends on **today**, and fills days with no runs.
- Removes the redundant collapsible "Definition (YAML)" section below the steps.
- Trims the GitOps-managed actions to a single **"Show manifest"** outline button (dropping "Remove via GitOps"), and aligns its dialog with the "Create workflow" modal: X icon in the title bar (no footer Close button), primary "Copy manifest" button, and the manifest shown read-only in the shared YAML editor with syntax highlighting. Adds a `readOnly` passthrough to `YamlEditorFormField`.
- Gives the dialog footers consistent padding (they were cramped against the modal edge).
- Sizes the "Executions" panels to their content with a viewport-based max-height, instead of always reserving a tall fixed height when empty.

A cleaner, more consistent workflow detail page: no redundant breadcrumb/YAML block, readable description text, a proper 30-day runs chart, GitOps manifest dialog that matches the create-workflow modal, and an Executions section that no longer reserves ~900px when empty.

Introduces recharts (v3) as a dependency of `ui-react` only; consuming plugins use the themed `StackedBarChart` wrapper rather than recharts directly.

No.

Towards https://github.com/giantswarm/giantswarm/issues/37055

### DIFF
--- a/plugins/muster/src/components/WorkflowDetailPage/WorkflowStatsPanel.tsx
+++ b/plugins/muster/src/components/WorkflowDetailPage/WorkflowStatsPanel.tsx
@@ -51,7 +51,14 @@ function padPerDay(perDay: WorkflowStatsPerDay[]): WorkflowStatsPerDay[] {
   }
 
   const byDate = new Map(perDay.map(day => [day.date, day]));
-  const epochs = perDay.map(day => parseDay(day.date));
+  // Drop any unparseable dates so a single malformed key can't turn `end`/`start`
+  // into NaN and collapse the whole chart to an empty range.
+  const epochs = perDay
+    .map(day => parseDay(day.date))
+    .filter(epoch => !Number.isNaN(epoch));
+  if (epochs.length === 0) {
+    return perDay;
+  }
   // Anchor the right edge to today, but never truncate data that (somehow)
   // carries a later date.
   const end = Math.max(...epochs, todayEpoch());
--- a/plugins/muster/src/components/WorkflowDetailPage/WorkflowStatsPanel.tsx
+++ b/plugins/muster/src/components/WorkflowDetailPage/WorkflowStatsPanel.tsx
@@ -1,7 +1,6 @@
 import {
   Box,
   Paper,
-  Tooltip,
   Typography,
   makeStyles,
   useTheme,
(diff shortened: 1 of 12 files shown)


plugins/muster/src/components/WorkflowsListPage/WorkflowMutationActions.tsx
```diff
@@ -108,9 +102,18 @@ function GitOpsManifestDialog({
 
   return (
     <Dialog open={open} onClose={onClose} maxWidth="md" fullWidth>
-      <DialogTitle>
-        {intent === 'delete' ? 'Remove via GitOps' : 'Edit via GitOps'} —{' '}
-        {workflow.getName()}
+      <DialogTitle disableTypography className={classes.titleBar}>
+        <Typography variant="h6">
+          {intent === 'delete' ? 'Remove via GitOps' : 'Edit via GitOps'} —{' '}
```
marians: The button that opens this dialog now reads **"Show manifest"** (line 439), but the dialog title still says **"Edit via GitOps — <name>"**. Since `intent` is now always `"edit"` (the only caller passes `intent="edit"`), the `intent === 'delete'` branches here and in the body text (line 129) are also dead. Consider retitling to match the neutral "Show manifest" action and dropping the now-unreachable `delete` branch / `intent` prop.

plugins/muster/src/components/WorkflowDetailPage/WorkflowStatsPanel.tsx
```diff
@@ -10,10 +9,84 @@ import {
...
+function padPerDay(perDay: WorkflowStatsPerDay[]): WorkflowStatsPerDay[] {
+  if (perDay.length === 0) {
+    return perDay;
+  }
+
+  const byDate = new Map(perDay.map(day => [day.date, day]));
+  const epochs = perDay.map(day => parseDay(day.date));
+  // Anchor the right edge to today, but never truncate data that (somehow)
+  // carries a later date.
+  const end = Math.max(...epochs, todayEpoch());
```
marians, line 64: Defensive edge: a single malformed `date` in `data.per_day` silently hides the entire chart. `parseDay('bad')` returns `NaN`, so `end = Math.max(...epochs, todayEpoch())` is `NaN`, `start` is `NaN`, and the `for (epoch = start; epoch <= end; ...)` loop never runs — `padPerDay` returns `[]` and the `perDay.length > 0` guard (line 237) hides the whole "Runs per day" panel even though most days were valid. Low risk since the backend controls the key format, but a guard (skip/`Number.isNaN` filter on parsed epochs) would be cheap insurance.

plugins/muster/src/components/WorkflowsListPage/WorkflowMutationActions.tsx
marians: Done in 4965e6d0. Retitled the dialog to "Workflow manifest — <name>", removed the `intent` prop and both `delete` branches (title + body text), and renamed the state to `manifestOpen`.

plugins/muster/src/components/WorkflowDetailPage/WorkflowStatsPanel.tsx
marians: Done in 4965e6d0. `padPerDay` now filters out `NaN` epochs before computing the range (and returns the input unchanged if none parse), so one malformed date no longer collapses the chart.

PR #1864, opened 2026-07-01
feat(gs): compact cluster access widget and cover non-broker installations

marians:
### What does this PR do?

Two improvements to the "Cluster access" sidebar popover widget:

1. **Cleaner, more compact layout.** The header now shows a problem-first count summary (`1 degraded · 4 healthy`, or `All N healthy` when nothing needs attention) with problem counts colored, instead of a single worst-state word. Each installation is a single-line row: healthy/connecting rows drop the redundant status label, and degraded/session-expired rows show the reason left-aligned right next to the name so the list is easy to scan across many rows.

2. **Non-broker installations are now visible.** Previously the widget only listed broker-covered installations, because those are the only ones the connector probes. Installations connected via a per-cluster OIDC popup never appeared. They can't be probed proactively (a proxy call without a live session would trigger a login popup), so the connector now mirrors their auth session state instead: an installation shows up while the user is signed in and drops off on sign-out. This adds a `remove` method to the cluster-access status store.

- The Cluster access overlay is denser and quicker to read, especially with many installations.
- Installations you're signed in to via a per-cluster login now appear in the widget alongside broker-covered ones.

### How does it look like?

Each row now reads e.g. `● grouse  API unreachable (timeout)` — status dot, name, then the reason immediately after the name (left-aligned). Healthy clusters are just `● name`. Header example: `Cluster access` / `1 degraded · 4 healthy`.

image

Non-broker installations are surfaced via `authApi.sessionState$()` (`SignedIn` → record healthy, `SignedOut` → remove), which never triggers a popup — the same signal the provider settings page already uses.

No.

### DIFF
--- a/plugins/gs/src/components/ClusterAccessStatus/ClusterAccessStatusSidebarItem.tsx
+++ b/plugins/gs/src/components/ClusterAccessStatus/ClusterAccessStatusSidebarItem.tsx
@@ -94,23 +94,16 @@ const useStyles = makeStyles(theme => ({
 }));
 
 /**
- * Computes the worst (most actionable) state across all tracked clusters.
- * session-expired wins over degraded, because an expired main session is the
- * one thing the user can immediately fix. `connecting` ranks above plain
- * healthy so the badge reflects in-flight probes on startup, but below any
- * real problem.
+ * Computes the worst (most actionable) state across all tracked clusters, using
+ * the shared {@link STATE_ORDER} severity ranking: session-expired wins over
+ * degraded, because an expired main session is the one thing the user can
+ * immediately fix; `connecting` ranks above plain healthy so the badge reflects
+ * in-flight probes on startup, but below any real problem. No entries → healthy.
  */
 function overallState(entries: ClusterAccessStatusEntry[]): ClusterAccessState {
-  if (entries.some(e => e.state === 'session-expired')) {
-    return 'session-expired';
-  }
-  if (entries.some(e => e.state === 'degraded')) {
-    return 'degraded';
-  }
-  if (entries.some(e => e.state === 'connecting')) {
-    return 'connecting';
-  }
-  return 'healthy';
+  return (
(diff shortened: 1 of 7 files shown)


plugins/gs/src/components/ClusterAccessStatus/ClusterAccessStatusSidebarItem.tsx
```diff
@@ -36,20 +32,62 @@ const STATE_LABELS: Record<ClusterAccessState, string> = {
...
+  'session-expired': 'session expired',
+};
+
+// Severity order shared by the summary and the row list -- worst first.
+const STATE_ORDER: ClusterAccessState[] = [
+  'session-expired',
+  'degraded',
+  'connecting',
+  'healthy',
+];
```
marians, line 49: **Cleanup — duplicated state-priority ordering (drift risk).**

This PR introduces `STATE_ORDER` (`session-expired > degraded > connecting > healthy`), but `overallState()` (lines 103–114) still hand-rolls the exact same priority with a chain of `.some()` checks. The two now encode the same ordering in two places, so a future reordering of `STATE_ORDER` would silently diverge from the badge color computed by `overallState()`.

Since the priorities are identical, `overallState()` can derive from `STATE_ORDER` and the duplication goes away:

```ts
function overallState(entries: ClusterAccessStatusEntry[]): ClusterAccessState {
  return (
    STATE_ORDER.find(state => entries.some(e => e.state === state)) ?? 'healthy'
  );
}
```

Empty entries → `find` returns `undefined` → `'healthy'`, matching the current behavior.

marians: Good catch — done in 4c92b833. `overallState()` now derives from `STATE_ORDER` (`STATE_ORDER.find(...) ?? 'healthy'`), so the ordering lives in one place.

PR #1867, opened 2026-07-02
feat(gs): add Gateways tab with HTTPRoute hostnames and certificates

marians:
### What does this PR do?

Adds a **Gateways** tab to the deployment detail page (`/deployments/{installation}/helmrelease/{namespace}/{name}/gateways`) that lists the hostnames exposed via Gateway API HTTPRoutes for the workload, alongside their serving TLS certificates.

- **Hostnames table** (bui `Table`): built from the `gatewayapi_httproute_hostname_info` metric, scoped by `cluster_id` + the workload's target namespace. Duplicate series (a route and its HTTP→HTTPS redirect route sharing a hostname) are de-duplicated. Each hostname links to `https://<hostname>/`.
- **Certificate column**: each row resolves its serving certificate by walking the Gateway API chain — HTTPRoute `parentRef` → Gateway listener (wildcard-matched against the hostname) → cert-manager Certificate via the conventional `gateway-<gateway>-<listener>` name. Shows readiness + expiry, color-coded by severity (red if not-ready/expired, amber if <14 days, green if ready). Issuer, certificate name/namespace, and the matched host pattern are behind an info (ⓘ) tooltip.
- **About card**: now shows the workload's **Target namespace** when it differs from the HelmRelease/App CR namespace.
- Registers the new metrics in the central Mimir metrics registry, and documents the bui `Table` (plus its loading-skeleton gotcha) in the `tables` skill.

Users can see, per deployment, which hostnames it exposes through Envoy Gateway HTTPRoutes and whether the TLS certificates serving those hostnames are healthy and when they expire — without leaving Backstage.

### How does it look like?

image

image

The certificate join is metrics-only. The final listener→certificate hop relies on cert-manager's gateway-shim naming convention (`gateway-<gateway>-<listener>`); when no matching certificate series is found, the certificate is shown as unresolved rather than guessed. The metric carries no backendRef, so hostnames are correlated to the workload by namespace — the tab copy notes the table may include entries unrelated to the specific release.

Yes

### DIFF
--- a/plugins/gs/src/components/deployments/deployment-details/DeploymentGateway/DeploymentHttpRoutesCard/DeploymentHttpRoutesCard.tsx
+++ b/plugins/gs/src/components/deployments/deployment-details/DeploymentGateway/DeploymentHttpRoutesCard/DeploymentHttpRoutesCard.tsx
@@ -1,0 +1,291 @@
+import { InfoCard } from '@giantswarm/backstage-plugin-ui-react';
+import {
+  ButtonIcon,
+  Cell,
+  CellText,
+  ColumnConfig,
+  Link,
+  Table,
+  Tooltip,
+  TooltipTrigger,
+} from '@backstage/ui';
+import { Typography, makeStyles } from '@material-ui/core';
+import Alert from '@material-ui/lab/Alert';
+import InfoOutlined from '@material-ui/icons/InfoOutlined';
+import { useCurrentDeployment } from '../../../DeploymentDetailsPage/useCurrentDeployment';
+import { findTargetClusterName } from '../../../utils/findTargetCluster';
+import { getWorkloadNamespace } from '../../../utils/getWorkloadIdentifiers';
+import { useMimirHttpRouteHostnames } from '../../../../hooks/useMimirHttpRouteHostnames';
+import { HostnameCertInfo } from '../../../../hooks/resolveHostnameCert';
+import { HttpRouteHostname } from '../../../../hooks/useMimirHttpRouteHostnames';
+
+type HostnameRow = HttpRouteHostname & { id: string };
+
+const useStyles = makeStyles(theme => ({
+  caption: {
+    marginBottom: theme.spacing(1),
+    maxWidth: 1000,
+  },
+  footnote: {
+    marginTop: theme.spacing(2),
+    maxWidth: 1000,
+  },
+  certPrimary: {
+    display: 'flex',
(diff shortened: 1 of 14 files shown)


plugins/gs/src/components/hooks/useMimirHttpRouteHostnames.ts
```diff
@@ -0,0 +1,152 @@
...
+  installationName: string;
+  clusterName: string | undefined;
+  namespace: string | undefined;
+  refetchInterval?: number | false;
+}) {
+  const { installationName, clusterName, namespace, refetchInterval } = options;
+
+  const isEnabled = Boolean(clusterName && namespace);
+  const nsSelector = `cluster_id="${clusterName}",namespace="${namespace}"`;
+  const clusterSelector = `cluster_id="${clusterName}"`;
```
marians, line 81: Low-severity / non-blocking: the listener, expiration and ready queries are scoped only by `cluster_id`, so each render of this tab fetches **every** cert-manager Certificate and **every** Gateway listener in the cluster (polled every 30s) just to resolve a handful of hostnames. Fine at today's cardinality, but worth keeping in mind as clusters grow. Narrowing isn't trivial here since the Gateway namespace isn't known until the parent query resolves, so I'd leave it as-is unless it shows up in Mimir load — flagging for awareness.

plugins/gs/src/components/deployments/deployment-details/DeploymentGateway/DeploymentHttpRoutesCard/DeploymentHttpRoutesCard.tsx
```diff
@@ -0,0 +1,291 @@
...
+}
+
+function certGroupVersion(labels: Record<string, string>): string {
+  const group = labels.customresource_group;
+  const version = labels.customresource_version;
+  if (group && version) return `${group}/${version}`;
+  return group ?? version ?? '—';
+}
+
+function CertificateCell({ cert }: { cert?: HostnameCertInfo }) {
```
marians: Nit / optional: `CertificateCell` (plus its dedicated styles and the `certSeverity`/`formatExpiry`/`readyLabel` helpers) is ~150 lines inlined in this 291-line file. Per the repo's component-structure convention it could move to its own `CertificateCell/` directory with an `index.ts`. It's a private render helper so this is a judgment call, not a rule break — mentioning in case you'd prefer to split it out.

PR #1874, opened 2026-07-03
fix(auth): reclassify cluster token-exchange failures to cut Sentry noise

marians:
### What does this PR do?

The cluster token-exchange router in `plugins/auth-backend-module-gs/src/clusterToken/router.ts` logged a `warn` for **every** non-OK broker response, embedding the installation name and raw response body in the message. The winston→Sentry bridge forwards `warn`/`error` logs, and Sentry fingerprints on the message text, so a handful of root causes exploded into dozens of Sentry issues and ~1000+ events — even though all of these are already handled gracefully (the handler returns a structured `502`/`reason` to the frontend, 0 user impact).

Two changes:

1. **Reclassify by cause.** Routine subject-token rejections (`invalid_grant` / `invalid_token` / `invalid_request`, and bare 401s that are not `invalid_client`) — the `subject_invalid` cases — now log at `debug`. They're expected and already handled, so they no longer reach Sentry. Genuine broker faults stay at `warn`: `invalid_client` (broker self-auth failure → `broker_client_invalid`), `invalid_target` / other broker-side failures (`exchange_failed`), an unreachable broker, and a missing `access_token`.
2. **Stop per-installation Sentry sprawl.** The installation name and raw body move out of the log *message* and into structured metadata, and each root cause uses a constant message. Since the Sentry transport groups on message text, one broker outage now collapses into a single Sentry issue instead of ~24 (one per installation).

> Note: the `winston-sentry-log` transport computes a `fingerprint` but never applies it to the scope, so grouping is driven purely by message text — hence the constant-message approach rather than an explicit fingerprint.

This path is only active where `gs.clusterTokenBroker` is configured (today, the internal dev portal instance).

No user-facing change. Handler responses (`502` + `reason`) are unchanged; this is observability hygiene only.

### How does it look like?

Log level and message change only. Example: a rejected/expired subject token now logs at `debug` (`Cluster token exchange rejected: subject token invalid or expired`) with `{ installation, status, oauthError, body }` in metadata, instead of a per-installation `warn`.

Closes giantswarm/giantswarm#37086.

No.

  The affected package (`@giantswarm/backstage-plugin-auth-backend-module-gs`) is `private`, so no changeset is required.

### DIFF
--- a/plugins/auth-backend-module-gs/src/clusterToken/router.ts
+++ b/plugins/auth-backend-module-gs/src/clusterToken/router.ts
@@ -157,10 +157,13 @@ export function createClusterTokenRouter(
           body: params.toString(),
         });
       } catch (error) {
-        logger.warn(
-          `Cluster token exchange for installation "${installation}" failed: broker unreachable`,
-          error as Error,
-        );
+        // A broker fault, not a user problem. Keep the installation out of the
+        // message so the winston->Sentry bridge groups every installation's
+        // outage into a single issue instead of one per installation.
+        logger.warn('Cluster token exchange failed: token broker unreachable', {
+          installation,
+          error: String(error),
+        });
         res.status(502).json({
           error: 'Token broker is unreachable',
           reason: 'broker_unreachable',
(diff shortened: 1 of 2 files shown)


plugins/auth-backend-module-gs/src/clusterToken/router.ts
```diff
@@ -157,10 +157,13 @@ export function createClusterTokenRouter(
...
-          `Cluster token exchange for installation "${installation}" failed: broker unreachable`,
-          error as Error,
-        );
+        // A broker fault, not a user problem. Keep the installation out of the
+        // message so the winston->Sentry bridge groups every installation's
+        // outage into a single issue instead of one per installation.
+        logger.warn('Cluster token exchange failed: token broker unreachable', {
+          installation,
+          error: String(error),
+        });
```
marians, line 173: `String(error)` drops the actionable root cause on the one path here that still reaches Sentry.

For an undici `fetch` rejection, `String(error)` collapses to `"TypeError: fetch failed"` — the underlying code (`ECONNREFUSED`, `ENOTFOUND`, DNS/TLS failure) lives on `error.cause`, not on the top-level message. So a genuine broker outage — the case this branch deliberately keeps at `warn` so it *does* surface in Sentry — arrives with no distinguishing detail, and every outage cause looks identical. (Matches the earlier finding that k8s-proxy fetch errors hide the real code in a nested field, not the message.)

```suggestion
        // A broker fault, not a user problem. Keep the installation out of the
        // message so the winston->Sentry bridge groups every installation's
        // outage into a single issue instead of one per installation.
        logger.warn('Cluster token exchange failed: token broker unreachable', {
          installation,
          error: String(error),
          // undici puts the actionable code (ECONNREFUSED/ENOTFOUND) on
          // error.cause; String(error) alone collapses to "TypeError: fetch failed".
          cause:
            error instanceof Error && error.cause !== undefined
              ? String(error.cause)
              : undefined,
        });
```

marians: Good catch — fixed in 71c3085d. Added a `cause` field that pulls the underlying code off `error.cause` (falling back to `null` rather than `undefined` to keep the meta a valid `JsonObject`, consistent with `oauthError ?? null`). Also strengthened the unreachable-broker test to reject via `new TypeError('fetch failed', { cause: new Error('ECONNREFUSED') })` and assert the outage log surfaces `cause: ECONNREFUSED`.

PR #1896, opened 2026-07-09
fix(plans): surface GitHub comment errors and highlight comment threads

paurosello:
## What

Two fixes to the plans plugin's PR-comment experience:

### Backend — actionable errors instead of opaque 500s
The `plans-backend` GitHub proxy collapsed every non-2xx GitHub response into a generic `Error` → **HTTP 500**. So when the deployed GitHub App lacks `Pull requests` / `Issues` **write** permission, replying to a comment failed with an opaque 500 and no explanation (and was reported to Sentry as a server fault).

Now a GitHub **403** is mapped to a `NotAllowedError` (**HTTP 403**) and GitHub's own message (e.g. *"Resource not accessible by integration"*) is included, so the reply form shows an actionable error. As a 4xx it also no longer pages us via Sentry. Other non-2xx responses still surface GitHub's body in the message.

### Frontend — comment threads stand out
Review threads and the PR discussion now render inside a single primary-colored box (with a left accent) instead of blending into the page, so each conversation reads as one unit.

## Why
While testing comment replies against `giantswarm/bumblebee-plans`, a permissions 403 surfaced as a 500 with no detail. The root cause is a GitHub App permission gap (fixed separately on the App), but the proxy should report it clearly regardless.

## Testing
- `yarn test plugins/plans-backend` — added a test asserting a GitHub 403 → `NotAllowedError` (403) with the message surfaced; 28/28 pass.
- `yarn test plugins/plans` — 73/73 pass.
- Lint clean on changed files.

A changeset is included (both plugin packages, patch).

### DIFF
--- a/.changeset/plans-comment-errors-and-thread-highlight.md
+++ b/.changeset/plans-comment-errors-and-thread-highlight.md
@@ -1,0 +1,16 @@
+---
+'@giantswarm/backstage-plugin-plans-backend': patch
+'@giantswarm/backstage-plugin-plans': patch
+---
+
+Surface GitHub errors on plan comment writes and make comment threads stand out.
+
+- The backend proxy no longer collapses every non-2xx GitHub response into an
+  opaque 500. A GitHub 403 (typically the GitHub App missing `Pull requests` /
+  `Issues` write permission) is now mapped to a `NotAllowedError` (HTTP 403) and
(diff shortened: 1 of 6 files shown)


paurosello:
image

QuentinBisson (approved):

PR #1897, opened 2026-07-09
feat(agent-platform): agent creation flow with direct-apply deploy

marians:
### What does this PR do?

Adds a new `agent-platform` frontend plugin with an agent-creation flow
(`/agents/new` → `/agents/new/review`), a native re-implementation of the APUI
(Agent Platform User Interface) prototype.

The user picks an installation and a kagent `ModelConfig`, edits a system prompt,
selects **skills**, and reviews the generated Flux manifests (`OCIRepository` +
`HelmRelease`, with the agent values inlined). **Deploying applies those
resources directly** to the selected installation via the `kube:apply` scaffolder
action, authenticated with the user's per-installation OIDC token (minted the
same way the `GSOIDCToken` field does). Copy-manifests and `helm install` remain
as a manual fallback — there is no pull-request path.

**Skills discovery.** Skills are discovered from the GitHub repositories
configured in `agentPlatform.skills.repositories` (default:
`giantswarm/agent-skills`). A skill is any directory containing a `SKILL.md`; its
`name`/`description` frontmatter drives a multi-select picker. Discovery runs
backend-side in `gs-backend` (`GET /agent-skills?repoUrl=…`): it walks the repo's
git tree, finds every `SKILL.md`, and parses the frontmatter, authenticating via
the GitHub integration (public repos also work unauthenticated). A selected skill
becomes a kagent `spec.skills.gitRefs` entry (`{ url, path, ref, name }`) — the
real v1alpha2 shape (there is no OCI/image skill source, despite a stale CRD doc
comment).

Key decisions:

- The resources land in the **selected ModelConfig's namespace** (co-located with
  the model, where kagent watches), rather than a hardcoded namespace.
- The `HelmRelease` sets `spec.serviceAccountName` (from
  `agentPlatform.fluxServiceAccountName`), required by the GS Flux multi-tenancy
  admission policy for HelmReleases in tenant namespaces.
- Deploy runs through a hidden `agent-deployment` catalog template.
- The fleet-wide `ModelConfig` query only hits **reachable** installations (via
  the shared cluster-access status) and skips API-version discovery, so it no
  longer stalls on unreachable clusters.

Also includes a small, separate `fix(deps)` commit pinning `@assistant-ui/core`'s
zustand peer to v5 (a pre-existing conflict with `@stoplight/mosaic`'s zustand v3
that broke the frontend build on a fresh install — unrelated to the feature).

> ⚠️ **Draft — not ready to merge.** The flow works end-to-end (token → apply →
> resources created), but there are real open items (see below and
> `docs/agent-platform.md`).

A new "Agents" area for creating a kagent agent from Backstage. The page is
`disabled` by default and enabled per environment via `app.extensions` in
app-config, so there is no effect until it is switched on.

### How does it look like?

Two bui pages: a create form (identity + configuration — system prompt,
installation, model cards with a selection check-circle, and a multi-select skill
picker showing each skill's source repo) and a review page that renders the
generated YAML in CodeMirror with copy buttons, a Deploy action, and a
manual-install fallback.

_(Screenshots to be added.)_

This started as an "open a PR" flow and pivoted to **direct apply** via
`kube:apply` — the same action the `app-deployment` template uses. The first live
run on `gazelle` surfaced the Flux multi-tenancy ServiceAccount requirement, now
handled via config.

The `general-purpose-agent` chart **does not exist yet**, so the chart
URL/version/values shape and the deploy ServiceAccount are provisional: a deploy
will pass admission and create the resources, but the `HelmRelease` will not
reconcile until the chart is published. Full context and the open TODOs are in
`docs/agent-platform.md` — notably:

- `ModelConfig` querying still lists across **all namespaces** (admin-only); a
  namespace-scoped strategy and non-admin RBAC (defined outside Backstage) are
  still needed. The slowness/silent-drop behaviour is fixed.
- The canonical Flux deploy ServiceAccount + RBAC is an open platform decision.
- The `agent-deployment` template must be added to `giantswarm/backstage-catalogs`
  for the deploy button to work outside local development.
- Skills: discovery/selection now work; the chart values shape for
  `gitRefs`, private-repo auth (`gitAuthSecretRef`), and per-skill version
  selection are still open.
- An agent list/management view and a main-menu entry are not yet built.

Yes — added `docs/agent-platform.md` in this PR.

Not needed: the `agent-platform` plugin is `private: true` (not published to npm),
so no changeset applies. The app image/chart release is driven by the
conventional-commit PR title.


plugins/agent-platform/src/lib/composeManifests.ts
```diff
@@ -0,0 +1,251 @@
...
+  // Pin the technical (CR) name to the slug so it doesn't depend on Flux's
+  // release-name derivation; the chart would otherwise default it.
+  lines.push(`  name: ${quote(model.slug)}`);
+  lines.push(`  displayName: ${quote(model.name)}`);
+  lines.push(`  description: ${quote(model.description)}`);
+  // Only override the prompt when the user provided one; an empty field means
+  // "use the chart's default agent.systemMessage" (and the chart requires a
+  // non-empty value, so we must not emit an empty one).
+  if (model.systemMessage.trim()) {
+    lines.push('  systemMessage: |-');
```
marians: **Correctness (confirmed — reproduced):** The hand-rolled `|-` block scalar produces **invalid YAML** when the system prompt's first non-empty line is more indented than a later line. YAML fixes the block indentation from the first content line, so a less-indented later line terminates the block early and is re-parsed as a mapping key.

This input:
```
  Introduction paragraph
Actual instructions here
```
composes a HelmRelease that `js-yaml`/`kube:apply` rejects with `bad indentation of a mapping entry`. A user pasting a prompt with an indented first line (markdown list, indented intro, fenced code) gets an opaque deploy failure; other shapes silently truncate the prompt.

Root cause is emitting YAML by string concatenation. `js-yaml` is already a dependency (used in this PR's `frontmatter.ts`) — building the values object and `yaml.dump(obj, { lineWidth: -1 })` removes the whole quote/indent/blockScalar machinery and this bug class.

plugins/agent-platform/src/components/NewAgentFormProvider/NewAgentFormProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    return {
+      state,
+      setName: name =>
+        setState(prev => ({
+          ...prev,
+          name,
+          slug: slugEdited ? prev.slug : slugify(name),
+        })),
+      setSlug: slug => {
```
marians, line 111: **Correctness:** `setSlug` stores the manually-typed value verbatim, and `missingRequired` (L79-89) only checks it's non-empty. The auto-`slugify` guard is bypassed once the user edits the Slug field, so `My Agent!`, a leading/trailing hyphen, or a >63-char value passes validation and is emitted **unquoted** as `metadata.name` (composeManifests.ts L182) → invalid DNS-1123 name → deploy fails late at `kube:apply` with an opaque k8s error instead of an inline form error. Validate the slug against the RFC1123 label regex before enabling review.

plugins/gs-backend/src/agentSkills/discoverAgentSkills.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const branch = ref ?? (await resolveDefaultBranch(owner, repo, token));
+
+  const treeResponse = await githubApiFetch(
+    `https://api.github.com/repos/${owner}/${repo}/git/trees/${encodeURIComponent(
+      branch,
+    )}?recursive=1`,
+    token,
+  );
+  const tree = (await treeResponse.json()) as {
```
marians, line 180: **Correctness:** `tree.truncated` is destructured into the type but never checked. GitHub truncates the recursive git-tree at ~100k entries / 7MB; past that, `tree.tree` is partial and any `SKILL.md` beyond the cap is silently dropped from the picker with no signal the list is incomplete. Detect `truncated` and surface it (or fall back to paginated tree traversal).

plugins/gs-backend/src/router.ts
```diff
@@ -154,5 +155,45 @@ export async function createRouter({
...
+      ref: z.string().optional(),
+    });
+    const parsed = schema.safeParse(req.query);
+    if (!parsed.success) {
+      throw new InputError(parsed.error.toString());
+    }
+
+    const { repoUrl, ref } = parsed.data;
+
+    const skills = await discoverAgentSkills({
```
marians: **Robustness / Sentry noise:** Unlike the sibling `/github/raw-content` handler (which inspects `response.ok` and forwards the upstream status), `/agent-skills` lets `discoverAgentSkills` → `githubApiFetch` throw a generic `Error` on any non-2xx GitHub response. Expected outcomes (repo not found → 404, rate limit → 403/429) reach the client as an undifferentiated 500, and per this repo's logging setup an error-level log for these expected cases becomes recurring Sentry noise. Forward the real upstream status the way the neighboring route does.

plugins/gs-backend/src/agentSkills/discoverAgentSkills.ts
```diff
@@ -0,0 +1,172 @@
...
+    tree?: TreeEntry[];
+    truncated?: boolean;
+  };
+
+  const skillFiles = (tree.tree ?? []).filter(
+    entry => entry.type === 'blob' && isSkillFile(entry.path),
+  );
+
+  const skills = await Promise.all(
```
marians: **Efficiency / reliability:** `Promise.all(skillFiles.map(fetchRawContent))` issues one GitHub contents request per `SKILL.md` with no concurrency cap. A repo with many skills fires that many simultaneous requests, which can trip GitHub's secondary (abuse) rate limits or exhaust the unauthenticated 60/hr quota; because it's `Promise.all` (not `allSettled`), a single 403/429 rejects the whole discovery and blanks the picker. Add a small concurrency limit, or read frontmatter in fewer calls (GraphQL / tree blobs).

plugins/agent-platform/src/hooks/useAgentChart.ts
```diff
@@ -0,0 +1,144 @@
...
+          'values.schema.json',
+          'values.yaml',
+        );
+
+        // 3. Fetch values.yaml (via the GitHub proxy for private-repo auth when
+        // it lives on raw.githubusercontent.com) and read agent.systemMessage.
+        if (valuesUrl) {
+          const valuesText = await fetchValues(baseUrl, fetchApi, valuesUrl);
+          if (valuesText) {
+            const parsed = load(valuesText) as
```
marians: **Robustness:** The hook is documented as "every step degrades gracefully", but `load(valuesText)` is unguarded: if the fetched `values.yaml` is malformed (or the `values.schema.json`→`values.yaml` replace fell through and JSON is parsed), `load()` throws, the whole query errors, and `version` falls back to the config floor (L124) even though the latest stable tag was already resolved. Wrap the values parse in try/catch so a bad prompt lookup doesn't also drop the resolved version (shown in the review summary and the manual `helm install` snippet).

plugins/agent-platform/src/hooks/useAgentChart.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Every step degrades gracefully: the version falls back to the configured
+ * floor and the prompt to '' (the user then types their own).
+ */
+export function useAgentChart(): AgentChart {
+  const configApi = useApi(configApiRef);
+  const discoveryApi = useApi(discoveryApiRef);
+  const fetchApi = useApi(fetchApiRef);
+
+  const ociUrl =
+    configApi.getOptionalString('agentPlatform.chart.ociUrl') ??
```
marians, line 36: **Reuse:** This `queryFn` re-implements the App Deployment flow's OCI resolution — tags → tag-manifest → values-schema annotation → values.yaml. `plugins/gs` already has `useHelmChartTags`/`useHelmChartTagManifest`/`useHelmChartValuesYaml` (backed by `containerRegistryApiRef`), `parseChartRef` (gs-common) in place of the local `parseOciRef`, and `fetchContent`/`gitHubApiRef` in place of `fetchValues`. It even redefines the same `VALUES_SCHEMA_ANNOTATION` constants. Reusing them avoids drift — the existing hooks already handle the 401/403/404 error naming this copy drops.

plugins/agent-platform/src/components/NewAgentReviewPage/NewAgentReviewPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+  // send the user back to fill it in.
+  if (!isComplete) {
+    return <Navigate to={newAgentLink ? newAgentLink() : '..'} replace />;
+  }
+
+  // The agent's resources are applied alongside the ModelConfig it uses — that
+  // namespace already exists and is where kagent watches.
+  const namespace = state.modelConfigNamespace!;
+
+  const { files, combinedManifest, valuesYaml, helmInstallCommand } =
```
marians: **Efficiency:** `composeManifests(...)` runs in the render body, so the full YAML composition re-runs and yields new `content` string identities on every re-render — and this page re-renders several times per deploy (`useAgentChart` resolving, `deployError`, `status.phase` authenticating→submitting→done). Each pass re-seeds the CodeMirror `YamlEditor`s with byte-identical content. Wrap it in `useMemo` keyed on the form state + ctx.

plugins/agent-platform/src/components/ModelConfigPicker/ModelConfigPicker.tsx
```diff
@@ -0,0 +1,171 @@
+import { Alert, FieldLabel, Flex, Text } from '@backstage/ui';
+import { makeStyles } from '@material-ui/core';
+import CheckCircleIcon from '@material-ui/icons/CheckCircle';
+import RadioButtonUncheckedIcon from '@material-ui/icons/RadioButtonUnchecked';
+
+import { useNewAgentForm } from '../NewAgentFormProvider';
+import { useModelConfigs } from '../ModelConfigsProvider';
+
+const useStyles = makeStyles(theme => ({
```
marians: **Duplication:** This ~45-line `makeStyles` block (card/selected/indicator/…) and the full-area selectable-`<button>` markup are duplicated near-verbatim in `SkillPicker.tsx` (L10). The only differences are the grid `minmax`, `radiogroup` vs `group`, and the indicator icons. Extract a shared `SelectableCard`/`useCardStyles` so a visual/focus/a11y change lands once — the two copies are already drifting in comment wording.

plugins/agent-platform/src/components/NewAgentFormProvider/NewAgentFormProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+        setState(prev => ({
+          ...prev,
+          modelConfigName: name,
+          modelConfigNamespace: namespace,
+        })),
+      setSystemMessage: systemMessage => {
+        setSystemMessageEdited(true);
+        setState(prev => ({ ...prev, systemMessage }));
+      },
+      applyDefaultSystemMessage: systemMessage =>
```
marians: **Simplification (minor):** `applyDefaultSystemMessage` must return the identical `prev` reference to avoid an infinite render loop, because `NewAgentPage` re-invokes it from an effect on every render — coupling the reducer's correctness to the caller's effect lifecycle. Seeding the default once (effect keyed on the resolved value, or a ref guard), or using the default as the textarea placeholder rather than writing it into state, removes both this guard and the `systemMessageEdited` flag.

marians:
Thanks for the thorough review — all points addressed in the latest 3 commits (through `9dabb151`). Summary:

**Correctness**
- **Invalid YAML from string concatenation** — manifests are now built as objects and serialized with `yaml.dump`; added a regression test with an irregularly-indented prompt.
- **Slug not validated** — rejected against the RFC1123 label regex (≤63 chars) before review, with an inline error instead of a late `kube:apply` failure.
- **`tree.truncated` ignored** — discovery returns a `truncated` flag (also set when a content read fails); surfaced as a warning in the picker.

**Robustness / efficiency**
- **Undifferentiated 500s** — `/agent-skills` forwards the upstream GitHub status (404/403/429) via a typed `GitHubApiError`, matching `/github/raw-content`.
- **Unbounded `Promise.all`** — bounded concurrency (5) + `allSettled`, so one 403/429 drops that skill rather than failing discovery.
- **Unguarded `load()`** — values.yaml parse wrapped in try/catch; a malformed file yields an empty prompt without dropping the resolved version.

**Reuse / quality**
- **Re-implemented OCI resolution** — `useAgentChart` now reuses the gs `useHelmChartTags` + `useHelmChartValuesYaml` hooks (exported from gs), dropping the duplicated fetch chain, `parseOciRef`, and the values-schema constants.
- **`composeManifests` in render body** — wrapped in `useMemo` keyed on form state + ctx.
- **Duplicated card styles/markup** — extracted a shared `SelectableCard` / `SelectableCardGrid` used by both pickers.
- **`applyDefaultSystemMessage` guard coupling** — removed it and the `systemMessageEdited` flag; the prompt is seeded once via a ref guard.

Verified end-to-end in a local browser: prompt pre-fills from the chart, submit-time validation lists what's missing, and the review page renders with valid `yaml.dump` manifests and the dynamically-resolved chart version. `tsc`/`lint` clean, tests green.

PR #1919, opened 2026-07-13
fix: restore Flux/Muster page tab navigation via custom PageLayout

marians:
### What does this PR do?

Registers a custom NFS `PageLayout` swappable component (`GSPageLayout`) that renders the bui `PluginHeader`, so page headers and — crucially — the tabs rendered by `PageBlueprint` for `SubPageBlueprint` sub-pages (Flux **List view**/**Tree view**, the Muster section) work correctly.

The app never overrode the `PageLayout` swappable component, so it fell back to the upstream stub whose tabs are plain **relative** `<a href="list">` anchors with **no active state**. On the `/flux` page this produced three regressions:

1. On `/flux/list`, the **List view** tab was not highlighted.
2. Clicking **Tree view** navigated to `/flux/list/tree` (relative append) instead of `/flux/tree` — the page stayed on the list view.
3. Clicking **List view** again went to `/flux/list/tree/list`.

`GSPageLayout` resolves each sub-page's relative `path` to an absolute href and uses segment-aware prefix matching for the active tab (the same mechanism the working Clusters/Deployments sections use via `useLayoutTabs`). Pages that render their own `PluginHeader` pass `noHeader: true`, which the component honors.

Tab navigation on the Flux and Muster pages works again: tabs switch views correctly and the active tab is highlighted.

Regression introduced by #1712 ("migrate pages to the new frontend system page header"), which switched Flux from a classic `RoutedTabs`-based `FluxPageLayout` to `SubPageBlueprint` tabs but never added the assumed bui `PageLayout` swap. The upstream `PageLayout` docstring explicitly invites this override: *"Apps can override this with a custom implementation (e.g., using @backstage/ui)."*

Verified with a unit test (`GSPageLayout.test.tsx`) that mounts the component at a splat route and asserts absolute tab hrefs, active-tab state, and the `noHeader` path.

Yes — done in this PR. Added a "Page headers and tabs (New Frontend System)" section to `docs/ui.md` and a pointer from the `ui` Claude Code skill.

Not applicable — the change is in `packages/app`, which is released via the app image/Helm chart flow (git-cliff from the PR title), not the Changesets-published `@giantswarm/backstage-plugin-*` packages.

### DIFF
--- a/packages/app/src/modules/app/AppOverrides.tsx
+++ b/packages/app/src/modules/app/AppOverrides.tsx
@@ -10,10 +10,12 @@ import {
   analyticsApiRef,
   ApiBlueprint,
   AppRootElementBlueprint,
+  PageLayout,
 } from '@backstage/frontend-plugin-api';
 import {
   IconBundleBlueprint,
   SignInPageBlueprint,
+  SwappableComponentBlueprint,
   ThemeBlueprint,
 } from '@backstage/plugin-app-react';
 import DarkIcon from '@material-ui/icons/Brightness2';
@@ -53,6 +55,7 @@ import {
 } from '../../assets/icons/CustomIcons';
 import { BrandingFavicon } from '../branding';
 import { DarkThemeProvider, LightThemeProvider } from './customThemes';
+import { GSPageLayout } from './GSPageLayout';
 
 // The Grafana plugin is a legacy plugin whose API factory is not
 // auto-registered in the NFS. Extract it and provide via ApiBlueprint.
@@ -180,6 +183,21 @@ export const appOverrides = createFrontendModule({
       params: {
         element: <BrandingFavicon />,
       },
+    }),
+    /**
+     * Override the NFS `PageLayout` swappable component so every page header —
+     * and, crucially, the sub-page tabs rendered by `PageBlueprint` (flux
+     * list/tree, muster, …) — uses the bui `PluginHeader` with absolute tab
+     * links and active-state highlighting, instead of the upstream stub whose
+     * relative `<a href>` tabs break navigation.
+     */
+    SwappableComponentBlueprint.make({
+      name: 'page-layout',
+      params: define =>
+        define({
+          component: PageLayout,
+          loader: () => GSPageLayout,
+        }),
     }),
     /**
      * Override the built-in `theme:app/light` and `theme:app/dark` extensions
--- a/packages/app/src/modules/app/GSPageLayout.tsx
+++ b/packages/app/src/modules/app/GSPageLayout.tsx
@@ -1,0 +1,65 @@
+import { useLocation, useParams } from 'react-router-dom';
+import type { PageLayoutProps } from '@backstage/frontend-plugin-api';
+import { PluginHeader } from '@backstage/ui';
+import type { HeaderTab } from '@backstage/ui';
+
+/**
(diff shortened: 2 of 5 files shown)


packages/app/src/modules/app/GSPageLayout.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const { pathname } = useLocation();
+  const params = useParams();
+
+  // The page is mounted at a splat route (e.g. `/flux/*`), so the base path is
+  // the current pathname with the matched remainder removed. Sub-page tab paths
+  // (`list`, `tree`) are relative to that base.
+  const splat = params['*'] ?? '';
+  let basePath =
+    splat && pathname.endsWith(splat)
```
marians: `pathname.endsWith(splat)` compares the **raw** `location.pathname` against the **decoded** `useParams()['*']` value. React Router decodes route params but leaves `location.pathname` percent-encoded, so for a sub-route segment containing encoded chars (e.g. a space or unicode → `%20`), `endsWith` returns `false`, `basePath` falls back to the full deep pathname, and every tab href becomes wrong (`/muster/workflows/my%20flow/list` instead of `/muster/list`).

In practice sub-route ids here are DNS-safe k8s names, so this is currently unreachable — flagging as a latent edge case, not a live bug. If you want to be safe, decode the pathname before comparing (or derive the base by stripping the known page `path` prefix instead of the splat suffix).

packages/app/src/modules/app/AppOverrides.tsx
```diff
@@ -181,6 +184,21 @@ export const appOverrides = createFrontendModule({
...
+     * list/tree, muster, …) — uses the bui `PluginHeader` with absolute tab
+     * links and active-state highlighting, instead of the upstream stub whose
+     * relative `<a href>` tabs break navigation.
+     */
+    SwappableComponentBlueprint.make({
+      name: 'page-layout',
+      params: define =>
+        define({
+          component: PageLayout,
+          loader: () => GSPageLayout,
```
marians, line 199: The actual override mechanism this PR is about — the `SwappableComponentBlueprint` registration and the `PageBlueprint`→`PageLayout`→`noHeader` wiring — is **not** exercised by any test. `GSPageLayout.test.tsx` renders the component directly, so a mis-wired `component`/`loader`/`name` here (or a page that forgets `noHeader: true` and double-headers) would still pass the unit test and only surface in the running app. Not a bug in the current code (it type-checks and you verified manually), just noting the coverage gap since the whole regression class lives in this wiring, not in `GSPageLayout` itself.

packages/app/src/modules/app/GSPageLayout.tsx
```diff
@@ -0,0 +1,65 @@
...
+      ? pathname.slice(0, pathname.length - splat.length)
+      : pathname;
+  basePath = basePath.replace(/\/+$/, '');
+
+  const headerTabs: HeaderTab[] | undefined = tabs?.map(tab => {
```
marians, line 44: Minor: `basePath`/`headerTabs` are computed unconditionally, but when `noHeader` is set (the clusters/deployments/ai-chat/home pages, i.e. the common case) they are thrown away by the early `return <>{children}</>` below. Moving the `if (noHeader) return ...` above this block avoids the wasted `tabs.map` on every render of those pages. Trivial, take it or leave it.

PR #1931, opened 2026-07-15
feat: show blocking ancestor Kustomization on deployment details page

weatherhog:
### What does this PR do?

When a deployment is managed through GitOps and an *ancestor* Flux Kustomization in its chain is suspended or has `Ready=False`, changes to the deployment silently stop being applied. Developers looking at the deployment see a healthy-looking page and have no way to know a parent Kustomization is the culprit.

This PR adds a warning banner to the deployment details overview that:

- walks the Kustomization ancestor chain via the `kustomize.toolkit.fluxcd.io/name`/`namespace` labels Flux stamps on every applied object (no inventory parsing needed, works even when `.status.inventory` is unpopulated),
- names the **topmost** blocked Kustomization (usually the root cause) and shows its Ready condition message,
- lists any further blocked Kustomizations in the chain,
- links to the Flux overview tree with the blocking Kustomization's details pane open.

New exports from `flux-react`: `FluxBlockedByCard`, `findKustomizationAncestors`, `findBlockedAncestors`. The tree view's details-pane id/prefix constants were extracted to `FluxResourcesTreeView/constants.ts` so the deep link can be built outside the tree view.

The banner renders nothing while loading and when the whole chain is healthy. The walk is cycle-guarded and stops at self-referencing Kustomizations (e.g. `flux-system`).

App developers can immediately see *why* their deployment isn't updating ("I changed the version — why doesn't it apply?") instead of having to know the Flux resource hierarchy and hunt for the failing parent Kustomization among hundreds.

Addresses the Backstage part of https://github.com/giantswarm/giantswarm/issues/35859 (Flux troubleshooting case — developers can't trace a stuck deployment upward to the failing parent Kustomization). This is the first of a small series of PRs; follow-ups will add a failing-paths rollup/filter and failure-message search in the Flux tree.

No.


plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+      status="warning"
+      icon
+      title="Deployment updates blocked"
+      description={
+        <>
+          <Typography variant="body2">
+            Kustomization {formatKustomizationRef(rootCauseKustomization)}{' '}
+            {rootCause.reason === 'suspended'
+              ? 'is suspended, so changes to this deployment are not applied.'
+              : 'is failing, so changes to this deployment are not applied:'}
```
marians: **Dangling colon when a failing ancestor has an empty Ready-condition message.**

The `not-ready` branch always ends the sentence with a colon (`…not applied:`) on the assumption that a `ConditionMessage` will follow. But `findBlockedAncestors` stores `readyCondition.message` verbatim, and Flux can set an empty message (the mock helper here even defaults it to `''`). When `rootCause.message` is falsy the `ConditionMessage` block at line 97 is skipped, leaving the banner reading *"… is failing, so changes to this deployment are not applied:"* with nothing after the colon.

```suggestion
            {rootCause.reason === 'suspended'
              ? 'is suspended, so changes to this deployment are not applied.'
              : rootCause.message
                ? 'is failing, so changes to this deployment are not applied:'
                : 'is failing, so changes to this deployment are not applied.'}
```

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const { getRoute } = useDetailsPane(FLUX_RESOURCE_PANE_ID, {
+    prefix: FLUX_RESOURCE_PANE_PREFIX,
+  });
+
+  if (isLoading || blockedAncestors.length === 0) {
+    return null;
+  }
+
+  // The last entry is the topmost blocked ancestor and usually the root cause.
+  const rootCause = blockedAncestors[blockedAncestors.length - 1];
```
marians: **"Topmost blocked ancestor = root cause" breaks when a healthy Kustomization sits between two blocked ones.**

`rootCause` is unconditionally the last (topmost) entry of `blockedAncestors`. For a linear failing chain that's correct, but consider `deployment ← A (Ready=False) ← B (Ready=True) ← C (Ready=False)`: the walk yields `[A, C]`, so the banner names **C** as the root cause and demotes **A** — the ancestor actually adjacent to the deployment — into the "also blocked" list. Because B reconciled successfully, C's failure does not propagate down to A, so A (not C) is the block that's actually stopping this deployment's updates. Low severity / unusual topology, but the "usually the root cause" heuristic silently mis-attributes here.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const rootCauseKustomization = rootCause.kustomization;
+
+  const rootCauseUrl = fluxOverviewRoute
+    ? `${getRoute(fluxOverviewRoute(), {
+        cluster: installationName,
+        kind: Kustomization.kind.toLowerCase(),
+        name: rootCauseKustomization.getName(),
+        namespace: rootCauseKustomization.getNamespace(),
+      })}&cluster=${installationName}`
```
marians: **Deep link is assembled by hand-concatenating an un-encoded `&cluster=` onto `getRoute`'s output.**

`getRoute(...)` already produces a fully-encoded query string (via `URLSearchParams`), and then `&cluster=${installationName}` is appended raw. This works today only because (a) cluster names are DNS-safe so encoding is moot, (b) `getRoute` happens to emit the pane's cluster under the `sr-cluster` key so there's no collision with the unprefixed `cluster` the `ClusterPicker` reads, and (c) `ClusterPicker`'s `urlParameterName` defaults to `'cluster'`. All three are invisible couplings — if any changes, the link silently breaks. Consider a single builder that emits both the pane params and the cluster filter, or at minimum encoding the value.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+export function FluxBlockedByCard({
+  deployment,
+  installationName,
+  fluxOverviewRouteRef,
+}: FluxBlockedByCardProps) {
+  const {
+    resources: kustomizations,
+    errors,
+    isLoading,
+  } = useResources(installationName, Kustomization);
```
marians: **Every GitOps deployment overview now issues a cluster-wide `Kustomization` LIST.**

`useResources(installationName, Kustomization)` lists *all* Kustomizations in the installation (all namespaces) purely to walk one ancestor chain, on every deployment-details view. On a management cluster with hundreds/thousands of Kustomizations this is a non-trivial fetch that isn't shared with the Flux-overview page's own Kustomization query (different provider/options → different react-query key). Given the chain is typically 1–3 hops, consider whether the already-cached overview data can be reused, or whether this needs field/namespace scoping.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+ * Warns when an ancestor Flux Kustomization of the given deployment is
+ * suspended or failing and therefore blocking updates to the deployment.
+ * Renders nothing while loading or when the whole chain is healthy.
+ */
+export function FluxBlockedByCard({
+  deployment,
+  installationName,
+  fluxOverviewRouteRef,
+}: FluxBlockedByCardProps) {
+  const {
```
marians, line 54: **The `isManagedByFlux` guard lives in the consumer, not in the exported component.**

`FluxBlockedByCard` is a new public export of `flux-react`, but it has no internal guard — it relies on `DeploymentBlockedByCard` (in the `gs` plugin) to check `isManagedByFlux` before rendering it. Any other caller of the exported `FluxBlockedByCard` that forgets that gate will trigger a cluster-wide Kustomization LIST and ancestor walk for a resource that isn't Flux-managed. Since the flux-domain knowledge belongs to flux-react, the `isManagedByFlux(deployment)` early-return would be more robust inside `FluxBlockedByCard` itself (it already returns `null` in the not-blocked case anyway).

marians (changes requested):
See comments

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
weatherhog: Fixed — the colon is now conditional on a non-empty message.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
weatherhog: Agreed, the unconditional "topmost" pick mis-attributed in the split-chain topology. `BlockedAncestor` now carries its `chainIndex`, and a new `selectBlockingRootCause` helper headlines the topmost entry of the *contiguous* blocked run nearest the deployment: in a full cascade that's still the topmost ancestor (the real root cause), but in `deployment ← A(F) ← B(T) ← C(F)` it now names A, with C listed under "also blocked in the chain". Covered by tests for both topologies.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
weatherhog: Reworked: the link is now built by parsing `getRoute`'s query string into `URLSearchParams`, setting `cluster` there, and re-serializing — so the value is encoded and the separator handling is robust. The `sr-` prefix / `cluster` key couplings remain (they're the tree page's public URL contract), but there's no raw string concatenation anymore.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
weatherhog: Fair point. Mitigations in place now: the query is `enabled` only for Flux-managed deployments (guard moved into the component, see other thread) and gets `staleTime: 60s`, so navigating between deployments on the same installation reuses one LIST from the react-query cache (the key is shared across all deployment pages of an installation) instead of refetching per view/focus. A per-hop GET chain isn't hook-friendly (chain length is only known at runtime), and sharing the Flux-overview provider's query would couple the deployment page to that provider — happy to look into a shared Kustomizations query hook as a follow-up if the LIST still turns out to be too heavy on large MCs.

plugins/flux-react/src/components/FluxBlockedByCard/FluxBlockedByCard.tsx
weatherhog: Done — `FluxBlockedByCard` now checks `isManagedByFlux(deployment)` itself (and gates the Kustomization LIST on it via `enabled`), so external callers can't trigger the fetch for non-Flux resources. The redundant guard in gs's `DeploymentBlockedByCard` is removed.

marians (approved):

PR #1932, opened 2026-07-15
feat(agent-platform): tabbed section with Agents and MCP Servers tabs

marians:
### What does this PR do?

Restructures `/agent-platform` into a tabbed section and moves the muster
("MCP Servers") experience underneath it.

- **agent-platform** becomes a tabbed page (`PageBlueprint` + `SubPageBlueprint`)
  with two top-level tabs: **Agents** and **MCP Servers**. The Agents tab is a
  stub landing plus the existing create flow (`/agent-platform/agents/new` and
  `.../new/review`).
- **muster** is no longer a standalone page; it contributes a `SubPageBlueprint`
  attached to `page:agent-platform`, mounted at `/agent-platform/muster`, with
  its four views (Dashboard, Servers, Workflows, Tool explorer) rendered as a
  second-level BUI tab row. Its route refs still resolve, so all internal links
  keep working.
- **ui-react** gains a small `PageHeaderActions` slot so routed tab content can
  inject header buttons (the create flow's Cancel / Review) into the single
  section header instead of rendering a second header.
- **app**: the nav item is renamed to "Agent Platform" (the standalone
  "MCP Servers" item is removed); `GSPageLayout` hosts the header-actions slot;
  telemetry page mappings are updated.

- New **Agent Platform** section in the main menu (replaces the **MCP Servers**
  item). It has **Agents** and **MCP Servers** tabs.
- URLs change:
  - `/muster/*` → `/agent-platform/muster/*` (bare `/agent-platform/muster`
    redirects to `.../dashboard`).
  - The "MCP servers" view is renamed **Servers**:
    `/agent-platform/muster/mcp-servers` → `/agent-platform/muster/servers`.
- The agent-platform section is still disabled by default and enabled
  per-installation via `app.extensions`.

### How does it look like?

```
Agent Platform                       [Cancel] [Review & deploy]   (create flow)
┌────────┬─────────────┐
│ Agents │ MCP Servers │                          ← level-1 tabs
└────────┴─────────────┘
 Dashboard · Servers · Workflows · Tool explorer   ← level-2 tabs (MCP Servers)
```

Part of the ongoing Agent Platform work: bringing the agent create flow and the
muster management UI together under one section with a consistent header and
navigation.

Yes — done in this PR (`plugins/muster/README.md` route references updated).

- [x] A changeset describing the change and affected packages was added
      (`@giantswarm/backstage-plugin-agent-platform`,
      `@giantswarm/backstage-plugin-muster`,
      `@giantswarm/backstage-plugin-ui-react`).


plugins/ui-react/src/components/PageHeaderActions/PageHeaderActions.tsx
```diff
@@ -0,0 +1,56 @@
...
+// Register header actions from routed content for as long as the calling
+// component is mounted; they are cleared automatically on unmount. Pass the
+// freshly-built element on every render — the setter is stable, so this does
+// not re-render the caller, and the header always reflects the latest closures.
+export function useProvidePageHeaderActions(actions: ReactNode): void {
+  const setActions = useContext(PageHeaderActionsSetContext);
+  useEffect(() => {
+    setActions(actions);
+    return () => setActions(null);
+  });
```
marians: `useProvidePageHeaderActions` has no dependency array, so the effect fires on **every render** of the consuming component (`NewAgentPage` / `NewAgentReviewPage`). Because `actions` is a freshly-built element on each render, `setActions(actions)` always receives a new reference and re-renders the whole `PluginHeader` (title, icon, and the level-1 tab strip) on every keystroke in the create form.

This isn't a correctness bug — the two-context split keeps the content itself from re-rendering, and freshness is intended so the closures stay current — but re-rendering the header + tabs on each character typed is avoidable churn. Consider comparing against the previously-set node (or accepting a stable render-prop) so the header only updates when the actions actually change.

plugins/muster/src/components/MusterSection/MusterSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+function useBasePath(): string {
+  const { pathname } = useLocation();
+  const params = useParams();
+  const splatSegments = (params['*'] ?? '').split('/').filter(Boolean).length;
+  const segments = pathname.split('/').filter(Boolean);
+  return `/${segments
+    .slice(0, Math.max(0, segments.length - splatSegments))
+    .join('/')}`;
+}
+
```
marians: `useBasePath` here is a verbatim copy of the splat→base-path logic in `packages/app/src/modules/app/GSPageLayout.tsx` (the comment even says "Same approach as the app's GSPageLayout"). Two copies of the same percent-encoding-safe segment math will drift independently. Consider extracting it into `ui-react` (both packages already depend on it) and importing it in both places.

packages/app/src/App.tsx
```diff
@@ -56,9 +56,9 @@ const app = createApp({
     fluxPluginOverrides,
     aiChatPlugin,
     aiChatPluginOverrides,
+    agentPlatformPlugin,
```
marians, line 65: Moving `agentPlatformPlugin` ahead of `musterPlugin` here looks load-bearing but is undocumented: the "Agents" tab is contributed by `agent-platform` while "MCP Servers" is attached externally by `muster` (`attachTo: page:agent-platform`), and the level-1 tab order — plus which tab a bare `/agent-platform` lands on — appears to depend on this registration order. If that's the reason, a one-line comment here would prevent a future reorder from silently flipping the tabs. (If ordering is actually controlled elsewhere, e.g. by `app.extensions` order, ignore.)

plugins/muster/src/components/MusterSection/MusterSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const MusterSection = () => {
+  const basePath = useBasePath();
+
+  return (
+    <MusterProviders>
+      {/* Inset the tab strip by the page gutter (bui space-5 = 20px) so it lines
+          up with the level-1 header tabs and the content below, which the
+          PluginHeader / Content apply automatically but this bare Tabs row does
+          not. */}
+      <Box px="5">
```
marians, line 40: Minor/altitude: the `px="5"` gutter is hand-tuned to match the horizontal padding that `PluginHeader`/`Content` apply automatically. If the bui gutter token ever changes, this level-2 tab strip silently falls out of alignment with the level-1 header tabs and the content below. Not blocking, but worth a note that this value is coupled to bui's internal gutter rather than derived from it.

marians (commented):
Code review (high effort) — a clean, well-commented refactor. No correctness/crash bugs found: telemetry switch ordering (muster before agent-platform) is correct, route refs rebind correctly under the new SubPageBlueprint structure, all deps are added, and no stale `/muster` / `/agents` / `mcp-servers` route references remain. The findings below are all quality/altitude observations, none blocking.

PR #1933, opened 2026-07-15
feat: roll up failing descendant status in Flux tree and add failing-only filter

weatherhog:
### What does this PR do?

Makes failing resources findable in the Flux resources tree, even among hundreds of Kustomizations:

- **Failing-status rollup**: `KustomizationTreeBuilder` now computes `isFailing` (the node's own `Ready=False`) and `hasFailingDescendants` (post-order aggregation) for every tree node. Parents with failing resources anywhere beneath them show a warning indicator with a "Contains failing resources" tooltip — visible while the subtree is collapsed, so a failing leaf can't hide.
- **"Failing only" status filter**: a new Status picker in the tree view prunes the tree to only the paths that lead to failing resources (`filterTreeToFailingPaths`). This is the UI equivalent of `flux get kustomizations -A --status-selector ready=false`, but with the ancestor paths preserved for context. Search and match navigation operate on the filtered tree.

Design note for review: suspended resources are intentionally *not* counted as failing in the rollup (consistent with `getAggregatedStatus`, which treats suspended as `inactive`). If you'd rather have suspended count as blocking here too, that's a one-line change in `isResourceFailing`.

Also adds the first unit tests for `KustomizationTreeBuilder` (inventory-based fixtures, rollup assertions, cycle termination).

Developers looking at a wall of green root Kustomizations can now spot the one with a broken descendant immediately, and collapse the tree to just the failing paths instead of expanding hundreds of nodes.

Addresses the Backstage part of https://github.com/giantswarm/giantswarm/issues/35859 (developers can't find which of hundreds of Kustomizations is blocking their deployment). Companion to #1931 (blocked-by banner on deployment details); a follow-up will make tree search match failure messages.

No.


plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
```diff
@@ -34,8 +34,16 @@ export type KustomizationTreeNodeData = {
...
   hasChildren: boolean;
   hasChildrenInCompactView: boolean;
+  isFailing: boolean;
+  hasFailingDescendants: boolean;
 };
 
+function isResourceFailing(
+  resource?: KustomizationTreeNodeData['resource'],
+): boolean {
+  return resource?.findReadyCondition()?.status === 'False';
```
marians, line 52: **Suspended resources are counted as failing, contradicting the design note.**

`isResourceFailing` only looks at the Ready condition:

```ts
return resource?.findReadyCondition()?.status === 'False';
```

But a suspended resource keeps its last `Ready` condition frozen — if it was `Ready=False` when it got suspended, `findReadyCondition().status` is still `'False'`. So a suspended-while-failing HelmRelease/Kustomization is treated as failing here, even though:

- the leaf's own `ResourceStatus` renders it as **"Suspended"** (`elStatus: 'aborted'`), *not* an error, because `isSuspended` overrides `readyStatus`; and
- the PR description explicitly states *"suspended resources are intentionally not counted as failing (consistent with `getAggregatedStatus`, which treats suspended as inactive)"* — but `getAggregatedStatus` returns `inactive` for `isSuspended` **before** checking `readyStatus`, whereas this helper never checks suspension.

Effect: every ancestor of a suspended (formerly-failing) resource lights up the "Contains failing resources" warning icon, and the resource is pulled into the "Failing only" view — the exact false positives the design note says to avoid.

```suggestion
function isResourceFailing(
  resource?: KustomizationTreeNodeData['resource'],
): boolean {
  if (resource?.isSuspended()) {
    return false;
  }
  return resource?.findReadyCondition()?.status === 'False';
}
```

weatherhog: Good catch — you're right, the frozen Ready condition on suspended resources made the rollup contradict both the design note and how the leaf itself renders. Applied your suggestion and added a test covering the suspended-while-failing case (suspended leaf → `isFailing` false, ancestors unmarked).

marians (approved):

PR #1934, opened 2026-07-15
feat: match failing condition messages in Flux tree search

weatherhog:
### What does this PR do?

Extends the Flux resources tree search to also match the **Ready condition messages of failing resources**, in addition to resource names.

Why this matters: when a Kustomization fails to apply a resource (invalid CR, schema change, dry-run failure), the *name of that resource* appears in the Kustomization's error message — but the resource itself doesn't exist anywhere in the tree, so name search can't find it. With this change, a developer wondering "why is my new service not showing up?" can type the service's name into the tree search and land directly on the Kustomization whose error mentions it.

Messages of healthy resources (`Ready=True`, e.g. "Applied revision main@sha1:...") are deliberately excluded — they would match almost any query.

Implementation: `findMatchingNodes` extracted from `useTreeSearch` into its own module with unit tests; the match predicate checks the failing condition message; existing highlight, auto-expand, and match navigation work unchanged. The search field gets a "Name or failure message" placeholder.

Out of scope (deliberately): a flat "failing resources" list page — the list view already has a status picker and table search.

Searching the Flux tree for a resource that never got created finds the Kustomization that fails to apply it.

Third PR for the Backstage part of https://github.com/giantswarm/giantswarm/issues/35859, following #1931 (blocked-by banner) and #1933 (failing-status rollup + filter).

No.


plugins/flux-react/src/hooks/useTreeSearch/findMatchingNodes.ts
```diff
@@ -0,0 +1,59 @@
+import { KustomizationTreeNode } from '../../components/FluxOverview/utils/KustomizationTreeBuilder';
+
+export type SearchResult = {
+  matches: string[];
+  pathsToExpand: Set<string>;
+};
+
+function getFailureMessage(node: KustomizationTreeNode): string | undefined {
+  const readyCondition = node.nodeData.resource?.findReadyCondition();
+  if (readyCondition?.status !== 'False') {
```
marians, line 18: **Failing state is inferred from the raw `Ready` condition, bypassing the canonical Flux status (`isSuspended`).**

`getFailureMessage` treats any resource with `Ready.status === 'False'` as a searchable failure. But Flux stops reconciling **suspended** resources, so a resource that was failing when it got suspended keeps a stale `Ready=False` condition with an old error message. This message will now be matched as if it were a live failure, surfacing a resource the rest of the UI tracks separately as suspended (`FluxResourceStatusMixin` already computes `isSuspended`/`isReconciling`/`readyStatus` via `getOrCalculateFluxStatus()`).\n\nConsider reusing the canonical `FluxResourceStatus` (e.g. skip when `isSuspended`) instead of re-deriving "failing" from the raw condition, so search stays consistent with the failing-status rollup added in #1933.

plugins/flux-react/src/hooks/useTreeSearch/findMatchingNodes.ts
```diff
@@ -0,0 +1,59 @@
...
+    for (const node of nodes) {
+      if (compactView && !node.displayInCompactView) {
+        continue;
+      }
+
+      const nodeName = node.nodeData.name.toLowerCase();
+      const failureMessage = getFailureMessage(node)?.toLowerCase();
+      if (
+        nodeName.includes(normalizedQuery) ||
+        failureMessage?.includes(normalizedQuery)
```
marians, line 58: **Substring matching against free-form error text will surface noisy matches on generic words.**

With `MIN_SEARCH_LENGTH = 3`, a short query such as `run`, `not`, `crd`, or `err` now matches *every* failing resource whose message contains that substring (`dry-run failed`, `not found`, `the CRD is invalid`, `error`…), not just the resource name the feature targets. The changeset addresses noise from *healthy* messages, but generic-word noise from failing messages remains. Not blocking — an inherent trade-off of substring search — but worth a comment noting the limitation, or scoping the match more tightly (e.g. only the object-reference portion Flux embeds).

plugins/flux-react/src/hooks/useTreeSearch/findMatchingNodes.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    ];
+
+    const result = findMatchingNodes(tree, 'my-new-service', false);
+
+    expect(result.matches).toEqual([
+      'test-cluster-Kustomization-flux-system-apps',
+    ]);
+  });
+
+  it('does not match messages of healthy resources', () => {
```
marians, line 216: **Test-coverage gaps for the new predicate.**

The suite covers `status: 'True'` and `status: 'False'`, but not:\n- `status: 'Unknown'` (reconciling / progressing) — the predicate `readyCondition?.status !== 'False'` deliberately treats `Unknown` like healthy and excludes it; that branch is currently untested.\n- A resource whose `Ready` condition has `message === undefined` (allowed at runtime despite the type) — worth a test to lock in that `getFailureMessage(...)?.toLowerCase()` stays crash-safe.\n- A non-Kustomization failing resource (e.g. `HelmRelease`), since `nodeData.resource` can be any Flux object type and the feature applies to all of them.

marians (approved):
Some possible improvements, nothing blocking

plugins/flux-react/src/hooks/useTreeSearch/findMatchingNodes.ts
weatherhog: Fixed — `getFailureMessage` now returns nothing for suspended resources (their frozen Ready condition can describe a stale failure), matching the #1933 rollup semantics. I kept the raw-condition read rather than `getOrCalculateFluxStatus()` to stay a pure function over the node (no status-manager singleton), but the suspended/failing precedence is now the same.

plugins/flux-react/src/hooks/useTreeSearch/findMatchingNodes.ts
weatherhog: Agreed it's inherent to substring search — documented the trade-off in the `findMatchingNodes` doc comment (short generic queries match every failing resource; accepted because failing resources are few and parsing the object-reference portion would be fragile across Flux error formats).

plugins/flux-react/src/hooks/useTreeSearch/findMatchingNodes.test.ts
weatherhog: Added tests for all three gaps: `Ready=Unknown` (reconciling) messages are not matched, a failing condition without a message doesn't crash and still matches by name, and a failing HelmRelease's message is matched (plus one for the new suspended-resource exclusion).

PR #1935, opened 2026-07-15
feat: explain Flux errors with AI chat

weatherhog:
### What does this PR do?

Makes cryptic Flux error messages explainable with one click via the existing AI chat:

- **New `buildExplainErrorMessage` prompt builder in ai-chat-react**: embeds a resource's failing condition message plus context (kind, name, namespace, management cluster, condition reason, revision), truncates very long messages, and asks the AI for a plain-language explanation, likely root cause, and concrete next steps. It also tells the AI it may read the resource live (the ai-chat backend already has muster MCP tools for that).
- **Flux resource card (tree/details pane)**: the existing "Troubleshoot with AI" button now sends the actual Ready condition error message instead of only asking the AI to look the resource up — the answer is immediate and specific. (Deliberately reused the existing button rather than adding a second AI button to the same card.)
- **Deployment details page**: the HelmRelease "Conditions" card gets an "Explain this error" button inside expanded failing conditions (`status === 'False'` or `Stalled`) — there was no AI affordance there before.

Graceful degradation is inherited from `AIChatButton`: on installations without ai-chat enabled (`aiChatApiRef` not registered), the buttons render nothing.

When a Kustomization or HelmRelease shows an error like `dry-run failed: error validating data...`, users can get a plain-language explanation and suggested next steps without knowing Flux internals — addressing the "condition messages can be quite cryptic" part of the underlying issue.

Fourth and final PR for the Backstage part of https://github.com/giantswarm/giantswarm/issues/35859, following #1931 (blocked-by banner), #1933 (failing rollup + filter), and #1934 (failure-message search).

No.


plugins/gs/src/components/deployments/deployment-details/DeploymentOverview/DeploymentStatusCard/HelmReleaseConditions/HelmReleaseConditions.tsx
```diff
@@ -129,6 +138,29 @@ const ConditionCard = ({
...
+                  items={[
+                    {
+                      message: buildExplainErrorMessage({
+                        kind: 'HelmRelease',
+                        name: helmrelease.getName(),
+                        namespace: helmrelease.getNamespace(),
+                        cluster: helmrelease.cluster,
+                        message: condition.message,
+                        reason: condition.reason,
+                        revision: helmrelease.getLastAppliedRevision(),
```
marians: **Failing revision vs. last-applied revision.** `getLastAppliedRevision()` returns the chart version of the most recent *successful* Helm release (from `status.history`), not the revision that failed. So for a failing HelmRelease the AI is told it is "failing at revision '<last-good-version>'", which is misleading — the failure actually pertains to `status.lastAttemptedRevision`. `getLastAttemptedRevision()` matches the failure context here (and the `revision` field is documented as "Source revision the failure was reported at").

```suggestion
                        revision: helmrelease.getLastAttemptedRevision(),
```

plugins/gs/src/components/deployments/deployment-details/DeploymentOverview/DeploymentStatusCard/HelmReleaseConditions/HelmReleaseConditions.tsx
```diff
@@ -81,6 +87,9 @@ const ConditionCard = ({
     conditionHeadline = `HelmRelease Not ${condition.type.toLowerCase()}`;
   }
 
+  const isFailing =
+    condition.status === 'False' || condition.type === 'Stalled';
```
marians: **`isFailing` flags a not-stalled `Stalled` condition.** `condition.type === 'Stalled'` is treated as failing regardless of status. If a `Stalled` condition is reported with `status === 'False'` (i.e. *not* stalled — a healthy state), `isFailing` is still `true`, so the red "Explain this error" button appears on a healthy condition. The intent is "stalled" = `Stalled` with `status === 'True'`. (This mirrors the existing icon/headline logic above, so consider fixing both, or leaving a note that this is deliberate.)

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
```diff
@@ -152,6 +155,27 @@ export const ResourceCard = ({
...
+
+  let aiChatMessage: string;
+  if (readyStatus === 'False' && failureMessage) {
+    aiChatMessage = buildExplainErrorMessage({
+      kind,
+      name,
+      namespace,
+      cluster,
+      message: failureMessage,
+      reason: resource?.findReadyCondition()?.reason,
```
marians: **Redundant `findReadyCondition()` calls.** `resource?.findReadyCondition()` is invoked up to three times per render (line 160 for `failureMessage`, and again on line 171 for `reason`; `findReadyCondition` re-fetches and re-scans `status.conditions` each time). Compute the ready condition once and reuse it:

```ts
const readyCondition =
  readyStatus === 'False' ? resource?.findReadyCondition() : undefined;
const failureMessage = readyCondition?.message;
// ...
  reason: readyCondition?.reason,
```

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
```diff
@@ -152,6 +155,27 @@ export const ResourceCard = ({
...
+    });
+  } else if (readyStatus === 'False') {
+    aiChatMessage = `Please read the ${kind} resource named '${name}' in namespace '${namespace}' on management cluster '${cluster}' and help me understand why it is not in a Ready state.`;
```
marians: **`namespace 'undefined'` in the fallback branches.** `namespace` is optional (`namespace?: string` on `ResourceCardProps`). The `buildExplainErrorMessage` path correctly omits the namespace clause when it is absent, but these two fallback templates interpolate it unconditionally, producing `...in namespace 'undefined' on management cluster...` when a resource has no namespace. Consider guarding the clause the same way `buildExplainErrorMessage` does (or routing these fallbacks through the same builder).

plugins/ai-chat-react/src/utils/buildExplainErrorMessage.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Builds an AI chat prompt asking for a plain-language explanation of a
+ * resource's error message and suggested next steps.
+ */
+export function buildExplainErrorMessage(context: ExplainErrorContext): string {
+  const { kind, name, namespace, cluster, reason, revision } = context;
+
+  const message =
+    context.message.length > MAX_ERROR_MESSAGE_LENGTH
+      ? `${context.message.slice(0, MAX_ERROR_MESSAGE_LENGTH)}…`
```
marians, line 27: **Error message can break the Markdown code fence.** The message is embedded inside a triple-backtick fence. If the error text itself contains a ``` sequence (or the 4000-char truncation cuts inside one), the fence terminates early and the remainder of the prompt renders as prose, degrading the context sent to the AI. Consider stripping/escaping backtick runs, or using a longer/uncommon fence delimiter derived from the content.

marians (changes requested):

plugins/gs/src/components/deployments/deployment-details/DeploymentOverview/DeploymentStatusCard/HelmReleaseConditions/HelmReleaseConditions.tsx
weatherhog: Right — `getLastAppliedRevision()` is the last *successful* release, which is misleading in a failure context. Switched to `getLastAttemptedRevision()` as suggested.

plugins/gs/src/components/deployments/deployment-details/DeploymentOverview/DeploymentStatusCard/HelmReleaseConditions/HelmReleaseConditions.tsx
weatherhog: Fixed properly: extracted an `isFailingCondition` helper (`Stalled` fails only with `status === 'True'`; everything else with `status === 'False'`) and used it for the button, the error/check icon, and the default-expanded logic in the parent — so all three no longer flag a healthy `Stalled=False` condition.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
weatherhog: Done — `findReadyCondition()` is now called once (gated on `readyStatus === 'False'`) and both `message` and `reason` read from it.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
weatherhog: Fixed — the fallback templates now use the same conditional namespace clause as `buildExplainErrorMessage` (I kept them as inline strings rather than routing through the builder, since they're "please read the resource" prompts, not error explanations).

plugins/ai-chat-react/src/utils/buildExplainErrorMessage.ts
weatherhog: Good catch — the fence is now sized dynamically: computed *after* truncation, one backtick longer than the longest backtick run in the message (minimum 3), so the error text can never terminate the code block early. Added a test with a ```-containing message.

marians (approved):

PR #1939, opened 2026-07-16
refactor(gs): drop custom entity-page header for upstream BUI header

marians:
### What does this PR do?

**R2** of giantswarm/giantswarm#37180 (the Backstage 1.53.0 upgrade work).

Backstage v1.53.0 ships the Backstage UI (BUI) entity-page header (upstream RFC backstage/backstage#31412), so the entity page now renders `EntityLayoutBui` + `EntityHeaderBui` by default. This removes our custom header fork, all in `plugins/gs`:

- Remove the `EntityHeaderBlueprint` wiring (`iconEntityHeader`) from `plugin.tsx`.
- Delete `CustomEntityHeader/` (a full reimplementation of the classic upstream header) and `EntityHeaderIcon/` (fragile `document.querySelector('main > header')` icon injection).

Every entity now uses the upstream default header, so the look is consistent across entities, there's no header fork to maintain, and the upstream context menu (Copy URL / Inspect / Unregister) stays intact.

The catalog entity page header is now the standard upstream BUI header for all entities (previously, entities with a `giantswarm.io/icon-url` annotation rendered a bespoke classic header with a large icon). The GS header icon is dropped — but the icon still appears next to the entity name in **catalog tables and entity ref links** via `CustomEntityPresentationRenderer`, which is unchanged.

### How does it look like?

The entity page uses the upstream BUI header (name, tags, owner/lifecycle metadata, favourite + ⋮ menu, tabs). No screenshot of a new component since this PR only removes GS code and falls back to the well-tested upstream header.

Why the GS icon couldn't be kept in/near the header (for reviewers):
- The bui `Header` has **no image-icon slot** (`title` is a plain string).
- A custom header layout — the only GS-controlled extension mounted across all tabs — does **not** receive the context-menu items; upstream passes those (`Copy URL` / `Inspect` / permission-gated `Unregister`) only to its own header. Re-forking to add a logo would drop/reimplement the ⋮ menu, re-introducing the fork this PR removes.
- DOM-injecting into the upstream React-managed header risks `removeChild` reconciliation crashes and can't persist across tabs.

- Issue: giantswarm/giantswarm#37180 (work package R2)
- Follows PR #1936 (R1, the 1.53.0 version bump)
- Upstream RFC: backstage/backstage#31412

No.

- [x] A changeset describing the change and affected packages was added (`@giantswarm/backstage-plugin-gs`: patch).


marians:
**Code review (medium effort)** — 1 finding.

This is a clean deletion PR; no correctness bugs found. One cleanup item:

- **Orphaned helper: `isEntityWithIcon`** (`plugins/gs/src/components/utils/entity.ts:96`) — its only consumer was the `iconEntityHeader` blueprint in `plugin.tsx`, deleted in this PR. It's not re-exported in any barrel and has no remaining references anywhere in `src`, so it's now dead code. Safe to remove. (`getIconUrlFromEntity` stays — still used by `CustomEntityPresentationRenderer`.)

_Posted as a general comment because the affected line isn't part of this PR's diff, so an inline comment isn't possible._

marians:
Good catch — removed the orphaned `isEntityWithIcon` helper in 5401c24e. Confirmed no remaining references in `src` (its only consumer was the deleted `iconEntityHeader` blueprint), and `getIconUrlFromEntity` stays since `CustomEntityPresentationRenderer` still uses it. tsc / lint / prettier / `plugins/gs` tests (172) all pass.

PR #1943, opened 2026-07-16
feat(ai-chat): add Google Vertex AI (Gemini) as an LLM provider

marians:
### What does this PR do?

Adds **Google Vertex AI (Gemini)** as an LLM provider for the Backstage AI chat backend, plus the first-class Helm chart plumbing to configure it. This is Parts A + B of giantswarm/giantswarm#37197.

A `gemini-*` model in `aiChat.model` now routes chat through Vertex via the first-class [`@ai-sdk/google-vertex`](https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex) provider. Vertex is **not** authenticated with a static API key — the provider uses `google-auth-library` to read a mounted service-account JSON and mint + auto-refresh short-lived (~1h) OAuth2 access tokens, so no custom token-refresh code is needed. Streaming and tool-calling work exactly like the existing providers.

**Backend** (`plugins/ai-chat-backend`)

- New `aiChat.google` config block (`project`, `location`, `keyFilename`).
- `createVertex` wiring with `gemini-` precedence ahead of the Azure/OpenAI selection (so it isn't swallowed by `isAzureConfigured`).
- Config-validation warning when project/location/credentials are missing; `google-vertex` reported on `/health` and in debug metadata.
- Tests for the configured and misconfigured cases.

**Helm chart** (`helm/backstage`)

- First-class `google` block mirroring `githubAppCredentials`: a `stringData` credentials secret (avoids the base64 pitfall), a `/app/google/credentials.json` mount, and `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION` env vars — all rendered only when `google.credentialsJson` is set.
- `app-config.yaml` and `.env.example` updated; `values.schema.json` + chart README regenerated via pre-commit.

Operators can now point the AI chat at a `gemini-*` model backed by Google Vertex AI, authenticating from a mounted service-account JSON. No behaviour change for existing Anthropic / Azure / OpenAI-compatible deployments.

### How does it look like?

`helm template` with a sample `google` block renders the dedicated secret (`stringData`), the volume + `/app/google/credentials.json` mount, and the `GOOGLE_CLOUD_*` env vars — and nothing when `credentialsJson` is unset. `GET /api/ai-chat/health` returns `provider: google-vertex`, `configured: true` once configured.

For the **egger** customer (enigma MC), IT mandates all LLM access go through Google Vertex / Gemini. Part C (customer gitops rollout to enigma) lives in a separate repo and is blocked on the customer-delivered SA JSON, so it is intentionally out of scope here. Refs giantswarm/giantswarm#37197.

`@ai-sdk/google-vertex` is pinned to `^4`, which pairs with the `ai@6` core (language-model spec v2/v3) — the same generation as the other providers after #1944. Verified end-to-end against a real Vertex project: `gemini-2.5-flash` streams a response authenticated from a mounted SA JSON. (Depends on #1944, now merged.)

Yes — done in this PR (`plugins/ai-chat-backend/README.md` provider section; chart README regenerated).


app-config.yaml
```diff
@@ -289,6 +289,14 @@ aiChat:
   #   # baseUrl: https://my-resource.openai.azure.com/openai  # alternative: full URL
   #   # apiVersion: '2024-10-21'  # optional API version override
 
+  # Google Vertex AI configuration (used when model is a gemini-* model).
+  # Authenticates from a mounted service-account JSON; google-auth-library
+  # mints and auto-refreshes short-lived OAuth2 tokens (no static API key).
+  google:
+    project: ${GOOGLE_CLOUD_PROJECT}
+    location: ${GOOGLE_CLOUD_LOCATION}
+    keyFilename: /app/google/credentials.json
```
marians, line 298: **Hardcoded `keyFilename` defeats the documented ADC / `GOOGLE_APPLICATION_CREDENTIALS` fallback.**

This base `app-config.yaml` is loaded in every deployment, so `aiChat.google.keyFilename` is *always* set to `/app/google/credentials.json`. That has two consequences the rest of the PR doesn't account for:

1. `config.d.ts` and the README promise that "if `keyFilename` is unset, `GOOGLE_APPLICATION_CREDENTIALS` (ADC) is honoured." Because this default is always present, that fallback is unreachable — Workload Identity / ADC can never kick in.
2. In `router.ts`, `googleCredentialsAvailable = !!googleKeyFile || ...` is therefore always `true`, so the "configuration incomplete" credential check can never detect a missing/unmounted SA JSON.

Consider gating this on the credentials actually being provided (e.g. only set `keyFilename` in the chart-rendered config when `google.credentialsJson` is set), so the ADC path stays reachable.

plugins/ai-chat-backend/src/router.ts
```diff
@@ -183,8 +184,22 @@ export async function createRouter(
...
+
   // Determine which provider to use based on model name
   const isAnthropicModel = modelName.startsWith('claude-');
+  const isGoogleModel = modelName.startsWith('gemini-');
+
+  // `@ai-sdk/google-vertex` reads credentials either from an explicit
+  // keyFilename or from the `GOOGLE_APPLICATION_CREDENTIALS` env var
+  // (Application Default Credentials). Treat either as "configured".
+  const googleCredentialsAvailable =
+    !!googleKeyFile || !!process.env.GOOGLE_APPLICATION_CREDENTIALS;
```
marians: **`googleCredentialsAvailable` cannot detect missing credentials in a real deployment.**

Since the base `app-config.yaml` hardcodes `aiChat.google.keyFilename`, `googleKeyFile` is always truthy here, so this expression is effectively a constant `true`. The "Google Vertex model selected but configuration is incomplete" warning below then reduces to `!googleProject || !googleLocation` — the credential half of the check is dead. A gemini deployment where the SA JSON is never mounted (file absent at the hardcoded path) passes validation and `GET /health` reports `googleConfigured: true` whenever project/location are set, then fails at the first `/chat` request when google-auth-library tries to read the nonexistent file.

Consider validating that the credentials source actually resolves (file exists, or ADC available) rather than "a path string was configured."

helm/backstage/templates/deployments.yaml
```diff
@@ -130,6 +130,12 @@ spec:
                   name: {{ .Chart.Name }}-{{ .Values.database.postgresql.clusterNameSuffix }}-app
                   key: password
             {{- end }}
+            {{- if .Values.google.credentialsJson }}
+            - name: GOOGLE_CLOUD_PROJECT
+              value: {{ .Values.google.project | quote }}
+            - name: GOOGLE_CLOUD_LOCATION
+              value: {{ .Values.google.location | quote }}
```
marians, line 137: **`GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION` are gated on `credentialsJson`, and `GOOGLE_APPLICATION_CREDENTIALS` is never exported.**

Two coupling problems for anyone not supplying a static SA JSON:

- If an operator wants Vertex via GKE Workload Identity (SA JSON not needed), they'd leave `google.credentialsJson` empty. But then this `{{- if .Values.google.credentialsJson }}` guard means `GOOGLE_CLOUD_PROJECT`/`GOOGLE_CLOUD_LOCATION` are *not* exported, so app-config's `${GOOGLE_CLOUD_PROJECT}` resolves empty and Vertex is reported unconfigured. Project/location can't be set independently of credentials.
- The SA JSON is mounted at `/app/google/credentials.json`, but the deployment never sets `GOOGLE_APPLICATION_CREDENTIALS` to that path. The mount is wired to the provider *only* via the hardcoded `keyFilename` in `app-config.yaml`. If that default is ever overridden or dropped, credentials silently won't be found. Exporting `GOOGLE_APPLICATION_CREDENTIALS=/app/google/credentials.json` here would make the mount self-describing and match `.env.example`.

Suggest gating the project/location env on `google.project`/`google.location` (not `credentialsJson`) and exporting `GOOGLE_APPLICATION_CREDENTIALS` alongside the mount.

plugins/ai-chat-backend/src/router.test.ts
```diff
@@ -53,4 +53,49 @@ describe('createRouter provider configuration logging', () => {
...
+        expect.stringContaining('Google Vertex model selected'),
+      ]),
+    );
+  });
+
+  it('warns when a gemini model is selected but google config is incomplete', async () => {
+    // Operator chose a gemini model but left project/location/credentials
+    // unset -- a genuine misconfiguration that should reach Sentry.
+    const logger = await buildRouter({
+      aiChat: { model: 'gemini-2.5-flash' },
```
marians: **This "incomplete config" test doesn't cover the production config path.**

The config here omits `keyFilename`, so the test only exercises the case where `googleCredentialsAvailable` is false. But the shipped `app-config.yaml` always sets `aiChat.google.keyFilename`, which flips `googleCredentialsAvailable` to `true` and silences the credential half of the warning. So this test passes while the real deployment behaves differently — it gives false confidence that a missing SA JSON is detected. Consider adding a case with `keyFilename` set but no reachable credentials to reflect the production path.

app-config.yaml
marians: Resolved via **Option B** (2ad996e1): dropped the ADC / `GOOGLE_APPLICATION_CREDENTIALS` fallback entirely — the mounted SA JSON is now the single supported path, so the hardcoded `keyFilename` (the fixed mount location) is correct by design. The "always reachable" problem is fixed differently: credential validation now checks the file **exists** (`existsSync`) rather than "a path string is set", so a missing/unmounted secret is detected. README + `config.d.ts` no longer promise the ADC fallback.

plugins/ai-chat-backend/src/router.ts
marians: Fixed in 2ad996e1: `googleCredentialsAvailable` is now `!!googleKeyFile && existsSync(googleKeyFile)`, so the credential half of the check is live again. A gemini deployment where the SA JSON isn't mounted now fails validation, `GET /health` reports `configured: false`, and the "configuration is incomplete" warning fires — instead of passing and 500ing on the first `/chat`.

helm/backstage/templates/deployments.yaml
marians: With **Option B** we dropped ADC / Workload Identity support, which makes this coupling intentional rather than a bug: Vertex now *requires* the mounted SA JSON, so gating `GOOGLE_CLOUD_PROJECT`/`LOCATION` (and the secret/volume/mount) on `credentialsJson` is coherent — no SA JSON means no Vertex. `GOOGLE_APPLICATION_CREDENTIALS` is no longer exported because the provider reads `keyFilename` directly, and the "silently won't be found if the default is dropped" risk is now caught loudly by the `existsSync` validation. So no `deployments.yaml` change. If we ever want Workload Identity, that's a deliberate follow-up (would reintroduce the ADC path + independent project/location env).

plugins/ai-chat-backend/src/router.test.ts
marians: Added in 2ad996e1: new test **"warns when a gemini keyFilename is set but the SA JSON is not mounted"** exercises exactly the production path (`keyFilename` set, file absent → warns). The existing positive test now writes a **real** temp SA file and points `keyFilename` at it, so it reflects a genuinely-mounted secret rather than a path string that happens to be truthy.

PR #1945, opened 2026-07-16
feat(ai-chat): migrate backend AI SDK from ai@6 to ai@7

marians:
### What does this PR do?

Migrates the **AI chat backend** from the `ai@6` generation of the Vercel AI SDK to `ai@7` and the matching `v4` providers. This is the coordinated upgrade tracked in giantswarm/giantswarm#37202 — but scoped **backend-only** (see below).

Changes:

- Bump `ai` → `^7`, `@ai-sdk/anthropic`/`@ai-sdk/openai`/`@ai-sdk/azure` → `^4`, `@ai-sdk/openai-compatible` → `^3`, `@ai-sdk/mcp` → `^2` (in `ai-chat-backend`, `gs-node`, `muster-backend`).
- **Drop** the `@ai-sdk/mcp` SSE transport patch — the fix (treat an `undefined` SSE `event` the same as `event: "message"`) is already upstreamed in `@ai-sdk/mcp@2`.
- `router.ts`: `ai@7` rejects `role: "system"` messages inside `messages`/`prompt` by default; the Anthropic prompt-caching path deliberately puts one there, so it opts back in via `allowSystemInMessages: true`.
- `MusterMcpClient`: pass the now-required `ToolExecutionOptions.context` (`undefined`).
- Decouple the global `resolutions.ai` pin via a **scoped override** so the backend runs `ai@7` while the frontend + assistant-ui stay on `ai@6`.
- Keep the Renovate `ai`/`@ai-sdk/*` major-hold; comment updated to reflect the split state.

**No `TS2883` / provider-utils dedupe was needed:** with current versions, `ai@7.0.28`, all `v4` providers, and `@ai-sdk/mcp@2.0.14` pin the same `@ai-sdk/provider@4.0.3` + `@ai-sdk/provider-utils@5.0.10`, so the backend subtree is already coherent.

### ⚠️ Scope: backend-only (deviates from the tracking issue)

The issue envisioned "one coordinated PR" moving **frontend + backend** to `ai@7` together. That isn't currently possible: **`@assistant-ui/react-ai-sdk` has no `ai@7` release** — its latest (`1.3.40`) still peer-requires `ai@^6` / `@ai-sdk/react@^3`. So the frontend (`plugins/ai-chat`) stays on `ai@6`.

This is safe because the **UI-message-stream wire protocol is unchanged between v6 and v7** (confirmed in the official v6→v7 migration guide), so the `ai@7` backend streams to the `ai@6` frontend unchanged. The two majors coexist via a scoped `resolutions` override; `ai-chat-react` doesn't import `ai` at all.

Follow-ups deferred: lifting the Renovate hold and moving the frontend to `ai@7` should wait until assistant-ui ships `ai@7` support. The provider-spec-compatibility test (giantswarm/giantswarm#37199) is tracked separately and hasn't landed, so it's not touched here.

No user-facing change. AI chat behaves identically; this unblocks `v4`-line provider features/fixes on the backend and stops the `@ai-sdk/*` major drift the Renovate hold was accumulating.

### How does it look like?

Automated verification (all green):

- `yarn install` clean (no age-gate failure), `yarn tsc` passes.
- `yarn test` — 158 tests across `ai-chat-backend` + `gs-node` + `muster-backend` + `ai-chat` pass (includes the Vertex tests merged from #1943).
- `ai-chat-backend` builds (ESM-only `ai@7` bundles).
- Backend **boots under `ai@7`**: ai-chat plugin initializes, DB migrations run, skills load, muster MCP proxies connect, providers instantiate without error.

**Live end-to-end chat verified** in the running app for both providers we ship:

- **Anthropic** (`claude-opus-4-8`): streaming + MCP tool round-trip. ✅
- **Google Vertex / Gemini** (`gemini-2.5-flash`, `google-vertex@5` on `ai@7`): streaming works. ✅ (Confirms the merged #1943 Vertex path on the v4 provider line.)

- Tracking issue: giantswarm/giantswarm#37202
- Regression + fix history: #1926 (broke chat), #1944 (revert + Renovate hold), original pin #1671
- Spec-compat test: giantswarm/giantswarm#37199

No.


renovate-custom.json5
```diff
@@ -29,9 +29,25 @@
...
-      // this rule. Minor/patch updates still flow.
+      // #1926 did, fixed in #1944).
+      //
+      // The ai@6 -> ai@7 migration is now PARTIAL: the backend
+      // (ai-chat-backend + its @ai-sdk/* providers, gs-node, muster-backend)
+      // moved to ai@7 / provider@4 / @ai-sdk/mcp@2, but the FRONTEND
+      // (plugins/ai-chat) is still on ai@6 because @assistant-ui/react-ai-sdk
+      // has no ai@7 release yet (its latest still peer-requires ai@^6 /
+      // @ai-sdk/react@^3). The two `ai` majors coexist via a scoped
+      // `resolutions` override (backend pinned to 7.x; the unscoped `ai` pin
```
marians: This comment (and the changeset) describes a mechanism that no longer exists. The diff **removed** the unscoped `"ai": "6.0.218"` resolution and replaced it with only `"@giantswarm/backstage-plugin-ai-chat-backend/ai": "7.0.28"` — there is no unscoped `ai` pin in `package.json` anymore. So "the unscoped `ai` pin keeps the frontend + assistant-ui on 6.x" is inaccurate: the frontend now stays on 6.x purely via its declared `ai@^6` range + this Renovate major-hold, not via a resolution pin.

Functionally this is fine today (yarn.lock resolves the frontend to `6.0.228`), but two things worth noting:
- The previous exact-version lock (`6.0.218`) is gone; the frontend now floats within `^6`.
- A future maintainer reading this comment will look for an unscoped pin that isn't there.

Consider either re-adding an explicit unscoped `"ai": "^6"` (or exact) resolution to make the split-major intent enforced-in-code, or updating this comment + the changeset to say the frontend stays on 6.x via its own `ai@^6` dependency range guarded by this hold.

plugins/ai-chat-backend/src/router.ts
```diff
@@ -518,6 +518,14 @@ export async function createRouter(
           : dereasonedMessages,
         system: isAnthropicModel ? undefined : effectiveSystemPrompt,
         abortSignal: req.socket ? undefined : undefined,
+        // ai@7 rejects `role: 'system'` messages inside `messages`/`prompt` by
```
marians, line 577: Not introduced by this PR, but this line sits in the hunk you're touching: `abortSignal: req.socket ? undefined : undefined` evaluates to `undefined` on both branches, so no abort signal is ever passed to `streamText`. Combined with the `res.once('close')` handler below (which only *logs* the early disconnect), a client that hangs up mid-stream never actually aborts the model call — the backend keeps generating (and paying for) tokens until the model finishes on its own.

If the intent was to cancel generation on client disconnect, wire a real `AbortController`: abort it in the `close` handler and pass `abortSignal: controller.signal` here. Otherwise this ternary is dead and should just be dropped.

marians (commented):
Reviewed the `ai@6 → ai@7` backend migration (medium-effort correctness pass). The actual code surface is small; most of the diff is lockfile/manifest churn, and the two real API adaptations (`allowSystemInMessages`, `ToolExecutionOptions.context`) look correct. I verified the dropped `@ai-sdk/mcp` SSE patch is genuinely upstreamed in `@ai-sdk/mcp@2.0.14` (`isMessageEvent()` treats `event === undefined` the same as `"message"`), so removing it is safe. Two inline notes below — neither blocks the migration itself.

renovate-custom.json5
marians: Good catch — fixed in 2e0dafed. You're right there is no unscoped `ai` resolution; I removed it deliberately because an unscoped pin overrides scoped ones in Yarn 4 (verified empirically: with both present, the backend resolved back to 6.0.218), so re-adding one would defeat the split. Went with your alternative: updated both the renovate comment and the changeset to say the frontend stays on 6.x via its own `ai@^6` range, guarded by this major-hold. The exact `6.0.218` lock is indeed gone — the frontend now floats within `^6` (currently 6.0.228).

plugins/ai-chat-backend/src/router.ts
marians: Fixed in 2e0dafed — dropped the dead ternary. Wiring a real `AbortController` (abort in the `close` handler, pass `signal` here) so a client disconnect actually cancels the model call is a worthwhile fix for the token-cost bug you describe, but it's a behavior change unrelated to the SDK bump, so I kept it out of this migration PR. Happy to do it as a small follow-up if you want.

PR #1948, opened 2026-07-17
test(ai-chat): guard provider/core spec-version compatibility

marians:
### What does this PR do?

Adds defense-in-depth test coverage for the AI-SDK provider ↔ core language-model **spec-version** compatibility in `plugins/ai-chat-backend`.

- **Enabling refactor:** extracts provider selection out of the `/chat` handler into a pure `selectModel()` helper (`selectModel.ts`) that builds the AI-SDK model and reports which provider backs it. Selection precedence is unchanged (`claude-` → Anthropic, `gemini-` → Vertex ahead of Azure, else Azure / openai-compatible / OpenAI).
- **New test (`selectModel.test.ts`):** drives each selectable provider model (Anthropic, OpenAI, Azure, openai-compatible, Vertex) through `streamText` and asserts it resolves **without** `AI_UnsupportedModelVersionError`, plus covers the selection precedence. Fully hermetic — injected `fetch`, stubbed Vertex auth client; no network, no credentials — so it runs in the normal `yarn test` suite.

The test does not hardcode a spec-version string — it only asserts the core *accepts* the provider's model — so it's correct across generations and needed no change when rebased onto the `ai@7` migration below.

`selectModel()` is called per request inside the handler's `try/catch` (not at router-creation time) so an incomplete provider config still produces the existing graceful warning + per-request error rather than a hard `createRouter` failure.

No end-user-facing change. Internal refactor + test coverage only.

### How does it look like?

```
Tests:       100 passed, 100 total   (11 new)
```

Rebased on top of **#1945** (backend `ai@6 → ai@7` migration), so this now guards the current stack: `ai@7` (spec `v4`) + providers at their `v4`-spec majors (`@ai-sdk/anthropic@4`, `azure@4`, `openai@4`, `openai-compatible@3`, `google-vertex@5`). All five resolve cleanly through `streamText`. This is exactly the migration the issue flagged as the moment such a guard earns its keep.

The test intentionally couples to the installed AI-SDK generation and fails loudly on a mismatch. Verified locally by driving a model whose advertised spec version the core rejects → `streamText` throws `AI_UnsupportedModelVersionError` during resolution, before any HTTP request — exactly the #1926 shape.

In #1926 a Renovate bump moved the `@ai-sdk/*` providers a spec version ahead of the `ai` core, so **every** chat request failed at runtime with `AI_UnsupportedModelVersionError`. It was reverted in #1944 (+ a Renovate major-hold). CI missed it because `router.test.ts` only inspects the logger — it never builds a provider model or calls `streamText`, which is where the error is thrown (during model resolution, before any HTTP request). This test exercises that exact path.

Implements giantswarm/giantswarm#37199. Related: #1926, #1943, #1944, #1945. Complements, does not replace, the Renovate major-hold from #1944.

No.

No changeset: `@giantswarm/backstage-plugin-ai-chat-backend` is a private package (not published to npm), and this is a test/internal-refactor change with no runtime behavior change.

### DIFF
--- a/plugins/ai-chat-backend/src/router.ts
+++ b/plugins/ai-chat-backend/src/router.ts
@@ -6,11 +6,6 @@ import {
 } from '@backstage/backend-plugin-api';
 import { Config } from '@backstage/config';
 import { InputError } from '@backstage/errors';
-import { createOpenAI } from '@ai-sdk/openai';
-import { createOpenAICompatible } from '@ai-sdk/openai-compatible';
-import { createAnthropic } from '@ai-sdk/anthropic';
-import { createAzure } from '@ai-sdk/azure';
-import { createVertex } from '@ai-sdk/google-vertex';
 import {
   convertToModelMessages,
   stepCountIs,
@@ -19,6 +14,11 @@ import {
   UIMessage,
   SystemModelMessage,
 } from 'ai';
+import {
+  selectModel,
+  isAzureConfigured,
+  type SelectModelOptions,
+} from './selectModel';
 import { z } from 'zod/v3';
 import express from 'express';
 import Router from 'express-promise-router';
@@ -179,10 +179,14 @@ export async function createRouter(
   );
   const azureBaseUrl = config.getOptionalString('aiChat.azure.baseUrl');
   const azureApiVersion = config.getOptionalString('aiChat.azure.apiVersion');
-  const isAzureConfigured = !!(
-    azureApiKey &&
-    (azureResourceName || azureBaseUrl)
-  );
+  // Reuse the same "Azure configured" predicate that drives provider selection
+  // so the config-validation warnings below can never disagree with which
+  // provider `selectModel` actually picks.
+  const azureConfigured = isAzureConfigured({
+    apiKey: azureApiKey,
+    resourceName: azureResourceName,
+    baseUrl: azureBaseUrl,
+  });
 
   // Get Google Vertex AI configuration. Vertex is not authenticated with a
   // static API key: `@ai-sdk/google-vertex` uses google-auth-library to read
@@ -219,12 +223,7 @@ export async function createRouter(
     );
   }
 
-  if (
-    !isAnthropicModel &&
-    !isGoogleModel &&
-    isAzureConfigured &&
-    !azureApiKey
-  ) {
+  if (!isAnthropicModel && !isGoogleModel && azureConfigured && !azureApiKey) {
     logger.warn(
       'Azure OpenAI configured but no API key set. Set aiChat.azure.apiKey in app-config.yaml',
     );
@@ -233,7 +232,7 @@ export async function createRouter(
   if (
     !isAnthropicModel &&
     !isGoogleModel &&
-    !isAzureConfigured &&
+    !azureConfigured &&
     !openaiApiKey
   ) {
     // Expected when a customer hasn't set up the AI chat feature: the default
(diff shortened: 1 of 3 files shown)


plugins/ai-chat-backend/src/selectModel.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    const result = streamText({
+      model,
+      messages: [{ role: 'user', content: 'ping' }],
+      // Stream errors from the fake fetch (e.g. the empty body producing no
+      // finish chunk) are routed here rather than thrown or logged, so only the
+      // synchronous resolution error can escape the try.
+      onError: () => {},
+    });
+    // Consume so resolution is driven end-to-end.
+    await result.consumeStream({ onError: () => {} });
```
marians: **The spec-version guard may never fail.** Both `streamText({ onError: () => {} })` (line 63) and `result.consumeStream({ onError: () => {} })` (this line) swallow every error delivered through the stream. `resolutionError` only returns an error for a *synchronous* throw from `streamText()`.

If the AI SDK surfaces `UnsupportedModelVersionError` asynchronously via the stream error channel (which is how streaming errors normally propagate in `ai`), `resolutionError` returns `undefined` and `expectResolvesWithoutVersionError` passes — so a real core/provider spec-version mismatch (the exact #1926 regression this PR exists to catch) would go undetected and the test would give false confidence.

Failure scenario: a future `@ai-sdk/*` major bump reintroduces a v2/v3 spec mismatch; the SDK reports it through the stream rather than throwing synchronously → this test stays green. Suggest capturing the stream error (record it in the `onError` callbacks and assert on it) rather than discarding it, and confirming the guard actually fails when pointed at a mismatched provider.

plugins/ai-chat-backend/src/selectModel.ts
```diff
@@ -0,0 +1,163 @@
...
+export interface SelectedModel {
+  model: LanguageModel;
+  providerName: ProviderName;
+}
+
+/**
+ * Determine whether the Azure OpenAI branch is usable. Azure needs an API key
+ * plus either a resource name or an explicit base URL.
+ */
+export function isAzureConfigured(azure: SelectModelOptions['azure']): boolean {
```
marians, line 72: **Reuse:** `router.ts` still keeps its own inline definition of "Azure configured" — `const isAzureConfigured = !!(azureApiKey && (azureResourceName || azureBaseUrl))` (router.ts:178) — used for the config-validation warnings, rather than importing this newly-exported `isAzureConfigured` helper. Two copies of the same predicate can drift: if the notion of "Azure configured" ever changes (e.g. base-URL-only auth), one site updates and the other silently disagrees, so warnings and actual provider selection diverge. Consider importing and reusing this helper in `router.ts`.

plugins/ai-chat-backend/src/router.ts
```diff
@@ -450,30 +422,10 @@ export async function createRouter(
...
-        selectedModel = anthropic(modelName);
-      } else if (isGoogleModel) {
-        selectedModel = vertex(modelName);
-      } else {
-        selectedModel = openaiCompatibleModel;
-      }
+      // Build the language model and resolve which provider backs it (see the
+      // note at `selectModelOptions` for why this is per request).
+      const { model: selectedModel, providerName } =
+        selectModel(selectModelOptions);
```
marians, line 431: **Efficiency (minor, documented):** `selectModel` now runs on every `/chat` request, constructing a fresh provider client (`createOpenAI`/`createAnthropic`/`createVertex`/…) per call, whereas the old code built the clients once at router creation. The per-request placement is deliberately justified in the comment above (avoid eager-validation failures at `createRouter` time), so this is just noting the tradeoff — factory construction is cheap closures, but if it ever grows heavier consider memoizing the selected client by provider.

plugins/ai-chat-backend/src/selectModel.test.ts
marians: Good catch — fixed in ea536b63.

`resolutionError` → `collectVersionError` now records errors from **every** channel: the synchronous throw *and* both `onError` callbacks (`streamText` and `consumeStream`). It then looks for the version error among them, so the guard no longer depends on the check being thrown synchronously.

Two things I found while fixing it:

1. **`UnsupportedModelVersionError.isInstance()` matches *any* `AISDKError` on ai@7** (it checks the shared base-class marker). Naively filtering with it flagged the benign offline error (empty-body `NoOutputGeneratedError`) as a version mismatch. Switched to matching the stable, documented `AI_UnsupportedModelVersionError` name instead.
2. **Added a committed non-vacuous test** ("detects a provider that outruns the core") that feeds `streamText` a model one spec generation *ahead* of what the installed core accepts and asserts the guard actually fires. The "ahead" version is derived from a live provider's current spec (`v{current+1}`), so it reproduces the #1926 shape and survives the next `ai` major bump without hardcoding a spec string.

Empirically on ai@7: the core is backward-compatible (accepts v2/v3/v4) and rejects only a spec *higher* than it supports (v5), synchronously — i.e. exactly the "provider ahead of core" regression, which both the sync path and the new stream-channel capture now catch.

plugins/ai-chat-backend/src/selectModel.ts
marians: Done in ea536b63. `router.ts` now imports the exported `isAzureConfigured` helper and derives its local `azureConfigured` predicate from it (used by the config-validation warnings and the `/health` endpoint), so the two definitions can no longer drift.

plugins/ai-chat-backend/src/router.ts
marians: Acknowledged — leaving as-is per your note. The per-request construction is the intentional tradeoff to avoid eager-validation failures at `createRouter` time (documented in the comment above the call). Factory construction is cheap closures; if it ever grows heavier we can memoize the selected client by provider.

PR #1951, opened 2026-07-17
fix: keep self-managed Kustomizations as Flux tree roots and guard empty tree

weatherhog:
### What does this PR do?

Fixes the Flux resources tree view crashing with `Error: TreeWalker must yield at least one root node on the first iteration`.

Two fixes (reframed after review — thanks @marians):

1. **Namespace-aware root detection.** `findRoots` matched inventory references by **name only**, so any inventory entry disqualified every Kustomization sharing that name in *any* namespace. On multi-org clusters with cross-namespace name reuse this can disqualify all Kustomizations, collapsing the tree to zero roots. References are now matched by namespace **and** name.
2. **Empty-tree crash guard.** `FluxOverview` guarded rendering with `{tree ? ...}` — an empty array is truthy — so a zero-root tree reached react-vtree, which throws when the tree walker yields nothing. An empty tree (also reachable via the "Failing only" filter on a cluster without failures) now renders an `EmptyState` instead.

Not in this PR (dropped after review): self-reference guards for self-managed Kustomizations — `parseInventoryEntries` already strips a Kustomization's own inventory self-entry at parse time, so those guards were dead code. A test remains that pins the combined invariant (self-managed root stays a root, no self-child), so a future parser change can't silently collapse the tree.

The Flux tree view no longer crashes on clusters where root detection previously found no roots, and the "Failing only" filter shows a friendly empty state on clusters without failures instead of an error.

Observed on the internal devportal (error reported on the tree view; the error boundary showed react-vtree's TreeWalker error). The empty-tree guard also hardens the "Failing only" filter added in #1933. Related to the https://github.com/giantswarm/giantswarm/issues/35859 series.

No.

### DIFF
--- a/plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
+++ b/plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
@@ -63,40 +63,40 @@ export const FluxOverview = ({
         />
       ) : null}
 
-      {/* An empty tree must not reach OverviewTree — react-vtree throws when
-          the tree walker yields no root node. */}
-      {tree && tree.length === 0 ? (
-        <EmptyState
-          missing="content"
-          title={
-            statusFilter === 'failing'
-              ? 'No failing resources'
-              : 'No resources to display'
-          }
-          description={
-            statusFilter === 'failing'
-              ? 'All Flux resources on this cluster are healthy.'
-              : 'No Flux resources were found on this cluster.'
-          }
-        />
-      ) : null}
-
-      {tree && tree.length > 0 ? (
-        <ContentContainer
-          renderContent={containerHeight => (
-            <OverviewTree
-              ref={treeRef}
-              tree={tree}
-              compactView={compactView}
-              selectedResourceRef={selectedResourceRef}
-              height={containerHeight}
-              onSelectResource={onSelectResource}
-              searchMatchIds={searchMatchIds}
-              currentMatchId={currentMatchId}
-              pathsToExpand={pathsToExpand}
-            />
-          )}
-        />
+      {tree ? (
+        // An empty tree must not reach OverviewTree — react-vtree throws
+        // when the tree walker yields no root node.
+        tree.length === 0 ? (
+          <EmptyState
+            missing="content"
+            title={
+              statusFilter === 'failing'
+                ? 'No failing resources'
+                : 'No resources to display'
+            }
+            description={
+              statusFilter === 'failing'
+                ? 'No Flux resources are currently failing.'
+                : 'No Flux resources were found on this cluster.'
+            }
+          />
+        ) : (
+          <ContentContainer
+            renderContent={containerHeight => (
+              <OverviewTree
+                ref={treeRef}
+                tree={tree}
+                compactView={compactView}
+                selectedResourceRef={selectedResourceRef}
+                height={containerHeight}
+                onSelectResource={onSelectResource}
+                searchMatchIds={searchMatchIds}
+                currentMatchId={currentMatchId}
+                pathsToExpand={pathsToExpand}
+              />
+            )}
+          />
+        )
       ) : null}
     </Box>
   );
--- a/plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
+++ b/plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
@@ -28,6 +28,7 @@ export const FluxOverview = ({
     tree,
     isLoading,
     resourceType,
+    statusFilter,
     searchMatches,
     currentMatchId,
     pathsToExpand,
@@ -62,7 +63,25 @@ export const FluxOverview = ({
         />
       ) : null}
 
-      {tree ? (
+      {/* An empty tree must not reach OverviewTree — react-vtree throws when
+          the tree walker yields no root node. */}
+      {tree && tree.length === 0 ? (
+        <EmptyState
+          missing="content"
+          title={
+            statusFilter === 'failing'
+              ? 'No failing resources'
+              : 'No resources to display'
+          }
+          description={
+            statusFilter === 'failing'
+              ? 'All Flux resources on this cluster are healthy.'
+              : 'No Flux resources were found on this cluster.'
+          }
+        />
+      ) : null}
+
+      {tree && tree.length > 0 ? (
         <ContentContainer
           renderContent={containerHeight => (
             <OverviewTree
--- a/plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
+++ b/plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
@@ -51,6 +51,45 @@ export const FluxOverview = ({
     return undefined;
   }, [currentMatchId]);
 
+  // Deriving the content from `tree` in one place keeps the empty state and
+  // the tree mutually exclusive. An empty tree must not reach OverviewTree —
+  // react-vtree throws when the tree walker yields no root node.
+  let treeContent = null;
+  if (tree) {
+    treeContent =
+      tree.length === 0 ? (
+        <EmptyState
+          missing="content"
+          title={
+            statusFilter === 'failing'
+              ? 'No failing resources'
+              : 'No resources to display'
+          }
+          description={
+            statusFilter === 'failing'
+              ? 'No Flux resources are currently failing.'
+              : 'No Flux resources were found on this cluster.'
+          }
+        />
+      ) : (
+        <ContentContainer
+          renderContent={containerHeight => (
+            <OverviewTree
+              ref={treeRef}
+              tree={tree}
+              compactView={compactView}
+              selectedResourceRef={selectedResourceRef}
+              height={containerHeight}
+              onSelectResource={onSelectResource}
+              searchMatchIds={searchMatchIds}
+              currentMatchId={currentMatchId}
+              pathsToExpand={pathsToExpand}
+            />
+          )}
+        />
+      );
+  }
+
   return (
     <Box display="flex" flexDirection="column" height="100%">
       {isLoading ? <Progress /> : null}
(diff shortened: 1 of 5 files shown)


plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
```diff
@@ -156,23 +156,29 @@ export class KustomizationTreeBuilder {
...
 
-    return Array.from(this.kustomizations.values()).filter(
-      k =>
-        !Boolean(
-          kustomizationInventoryEntries.find(e => e.name === k.getName()),
-        ),
+          return inventoryEntries
+            .filter(e => e.kind === Kustomization.kind)
+            .map(e => this.getKey(e.name, e.namespace))
+            .filter(entryKey => entryKey !== ownerKey);
```
marians: **The self-exclusion here is redundant — `parseInventoryEntries` already strips a Kustomization's own self-reference.**

`inventoryParser.ts:108-118` filters out any inventory entry whose group+kind+namespace+name equal the owning Kustomization's, so a self-managed Kustomization's parsed inventory (what lands in `this.inventories`) never contains itself. The `.filter(entryKey => entryKey !== ownerKey)` guard therefore never removes a real entry.

I verified this empirically: checking out **only** this PR's test file onto unpatched `main` source, the `keeps a self-referencing kustomization as root` test **passes**, while `matches inventory references by namespace and name` **fails**. The genuinely effective fix in this PR is the namespace-aware key matching (name-only matching was the real cause of gazelle's zero-roots), not the self-management handling. Worth either dropping the self-exclusion code + reframing the PR/changeset around the namespace fix, or — if you believe self-refs can survive parsing in prod — fixing that at the source in `parseInventoryEntries` instead.

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
```diff
@@ -268,8 +274,18 @@ export class KustomizationTreeBuilder {
...
+    // does not appear as its own child.
+    const childResources = this.sortChildResources(
+      this.findChildren(key).filter(
+        child =>
+          !(
+            child.kind === Kustomization.kind &&
+            this.getKey(child.name, child.namespace) === key
+          ),
+      ),
+    );
```
marians, line 278: **This self-child filter is also dead code for the same reason.** Since `parseInventoryEntries` (`inventoryParser.ts:108-118`) already removes the self-entry from a Kustomization's parsed inventory, `findChildren(key)` never returns a child whose key equals `key`, so this `.filter(...)` never excludes anything. It can be removed along with the `findRoots` self-exclusion.

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.test.ts
```diff
@@ -172,6 +172,94 @@ describe('KustomizationTreeBuilder', () => {
     expect(findNode(tree, 'my-app')).toBeDefined();
   });
 
+  describe('self-managed root kustomizations', () => {
+    it('keeps a self-referencing kustomization as root and hides the self-child', () => {
```
marians, line 176: **This test passes against unpatched `main` — it does not guard the regression it claims to.** Because `parseInventoryEntries` strips the self-reference before the builder sees it, the old code already keeps a self-referencing Kustomization as a root and never shows it as its own child. I confirmed this by running this test file against the pre-PR source: this case passes, only `matches inventory references by namespace and name` fails. If the intent is to protect the self-managed-root behavior, the test needs to reproduce the actual failing condition (or the fix belongs in `parseInventoryEntries`).

marians (approved):
Nothing blocking, but improvements suggested

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
weatherhog: **The self-managed-root guards are unreachable with real data, and the new test does not cover them.**

`this.inventories` is populated via `parseInventoryEntries` (constructor, line 91), which already strips any entry that exactly matches the owning Kustomization on group/kind/namespace/name (`inventoryParser.ts:110-118`). So by the time `findRoots` runs, a self-managed Kustomization has no self-entry left — this `entryKey !== ownerKey` filter, and the matching self-child filter in `buildSubtree` (lines 281-287), can never fire.

Empirical check: reverting only this file to `origin/main` and re-running the new tests, `keeps a self-referencing kustomization as root and hides the self-child` still **passes** (the mock self-entry is stripped at parse time); only `matches inventory references by namespace and name` fails. In other words, the real fix for the zero-roots crash in this PR is the namespace-aware matching — the self-managed guards are dead code and the headline test provides no regression coverage for them.

Suggest either:
- drop the two self-reference guards (and the first test) and attribute the fix to namespace-aware matching, or
- if you want them as belt-and-braces (e.g. in case the parse-time strip ever changes), add a test where the self-entry actually survives parsing, and consider enforcing the rule in one place (e.g. a shared `isSelfReference` helper or a single filter where inventories are built) instead of two divergent spellings.

plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
```diff
@@ -62,7 +63,25 @@ export const FluxOverview = ({
...
+        <EmptyState
+          missing="content"
+          title={
+            statusFilter === 'failing'
+              ? 'No failing resources'
+              : 'No resources to display'
+          }
+          description={
+            statusFilter === 'failing'
+              ? 'All Flux resources on this cluster are healthy.'
```
weatherhog: Minor copy issue: "All Flux resources on this cluster are healthy" is a false claim when the failing-filtered tree is empty but resources are suspended or in Ready=Unknown/reconciling state. `isResourceFailing` (KustomizationTreeBuilder.ts:41-53) returns `false` for suspended resources and for anything whose Ready condition is not `False`, so those all get pruned by the failing filter without being healthy.

```suggestion
              ? 'No Flux resources are currently failing.'
```

plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
```diff
@@ -62,7 +63,25 @@ export const FluxOverview = ({
         />
       ) : null}
 
-      {tree ? (
+      {/* An empty tree must not reach OverviewTree — react-vtree throws when
+          the tree walker yields no root node. */}
+      {tree && tree.length === 0 ? (
```
weatherhog: Nit: these two sibling branches (`tree && tree.length === 0` here and `tree && tree.length > 0` below) are kept mutually exclusive only by hand-matched predicates. A single `tree ? (tree.length === 0 ? <EmptyState .../> : <ContentContainer .../>) : null` makes the exclusivity structural, so a future edit to one predicate can't make both (or neither) render.

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
weatherhog: You're right — I verified: `parseInventoryEntries` strips the exact self-entry (group+kind+namespace+name), so the self-exclusion never removed a real entry. Dropped it (and the matching self-child filter), and reframed the PR/changeset around the namespace-aware matching as the actual fix. My original root-cause analysis was wrong on this point — I had reproduced the zero-roots condition against raw inventories without the parse-time strip.

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
weatherhog: Dropped along with the `findRoots` self-exclusion — same reasoning, the parse-time strip makes it unreachable.

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.test.ts
weatherhog: Kept the test but reframed it honestly: it now documents that the self-managed-root invariant is enforced at parse time (`parseInventoryEntries`), so it guards against a future parser change silently collapsing the tree — cross-layer coverage rather than coverage of builder code. The comment in the test says so explicitly. Happy to drop it entirely if you'd rather not keep a test that passes on main.

plugins/flux-react/src/components/FluxOverview/utils/KustomizationTreeBuilder/KustomizationTreeBuilder.ts
weatherhog: Addressed as part of the rework: both guards removed, PR/changeset reframed around the namespace-aware matching, and the remaining test re-commented as cross-layer invariant coverage (single enforcement point stays `parseInventoryEntries`).

plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
weatherhog: Applied — "No Flux resources are currently failing." is the accurate claim given suspended/reconciling resources are pruned without being healthy.

plugins/flux-react/src/components/FluxOverview/FluxOverview.tsx
weatherhog: Applied — the empty state and the tree are now branches of a single ternary under one `tree ?` guard, so the exclusivity is structural.

PR #1952, opened 2026-07-17
feat: list kagent agents on the Agent Platform Agents tab

marians:
### What does this PR do?

Replaces the stub landing on the Agent Platform "Agents" tab (`/agent-platform/agents`) with a real list of kagent `Agent` resources running across the fleet.

- Lists `Agent` resources across all reachable installations and namespaces in a bui table with columns: display name + description, installation, namespace, model, and skill count. The model column resolves each agent's `spec.declarative.modelConfig` reference to the referenced ModelConfig's friendly name (reusing `ModelConfigsProvider`), falling back to the raw ref and to `—` for agents with no model (e.g. BYO).
- A sticky per-installation data provider accumulates rows so agents don't flicker in and out as the reachable-installation set churns or a background refetch transiently fails; the query client now caches/persists results like the other fleet lists.
- Shows a progress bar until the first agents arrive (rather than an empty skeleton), and reports installations whose Agents couldn't be read in a warning card below the table — via a new shared `UnreachableInstallationsAlert` that the create flow's installation select now reuses too.
- `kubernetes-react`: fixes `Agent.getSkillRefs()` to read `spec.skills.gitRefs` (the real `v1alpha2` field — it previously read a non-existent `spec.skills.refs` and always returned `[]`), and adds `getDisplayName()`, `getSkillCount()`, and `getType()`.

Users can now see the agents that exist across their management clusters directly on the Agents tab, instead of only a "coming soon" placeholder.

### How does it look like?

The Agents tab now renders a table of agents (name/description, installation, namespace, model, skill count) with the "New agent" action in the section header. Screenshot to follow.

Direction comes from the [`giantswarm/agentic-platform-ui`](https://github.com/giantswarm/agentic-platform-ui) prototype (`/agents`), scoped down to just the table for now (no search/filter/sort/stats/row-actions/detail page yet).

A follow-up, [giantswarm/giantswarm#37209](https://github.com/giantswarm/giantswarm/issues/37209), tracks tightening installation reachability to query only confirmed-healthy installations — deliberately kept out of this PR because it also changes the already-merged create flow.

No.


plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
```diff
@@ -0,0 +1,188 @@
...
+      cluster => !agentsByInstallation[cluster]?.length,
+    );
+
+    // The fleet-wide query reports "loading" until every installation settles,
+    // so gate the blocking state on having no rows yet — otherwise one slow or
+    // failing installation would hide agents already loaded from healthy ones.
+    const isBusy = isProbing || isLoading;
+    return {
+      rows,
+      isLoading: isBusy && rows.length === 0,
```
marians, line 198: **Zero configured installations → perpetual spinner, "No installations configured" empty state is dead code.**

When `gs.installations` is empty, `allInstallations` is `[]`, and `useReachableInstallations([])` returns `{ installations: [], isProbing: true }` permanently (its `entries.length === 0` fallback), because the cluster-access status set never gets any entry. So here `isBusy = isProbing || isLoading` is always `true`, `rows.length === 0`, and this line makes the context report `isLoading: true` forever.

In `AgentsIndexPage`, the empty-state guard is `if (!isLoading && !hasInstallations)` — with `isLoading` stuck `true` it never runs, and the render falls through to `isLoading ? <Progress …/>`, so the user sees a perpetual "Loading agents" bar instead of the "No installations configured" state.

The empty-state test passes only because it mocks `useAgents` to return `{ hasInstallations: false, isLoading: false }` — a combination the real provider cannot produce. Consider gating the blocking state on `hasInstallations` too (e.g. `isLoading: hasInstallations && isBusy && rows.length === 0`).

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
```diff
@@ -0,0 +1,188 @@
...
+  );
+
+  // Stable signature of this render's per-cluster outcome, so the reconciling
+  // effect runs only when the actual results change (clustersData/errors are
+  // fresh arrays every render).
+  const readSignature = useMemo(() => {
+    const ok = clustersData.map(
+      ({ cluster, data }) =>
+        `ok:${cluster}:${data
+          .map(item => item.metadata?.name ?? '')
```
marians: **In-place agent edits are not reflected — the reconcile effect keys only on resource names.**

`readSignature` summarises each cluster's result as just the list of resource `metadata.name`s, and the reconciling `useEffect` below depends only on `[readSignature]` (with `resources`/`clustersData` intentionally omitted). So any change to an existing Agent whose name is unchanged — a new `spec.declarative.modelConfig`, edited `description`, changed `ui.giantswarm.io/display-name` annotation, or added/removed skills — produces an identical signature, the effect does not re-run, and `agentsByInstallation` keeps the stale `Agent` instance.

Result: after a background refetch picks up the edit, the table still shows the old model / description / display name / skill count until the *set of names* changes or the page remounts. (Model labels do update progressively via `modelConfigsFor`, but only because that path re-resolves in the `value` memo — agent-intrinsic fields have no such escape hatch.)

Include enough of each item to detect mutation, e.g. `resourceVersion` (or the fields the row actually renders) in the signature.

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
```diff
@@ -0,0 +1,188 @@
...
+      ),
+    );
+
+    // Only surface installations we have *nothing* for; if we're still showing
+    // an installation's last-known agents, don't also claim we couldn't read it.
+    const unreachableInstallations = erroredInstallations.filter(
+      cluster => !agentsByInstallation[cluster]?.length,
+    );
+
+    // The fleet-wide query reports "loading" until every installation settles,
```
marians, line 188: **Sticky per-installation caches are never pruned — installations that leave the reachable set show stale agents indefinitely with no indication.**

`agentsByInstallation` only ever *replaces* an entry when a cluster responds again; nothing removes an entry. Once an installation drops out of `reachableInstallations` (e.g. it transitions to `session-expired` or `degraded`, or is removed from `gs.installations`) it is no longer queried, so it produces neither a success nor an error — its last-known agents remain in the table for the life of the mount.

Because `unreachableInstallations` filters out any cluster that still has cached rows (`!agentsByInstallation[cluster]?.length`), such an installation is *also* not surfaced in the warning card. Net effect: a user whose session for installation X has expired keeps seeing X's agents as if live, with no "stale"/"couldn't refresh" signal. Stickiness across a *transient* refetch failure is the stated goal, but a *durable* drop-out is different — consider pruning (or flagging as stale) installations that have left the reachable set entirely.

plugins/agent-platform/src/components/AgentsIndexPage/AgentsIndexPage.tsx
```diff
@@ -1,46 +1,100 @@
...
-            </Button>
-          </Box>
-        </Flex>
-      </Box>
+      <Flex direction="column" gap="3">
+        <Text color="secondary">
+          Agents running across your management clusters.
+        </Text>
+
+        {isLoading ? (
```
marians, line 65: **During initial load, the unreachable-installations warning is suppressed along with the table.**

The whole results block — table *and* `UnreachableInstallationsAlert` — lives in the `else` branch of `isLoading ?`. Since `isLoading` (context) is `isBusy && rows.length === 0`, if some installations fail while others are still slow and no rows have arrived yet, the "Couldn't read N installations" warning is hidden behind the `<Progress>` bar until the *first row* appears. In a fleet where the only reachable installations are ones the user can't read (all error, zero rows), the warning stays suppressed for the full retry window before finally showing. Consider rendering the alert regardless of the loading branch.

plugins/agent-platform/src/components/QueryClientProvider.tsx
```diff
@@ -1,35 +1,64 @@
...
+          retryDelay: attempt => Math.min(1000 * 2 ** attempt, 30000),
+          gcTime,
+        },
+      },
+    };
+    return new QueryClient(config);
+  }, []);
+
+  const persister = useMemo(
+    () => createAsyncStoragePersister({ storage: window.localStorage }),
```
marians, line 52: **Persisting the full fleet cache to `localStorage` can exceed the storage quota.**

`createAsyncStoragePersister({ storage: window.localStorage })` serialises the entire react-query cache — every installation's Agents *and* ModelConfigs across all namespaces — into a single `localStorage` key. For a large fleet this can exceed the ~5 MB quota; the write then throws `QuotaExceededError` and persistence silently stops working (so the "show last-known agents instantly on reload" benefit is lost exactly when the fleet is biggest). This mirrors the existing gs `QueryClientProvider`, so it's a pre-existing pattern rather than new to this PR, but worth noting since agent lists across all namespaces can be sizeable. A `maxSize`/`throttle` or a `dehydrateOptions` filter to persist only the lists that matter would bound it.

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
marians: Fixed in 33efabcd. The blocking state is now gated on `hasInstallations` (`isBusy = hasInstallations && (isProbing || isLoading)`), so an empty `gs.installations` reports `isLoading: false` and the "No installations configured" empty state renders. Added a provider test for that real combination.

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
marians: Fixed in 33efabcd. `readSignature` now includes each item's `resourceVersion` (`name@resourceVersion`), so an in-place edit changes the signature, the reconcile effect re-runs, and the cached `Agent` instance is replaced. Added a test that bumps `resourceVersion` with an edited description and asserts the row updates.

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
marians: Fixed in 33efabcd. Added an effect keyed on the reachable-installation set that prunes `agentsByInstallation` and `erroredInstallations` for installations that have durably left the reachable set (session-expired/degraded/removed). A transient refetch miss keeps the installation reachable, so its rows are still retained — separate tests cover both the transient-miss (retained) and durable-drop (pruned) cases.

plugins/agent-platform/src/components/AgentsIndexPage/AgentsIndexPage.tsx
marians: Fixed in 33efabcd. The `UnreachableInstallationsAlert` is now rendered outside the `isLoading` branch, so a fleet whose only reachable installations all error surfaces the warning during initial load instead of only an indefinite progress bar. Added a test for loading + non-empty unreachable set.

plugins/agent-platform/src/components/QueryClientProvider.tsx
marians: Agreed it's a real, pre-existing concern (mirrors the gs `QueryClientProvider`). Since it's a shared cross-plugin issue rather than specific to this PR, tracking it separately in giantswarm/giantswarm#37210 (bound the persisters with `dehydrateOptions`/`throttleTime`/a size cap across both gs and agent-platform) and leaving persistence as-is here for consistency.

PR #1957, opened 2026-07-18
fix(agent-platform): query only confirmed-healthy installations for fleet lists

marians:
### What does this PR do?

Narrows the agent-platform fleet queries to installations whose cluster access is **confirmed healthy**, distinguishes "kagent isn't installed here" from a real read failure, and gives slow-but-healthy clusters more headroom before they're flagged degraded.

- `useReachableInstallations` now treats **only `healthy`** as reachable (drops `connecting`). On load the `ClusterAccessConnector` seeds every installation as `connecting`, so including it fanned queries out to installations that turn out to be degraded/unreachable — each hanging for the full proxy timeout and duplicating the sidebar Cluster-access widget's signalling.
- The agents "couldn't read" card is refined to `errored ∩ currently-healthy`, so a cluster that degrades mid-session drops out of the card (the sidebar widget owns that state).
- A **404 (`NotFoundError`) is treated as a successful empty read**, not a failure: a cluster can be reachable and listable but simply not have the `kagent.dev` API group installed (kagent isn't deployed there). Such an installation now contributes zero Agents / zero ModelConfigs and is **not** flagged in the "couldn't read" card — on both the agents list and the create-flow installation select (shared `ModelConfigsProvider`). Only genuinely actionable failures remain: `403 ForbiddenError` (no permission) and unreachable clusters.
- Raises the cluster-access probe timeout `2s → 5s` so slow-but-healthy clusters aren't flagged `degraded` during load spikes (still under the 10s default proxy timeout).

The Agents list and the create-flow installation dropdown no longer query or flag known-degraded/unreachable installations, so the list stops churning wide-then-narrow on load. The "Couldn't read N installations" notice now shows only genuinely actionable cases — a reachable installation that simply doesn't run kagent shows zero agents with no warning, instead of being reported as unreadable. Fewer transient "degraded" blips in the sidebar Cluster access widget.

### How does it look like?

No visual change beyond fewer spurious entries in the "Couldn't read …" card (kagent-absent clusters no longer listed) and steadier loading.

Implements giantswarm/giantswarm#37209. Note this refines that issue's original wording, which had listed a missing CRD / 404 as an "actionable" case: in practice an activated installation that just doesn't run kagent returns a 404, so it's treated as silently empty rather than a failure.

No.

### DIFF
--- a/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
+++ b/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
@@ -179,10 +179,16 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
       ),
     );
 
-    // Only surface installations we have *nothing* for; if we're still showing
-    // an installation's last-known agents, don't also claim we couldn't read it.
+    // Surface only the genuinely actionable case: a currently-healthy cluster we
+    // have *nothing* for. Intersecting with the reachable (healthy) set means a
+    // cluster that degrades mid-session drops out of the card immediately — the
+    // sidebar Cluster-access widget owns that state, so we'd otherwise duplicate
+    // it. And if we're still showing an installation's last-known agents, don't
+    // also claim we couldn't read it.
+    const reachableSet = new Set(reachableInstallations);
     const unreachableInstallations = erroredInstallations.filter(
-      cluster => !agentsByInstallation[cluster]?.length,
+      cluster =>
+        reachableSet.has(cluster) && !agentsByInstallation[cluster]?.length,
     );
 
     // The fleet-wide query reports "loading" until every installation settles,
@@ -208,6 +214,7 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
     isProbing,
     modelConfigsFor,
     allInstallationsKey,
+    reachableInstallationsKey,
   ]);
 
   return (
--- a/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
+++ b/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
@@ -9,6 +9,7 @@ import {
 import { configApiRef, useApi } from '@backstage/core-plugin-api';
 import {
   Agent,
+  isNotFoundError,
   useResources,
 } from '@giantswarm/backstage-plugin-kubernetes-react';
 import { useReachableInstallations } from '../../hooks/useReachableInstallations';
@@ -113,7 +114,16 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
           )
           .join(',')}`,
     );
-    const failed = errors.map(e => `err:${e.cluster}`);
+    // Include the error discriminator: the reconcile effect classifies by error
+    // name (a 404 is an empty read, anything else is a failure), so a same-cluster
+    // error→error transition that flips the name (404 ⇄ 403) must change the
+    // signature, or the effect wouldn't re-run and would keep the stale verdict.
+    const failed = errors.map(
+      e =>
+        `err:${e.cluster}:${
+          e.type === 'incompatibility' ? 'incompat' : e.error.name
+        }`,
+    );
     return [...ok, ...failed].sort().join('|');
   }, [clustersData, errors]);
 
@@ -124,11 +134,7 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
     // we can list it, there just are no Agents. Genuine failures (403 forbidden,
     // unreachable) still count as errors.
     const notInstalled = new Set(
-      errors
-        .filter(
-          e => e.type !== 'incompatibility' && e.error.name === 'NotFoundError',
-        )
-        .map(e => e.cluster),
+      errors.filter(isNotFoundError).map(e => e.cluster),
     );
 
     // Clusters that responded successfully this render (empty result included),
--- a/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
+++ b/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
@@ -118,8 +118,25 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
   }, [clustersData, errors]);
 
   useEffect(() => {
-    // Clusters that responded successfully this render (empty result included).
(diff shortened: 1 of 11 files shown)


plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
```diff
@@ -118,8 +118,25 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
...
-    const succeeded = new Set(clustersData.map(c => c.cluster));
+    // A 404 means the kagent.dev API group isn't installed on that cluster —
+    // kagent simply isn't deployed there. Treat it as a successful empty read
+    // (zero agents), not a "couldn't read" failure: the cluster is reachable and
+    // we can list it, there just are no Agents. Genuine failures (403 forbidden,
+    // unreachable) still count as errors.
+    const notInstalled = new Set(
+      errors
+        .filter(
+          e => e.type !== 'incompatibility' && e.error.name === 'NotFoundError',
```
marians: **Stale classification: `readSignature` ignores the error _name_ this effect now branches on.**

This effect now treats a `NotFoundError` (404) as a successful empty read and every other error as a "couldn't read" failure. But the effect only re-runs when `readSignature` changes, and the signature summarises errors as just `err:${cluster}` (line 116–117) — it does **not** include `error.name`/`type`.

So an error→error transition on the *same* set of errored clusters that flips the error name is invisible to the signature and the effect never re-runs:
- `grizzly` 404 (kagent not installed) → treated as empty, cleared from the card. RBAC/install changes and it becomes 403 forbidden on a background refetch → signature is still `err:grizzly`, so it is **never surfaced** in the "couldn't read" card despite now being a genuine failure.
- Reverse (403 → 404) leaves a cluster wrongly stuck in the card.

Fix: fold the discriminator into the signature, e.g. `err:${e.cluster}:${e.type === 'incompatibility' ? 'incompat' : e.error.name}`. (`ModelConfigsProvider` derives directly in `useMemo` with no sticky signature, so it does not have this bug — the asymmetry is the tell.)

plugins/agent-platform/src/components/ModelConfigsProvider/ModelConfigsProvider.tsx
```diff
@@ -64,9 +64,18 @@ export function ModelConfigsProvider({ children }: { children: ReactNode }) {
...
+    // A 404 means the kagent.dev API group isn't installed on that cluster, so it
+    // simply has no ModelConfigs — not a "couldn't read" failure. Only genuine
+    // failures (403 forbidden, unreachable) that produced no models are surfaced.
     const unreachableInstallations = Array.from(
-      new Set(errors.map(e => e.cluster)),
+      new Set(
+        errors
+          .filter(
+            e =>
+              e.type === 'incompatibility' || e.error.name !== 'NotFoundError',
```
marians: **Altitude: the "404 = kagent-not-installed, not a failure" rule is duplicated in two providers as inverted boolean expressions.**

This predicate — keep an error unless it is a non-incompatibility `NotFoundError` — is the exact De Morgan inverse of the one in `AgentsDataProvider.tsx` (`e.type !== 'incompatibility' && e.error.name === 'NotFoundError'`). Two copies expressing the same domain fact through inverted logic is a maintenance hazard: a future change (e.g. also treat a 404 on a specific sub-path differently, or add another "not really an error" status) has to be made in two places and negated correctly in one of them.

Prefer a single shared predicate, e.g. `isKagentNotInstalled(e)` co-located with the error type in `kubernetes-react`, and call it (or its negation) from both providers.

plugins/agent-platform/src/components/ModelConfigsProvider/ModelConfigsProvider.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+}: {
+  succeeded?: Record<string, number>;
+  failed?: string[];
+  notFound?: string[];
+  isLoading?: boolean;
+}) {
+  const resources = Object.entries(succeeded).flatMap(([cluster, count]) =>
+    Array.from({ length: count }, () => fakeModelConfig(cluster)),
+  );
+  const errors = [
```
marians: **Cleanup: the `result()` fixture builder is copy-pasted from `AgentsDataProvider.test.tsx`.**

The error-shape construction here (`failed` → `{ error: { name: 'ForbiddenError' } }`, `notFound` → `{ error: { name: 'NotFoundError' } }`) is duplicated almost verbatim from the sibling test. Since both encode the same `useResources` error contract, a future shape change (e.g. errors gaining a `type`/`retry` field the code starts reading) must be updated in both. Consider a shared test helper for the error-fixture mapping.

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
marians: Real bug — fixed in 8fbae4e4. Folded the discriminator into `readSignature`: `err:${e.cluster}:${e.type === 'incompatibility' ? 'incompat' : e.error.name}`. Now a same-cluster 404 ⇄ 403 transition changes the signature, so the reconcile effect re-runs and reclassifies. Added a regression test (`reclassifies a cluster when its error flips 404 → 403 on a refetch`) that fails without the signature change.

plugins/agent-platform/src/components/ModelConfigsProvider/ModelConfigsProvider.tsx
marians: Fixed in 8fbae4e4 — extracted a single `isNotFoundError(errorInfo)` predicate co-located with `ErrorInfoUnion` in `kubernetes-react` (`hooks/utils/queries.ts`, exported from the package). Both providers now call it (`errors.filter(isNotFoundError)` in AgentsDataProvider, `errors.filter(e => !isNotFoundError(e))` in ModelConfigsProvider), so the domain fact lives in one place. Added a direct unit test for the predicate (404 true; 403/generic/incompatibility false).

plugins/agent-platform/src/components/ModelConfigsProvider/ModelConfigsProvider.test.tsx
marians: Fixed in 8fbae4e4 — extracted the error-fixture mapping into a shared `buildResourceErrors({ failed, notFound })` helper (`plugins/agent-platform/src/components/resourceErrorFixtures.ts`), used by both providers' `result()` builders. A future error-shape change is now a one-place edit.

PR #1958, opened 2026-07-18
feat(gs): mute installations app-wide from the Cluster access popover

marians:
### What does this PR do?

Adds a per-installation **on/off toggle** to the sidebar **Cluster access** popover. A *muted* installation is switched off for the whole app:

- the `ClusterAccessConnector` stops probing it (and drops it from the access-status set);
- fleet views stop fetching from it — the shared `InstallationPicker` folds the muted set into `disabledClusters` (which `MultipleClustersSelector` already excludes from the active/fetched set), covering the **Clusters** and **Deployments** pages; and the **agent-platform** lists exclude it automatically (muted → not probed → not `healthy` → not reachable).

The preference is user intent, so it is persisted per-browser (`localStorage`, key `gs-muted-installations`) and synced across tabs — deliberately distinct from the health-based `useDisabledInstallations` hook, which auto-detects unreachable backends.

Implemented as a new `MutedInstallationsApi` / `MutedInstallationsStore` (mirrors the cluster-access status store, but persisted), registered in the gs plugin and exported. The toggle itself is a bui (`@backstage/ui`) `Switch`.

Also fixes a pre-existing quirk surfaced by the toggle: a successful broker cluster-token mint recorded the installation as **healthy** (green) even when the apiserver was unreachable, so un-muting an unreachable installation flashed green before the `/version` probe corrected it to degraded. Health is now owned solely by the authoritative `/version` probe; the mint still records the precise broker-level failure states (session-expired / degraded), which the probe defers to.

Users can turn individual installations off from the Cluster access popover. A muted installation is no longer probed and no longer appears / is fetched in the Clusters, Deployments, or Agent Platform views. It shows as an "Off" row in the popover and can be switched back on at any time. The setting persists across reloads. Un-muting an unreachable installation now shows connecting → degraded, with no misleading green flash (this also removes the brief false-green during initial app load).

### How does it look like?

image

Follow-up to the agent-platform reachability work (#1957). The toggle deliberately lives in the shared Cluster access widget rather than the agent-platform page so it applies app-wide.

No.

- [ ] A changeset describing the change and affected packages was added. (`@giantswarm/backstage-plugin-gs` is `private: true`, not published to npm — no changeset needed; released via the app image/chart auto-release.)

### DIFF
--- a/plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
+++ b/plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
@@ -18,10 +18,12 @@ export class MutedInstallationsStore implements MutedInstallationsApi {
 
   private constructor() {
     this.muted = MutedInstallationsStore.load();
-    // Keep multiple tabs consistent: another tab writing the key reloads ours.
+    // Keep multiple tabs consistent: reload when another tab writes our key.
+    // `event.key` is `null` when another tab calls `localStorage.clear()` (and
+    // for some bulk changes), so treat that as "storage changed, re-read".
     if (typeof window !== 'undefined') {
       window.addEventListener('storage', event => {
-        if (event.key === STORAGE_KEY) {
+        if (event.key === STORAGE_KEY || event.key === null) {
           this.muted = MutedInstallationsStore.load();
           this.notify();
         }
--- a/plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
+++ b/plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
@@ -1,0 +1,103 @@
+import { Observable } from '@backstage/types';
+import ObservableImpl from 'zen-observable';
+import { MutedInstallationsApi } from './types';
+
+const STORAGE_KEY = 'gs-muted-installations';
+
+/**
+ * localStorage-backed, multicast implementation of {@link MutedInstallationsApi}.
+ *
+ * Unlike the in-memory cluster-access status store, this holds user intent that
+ * must survive reloads, so it persists to localStorage (and syncs across tabs via
+ * the `storage` event). A single process-wide Set + a Set of subscribers is
+ * enough — the muted set is tiny (a handful of installation names).
+ */
+export class MutedInstallationsStore implements MutedInstallationsApi {
+  private muted: Set<string>;
+  private readonly subscribers = new Set<(snapshot: string[]) => void>();
+
+  private constructor() {
+    this.muted = MutedInstallationsStore.load();
+    // Keep multiple tabs consistent: another tab writing the key reloads ours.
+    if (typeof window !== 'undefined') {
+      window.addEventListener('storage', event => {
+        if (event.key === STORAGE_KEY) {
+          this.muted = MutedInstallationsStore.load();
+          this.notify();
+        }
+      });
+    }
+  }
+
+  static create(): MutedInstallationsApi {
+    return new MutedInstallationsStore();
+  }
+
+  private static load(): Set<string> {
+    try {
+      const raw = window.localStorage.getItem(STORAGE_KEY);
+      if (!raw) {
+        return new Set();
+      }
+      const parsed = JSON.parse(raw);
+      return Array.isArray(parsed)
+        ? new Set(parsed.filter((x): x is string => typeof x === 'string'))
+        : new Set();
+    } catch {
+      // Private-mode / quota / parse errors: fall back to "nothing muted".
+      return new Set();
+    }
+  }
+
+  private persist(): void {
+    try {
+      window.localStorage.setItem(
+        STORAGE_KEY,
+        JSON.stringify(this.getSnapshot()),
+      );
+    } catch {
+      // Best-effort; the in-memory set still reflects the change for this tab.
+    }
+  }
+
+  isMuted(installation: string): boolean {
+    return this.muted.has(installation);
+  }
+
+  setMuted(installation: string, muted: boolean): void {
+    if (muted === this.muted.has(installation)) {
+      return;
+    }
+    if (muted) {
+      this.muted.add(installation);
+    } else {
+      this.muted.delete(installation);
+    }
+    this.persist();
+    this.notify();
+  }
+
+  toggle(installation: string): void {
+    this.setMuted(installation, !this.muted.has(installation));
+  }
+
+  getSnapshot(): string[] {
+    return [...this.muted].sort((a, b) => a.localeCompare(b));
+  }
+
+  muted$(): Observable<string[]> {
+    return new ObservableImpl(subscriber => {
+      const notify = (snapshot: string[]) => subscriber.next(snapshot);
+      subscriber.next(this.getSnapshot());
+      this.subscribers.add(notify);
+      return () => {
+        this.subscribers.delete(notify);
+      };
+    });
+  }
+
+  private notify(): void {
+    const snapshot = this.getSnapshot();
+    this.subscribers.forEach(notify => notify(snapshot));
+  }
+}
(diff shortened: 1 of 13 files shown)


plugins/gs/src/components/ClusterAccessStatus/ClusterAccessConnector.tsx
```diff
@@ -121,9 +122,30 @@ export function ClusterAccessConnector() {
...
+  // Track the user's muted set reactively so muting/unmuting re-runs the probe
+  // effects (a muted installation must stop being probed and drop off the
+  // status set; unmuting must resume probing). The store emits only on real
+  // changes, so `muted`'s identity is stable between changes.
+  const [muted, setMuted] = useState<string[]>(() => mutedApi.getSnapshot());
   useEffect(() => {
-    const installations = authProvidersApi.getBrokerCoveredInstallations();
+    const subscription = mutedApi.muted$().subscribe(setMuted);
+    return () => subscription.unsubscribe();
+  }, [mutedApi]);
```
marians: **Startup double-probe: the initial `muted$` replay restarts the whole fleet probe.** The comment (l.127–130) says "the store emits only on real changes, so `muted`'s identity is stable between changes" — but `muted$()` replays `getSnapshot()` on subscribe, and `getSnapshot()` returns a **new array reference** every call (`[...this.muted].sort(...)`). So on mount: `useState(() => mutedApi.getSnapshot())` seeds array A, the effect runs and `connectAll()` starts probing the whole fleet; then the subscription immediately fires `setMuted(getSnapshot())` = array B (same contents, new ref). Since `muted` is a dependency of both probe effects (l.235, l.283) and `B !== A`, React tears down and re-runs them, cancelling the in-flight probes and calling `connectAll()` a second time. Given how much care the probe path takes to stay cheap (`PROBE_TIMEOUT_MS`, bounded concurrency), doubling the startup fleet probe is a real regression. Consider comparing snapshots by content (skip `setMuted` when the sorted arrays are equal) or gating the effect on a stable key like `muted.join('\n')`.

plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
```diff
@@ -0,0 +1,103 @@
...
+export class MutedInstallationsStore implements MutedInstallationsApi {
+  private muted: Set<string>;
+  private readonly subscribers = new Set<(snapshot: string[]) => void>();
+
+  private constructor() {
+    this.muted = MutedInstallationsStore.load();
+    // Keep multiple tabs consistent: another tab writing the key reloads ours.
+    if (typeof window !== 'undefined') {
+      window.addEventListener('storage', event => {
+        if (event.key === STORAGE_KEY) {
```
marians: **Cross-tab sync misses `localStorage.clear()`.** The handler only reloads when `event.key === STORAGE_KEY`, but the `storage` event fires with `key === null` when another tab calls `localStorage.clear()` (and some browsers also use `null` for bulk changes). In that case this tab keeps its stale muted set instead of resetting, so the "synced across tabs" guarantee has a hole. Consider also reloading when `event.key === null` (treat it as "storage changed, re-read our key").

plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export class MutedInstallationsStore implements MutedInstallationsApi {
+  private muted: Set<string>;
+  private readonly subscribers = new Set<(snapshot: string[]) => void>();
+
+  private constructor() {
+    this.muted = MutedInstallationsStore.load();
+    // Keep multiple tabs consistent: another tab writing the key reloads ours.
+    if (typeof window !== 'undefined') {
+      window.addEventListener('storage', event => {
```
marians, line 25: **`storage` listener is never removed.** The constructor adds a `window` `storage` listener with no corresponding `removeEventListener`. In production the store is a process-wide singleton so this is harmless, but every `MutedInstallationsStore.create()` in the test suite (each `renderSidebar` / `InstallationPicker` test) adds another permanent listener to the shared jsdom `window`. The `MutedInstallationsStore.test.ts` cross-tab test's `dispatchEvent` then fans out to every store created by earlier tests. It's benign today only because `beforeEach` clears `localStorage`, but it's a latent cross-test coupling. Worth noting since the store is otherwise cleanly unit-testable.

plugins/gs/src/components/ClusterAccessStatus/ClusterAccessStatusSidebarItem.tsx
```diff
@@ -151,26 +163,68 @@ function StatusDot({ color }: { color: string }) {
...
+        installation,
+        dotColor: STATE_COLORS[entry.state],
+        caption,
+        enabled: true,
+      };
+    });
+  }, [entries, muted]);
+
+  // Nothing accessed and nothing muted -- keep the sidebar uncluttered.
+  if (entries.length === 0 && muted.length === 0) {
```
marians, line 232: **All-muted shows a green badge reading "All 0 healthy".** The popover now renders when `entries.length === 0 && muted.length > 0`, but `summarize(entries)` and `overallState(entries)` still ignore `muted`. If the user mutes every installation, `entries` is empty, so `summarize([])` hits `counts.healthy === total` with `total === 0` and returns `[{ label: 'All 0 healthy' }]`, and `overallState([])` falls through to `'healthy'` → a green badge. So a fully-switched-off state advertises "All 0 healthy" in green, the opposite of reality. Consider special-casing `total === 0` (neutral badge / "All off" summary) when there are muted rows but no live entries.

marians (commented):
Code review (medium effort, correctness-focused). 4 findings posted inline — one efficiency regression (startup double-probe), one cosmetic-correctness (all-muted green badge), one cross-tab edge case, and one test-coupling note. No blocking bugs found; the core mute/unmute flow and the `disabledClusters` union look correct.

plugins/gs/src/components/ClusterAccessStatus/ClusterAccessConnector.tsx
marians: Good catch — fixed in 95602d89. Added a shared `useMutedInstallations()` hook that returns a **content-stable** array reference (it returns the previous array when the sorted contents are unchanged), so the initial `muted$()` replay no longer changes `muted`'s identity and the probe effects don't re-run / restart the fleet probe on mount. The connector, sidebar widget, and installation picker all use it now (also de-duplicates the three copies of the subscribe-to-muted logic).

plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
marians: Fixed in 95602d89 — the `storage` handler now reloads on `event.key === STORAGE_KEY || event.key === null`, so another tab's `localStorage.clear()` (which fires with `key === null`) resets this tab's muted set. Added a test covering the `key === null` case.

plugins/gs/src/apis/mutedInstallations/MutedInstallationsStore.ts
marians: Acknowledged — leaving as-is intentionally. In production the store is an app-lifetime singleton (one instance via the `ApiBlueprint` factory), so the listener living for the app's lifetime is correct and there's no natural teardown hook. The cross-test coupling is real but benign: `beforeEach` clears `localStorage`, so a fanned-out `storage` event just reloads every store to the same (cleared/seeded) value. Adding a `dispose()` purely for test hygiene would widen the API for no production benefit, so I'd rather not.

plugins/gs/src/components/ClusterAccessStatus/ClusterAccessStatusSidebarItem.tsx
marians: Fixed in 95602d89. When there are no live entries but some muted installations, the badge now uses a neutral grey (`MUTED_COLOR`) instead of green, and the summary shows "N off" instead of "All 0 healthy". `summarize`/`overallState` are unchanged (still entries-only); the muted case is handled in the component. Added a test asserting the all-muted state shows "N off" and not "All 0 healthy".

PR #1966, opened 2026-07-21
fix(gs): load installations configuration from an authenticated backend endpoint

marians:
### What does this PR do?

Changes how the frontend obtains the `gs.installations` configuration. Instead of reading it from the frontend app-config, the `gs-backend` plugin now exposes it through an authenticated endpoint (`GET /api/gs/installations`), and the frontend loads it once, after sign-in, via a small async provider.

To support this, the boot-time frontend APIs that previously read `gs.installations` synchronously from config (`DiscoveryApiClient`, `KubernetesClient`, `GSAuthProviders`) were refactored to obtain the data asynchronously from a shared source, and the per-installation auth providers are now initialized lazily once the main sign-in completes. The `gs.installations` config schema is changed to backend visibility.

Also includes a fix for a pre-sign-in deadlock found during local testing: auth base-URL resolution no longer blocks on the not-yet-loaded installations data, and the main sign-in provider is no longer treated as installation-scoped.

No user-facing behaviour change. Installation configuration is loaded transparently after authentication.

Context is tracked in the internal issue giantswarm/giantswarm#37213 (private).

This is one of two alternative implementations under consideration; it is opened as a **draft** for review and manual verification. It has been smoke-tested locally (sign-in and app load).

No.

> Note: this branch changes public plugin exports, so a changeset is still to be added before the `@giantswarm/backstage-plugin-*` packages are published.

### DIFF
--- a/plugins/gs/src/apis/auth/GSAuthProviders.ts
+++ b/plugins/gs/src/apis/auth/GSAuthProviders.ts
@@ -121,7 +121,12 @@ export class GSAuthProviders implements GSAuthProvidersApi {
       ? mainProviderName.split(OIDC_PROVIDER_NAME_PREFIX)[1]
       : mainProviderName;
 
-    return this.createKubernetesAuthApi(mainProviderName, providerDisplayName);
+    return this.createKubernetesAuthApi(
+      mainProviderName,
+      providerDisplayName,
+      undefined,
+      true,
+    );
   }
 
   /**
@@ -342,6 +347,7 @@ export class GSAuthProviders implements GSAuthProvidersApi {
     providerName: string,
     providerDisplayName: string,
     clusterTokenProvider?: () => Promise<ClusterToken | undefined>,
+    isMainProvider = false,
   ): AuthApi {
     const authConnector = new DefaultAuthConnector({
       configApi: this.configApi,
@@ -355,6 +361,7 @@ export class GSAuthProviders implements GSAuthProvidersApi {
         icon: GiantSwarmIcon,
       },
       clusterTokenProvider,
+      isMainProvider,
       sessionTransform({ backstageIdentity, ...res }): OAuth2Session {
         const session: OAuth2Session = {
           ...res,
--- a/plugins/gs/src/apis/auth/GSAuthProviders.ts
+++ b/plugins/gs/src/apis/auth/GSAuthProviders.ts
@@ -140,13 +140,22 @@ export class GSAuthProviders implements GSAuthProvidersApi {
     }
     if (!this.initPromise) {
       this.initPromise = (async () => {
-        const installations = await getInstallationsConfig();
-        this.kubernetesAuthProviders =
-          this.buildKubernetesAuthProviders(installations);
-        this.kubernetesAuthApis = this.buildKubernetesAuthApis(
-          this.kubernetesAuthProviders,
-        );
-        this.initialized = true;
+        try {
+          const installations = await getInstallationsConfig();
+          this.kubernetesAuthProviders =
+            this.buildKubernetesAuthProviders(installations);
+          this.kubernetesAuthApis = this.buildKubernetesAuthApis(
+            this.kubernetesAuthProviders,
+          );
+          this.initialized = true;
+        } catch (error) {
+          // Never latch a rejected promise: a permanently-cached rejection
+          // would make every later `ensureInitialized()` re-await the same
+          // failure and break k8s OIDC auth fleet-wide. Reset so a later call
+          // can retry.
+          this.initPromise = undefined;
+          throw error;
+        }
       })();
     }
     await this.initPromise;
@@ -155,13 +164,19 @@ export class GSAuthProviders implements GSAuthProvidersApi {
   private buildKubernetesAuthProviders(
     installations: Awaited<ReturnType<typeof getInstallationsConfig>>,
   ): AuthProvider[] {
-    return installations.map(installation => {
+    // A single malformed installation entry must not take down auth for the
+    // whole fleet: skip entries with a non-`oidc` auth provider or a
+    // missing/invalid OIDC token provider (logging a warning) so the valid
+    // installations still build.
+    return installations.flatMap(installation => {
       const installationName = installation.name;
       const authProvider = installation.authProvider;
       if (!authProvider || authProvider !== 'oidc') {
-        throw new Error(
-          `OIDC auth provider is not configured for installation "${installationName}".`,
+        // eslint-disable-next-line no-console
+        console.warn(
+          `Skipping installation "${installationName}": OIDC auth provider is not configured.`,
         );
+        return [];
       }
       const oidcTokenProvider = installation.oidcTokenProvider;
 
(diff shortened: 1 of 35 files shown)


plugins/gs/src/components/InstallationsConfigLoader/InstallationsConfigLoader.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const identityApi = useApi(identityApiRef);
+  const errorApi = useApi(errorApiRef);
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const load = async () => {
+      // Blocks until the app is signed in, so the authenticated endpoint does
+      // not 401.
+      await identityApi.getCredentials();
```
marians: **Pre-fetch awaits are outside the retry/degradation `try/catch` — a rejection here hangs every boot-time API forever.**

`load()` is invoked without a `.catch`, and `await identityApi.getCredentials()` (line 48) and `await discoveryApi.getBaseUrl('gs')` (line 53) run *before* the `for` loop's `try/catch`. If either rejects (e.g. sign-in errors out, discovery throws), `load()` rejects as an unhandled promise rejection and `setInstallationsConfig(...)` is **never** called. Every consumer that `await getInstallationsConfig()` — `KubernetesClient.getClusters`, `GSAuthProviders.ensureInitialized`, `DiscoveryApiClient` overrides — then waits on a promise that never resolves. The retry/`setInstallationsConfig([])` fallback only covers the fetch, not these two awaits.

Suggest moving both awaits inside a `try/catch` that publishes `[]` (and posts the error) on failure, so the app degrades to "no installations" instead of deadlocking.

plugins/gs/src/apis/auth/GSAuthProviders.ts
```diff
@@ -93,25 +103,67 @@ export class GSAuthProviders implements GSAuthProvidersApi {
...
+   * Lazily builds the per-installation kubernetes auth providers/APIs once the
+   * installations config has loaded from the backend. Safe to call repeatedly;
+   * the work runs at most once.
+   */
+  async ensureInitialized(): Promise<void> {
+    if (this.initialized) {
+      return;
+    }
+    if (!this.initPromise) {
+      this.initPromise = (async () => {
```
marians, line 142: **A throw in `buildKubernetesAuthProviders` permanently poisons the lazy init — no recovery for the whole session.**

`buildKubernetesAuthProviders` throws when any installation has a non-`oidc` `authProvider` or a missing/invalid `oidcTokenProvider`. Because that runs inside the IIFE assigned to `this.initPromise`, the throw leaves `initPromise` a **rejected** promise with `initialized` still `false`. Every later `ensureInitialized()` call hits `if (!this.initPromise)` = false and re-`await`s the same rejected promise, so it can never succeed again.

Downstream: `getKubernetesAuthApi()` rejects for **all** clusters (k8s OIDC auth broken fleet-wide), and the `ensureInitialized().then(_, () => {})` gates in `ClusterAccessConnector`/`ProviderSettings` never flip ready. One malformed installation entry silently disables cluster auth + the cluster-access widget for the entire session. Previously this same misconfig threw loudly at construction (app boot). Consider not caching the rejection (reset `initPromise` on failure) and/or skipping bad installations instead of throwing.

plugins/gs/src/components/ClusterAccessStatus/ClusterAccessConnector.tsx
```diff
@@ -132,7 +132,30 @@ export function ClusterAccessConnector() {
...
+  const [authProvidersReady, setAuthProvidersReady] = useState(false);
   useEffect(() => {
+    let cancelled = false;
+    authProvidersApi.ensureInitialized().then(
+      () => {
+        if (!cancelled) {
+          setAuthProvidersReady(true);
+        }
+      },
+      () => {},
```
marians: **Empty rejection handler swallows `ensureInitialized()` failures → the widget stays blank forever with no signal.**

The `() => {}` rejection arm means that if `ensureInitialized()` ever rejects (see the rejected-`initPromise` caching in `GSAuthProviders`), `authProvidersReady` never becomes `true`, so both probe effects permanently early-return and the cluster-access widget renders empty with nothing logged or surfaced to the user. `ProviderSettings.tsx:22` has the identical pattern (settings auth list stays blank). At minimum log/report the error so a permanently-empty widget isn't indistinguishable from "no clusters".

plugins/gs/src/apis/kubernetes/KubernetesClient.ts
```diff
@@ -131,29 +130,30 @@ export class KubernetesClient implements KubernetesApi {
...
-      throw new Error(`Missing gs.installations configuration`);
-    }
-
-    const installations = this.configApi.getConfig('gs.installations').keys();
+    // Installations are loaded from the authenticated backend endpoint after
+    // sign-in (they are no longer in the frontend config). `getClusters` is
+    // only ever called post-sign-in (the clusters/deployments pages), so
+    // awaiting the source here does not stall app boot.
+    const installations = await getInstallationsConfig();
     this.clusters = installations.map(installation => {
```
marians: **The empty-installations fallback is sticky for the whole session — this cache locks it in.**

If the loader exhausts its 5 retries it publishes `setInstallationsConfig([])`; `getClusters()` then caches `this.clusters = []` here. There is no periodic re-fetch, and `GSAuthProviders.initialized` similarly latches. So a transient backend hiccup during the (~15s) retry window degrades the app to "no clusters / no per-installation auth" for the rest of the session, recoverable only by a full reload — even though the backend is fine seconds later. Worth considering a background re-fetch, or not caching `this.clusters` when the loaded set is empty.

plugins/gs/src/apis/discovery/DiscoveryApiClient.ts
```diff
@@ -22,37 +18,57 @@ export class DiscoveryApiClient implements DiscoveryApi {
   async getBaseUrl(pluginId: string, installation?: string) {
     const selectedInstallation =
       installation || DiscoveryApiClient.getInstallation();
```
marians, line 43: **Main-provider auth can still be mis-scoped via the static `getInstallation()` fallback, defeating the new `isMainProvider` guard.**

`isMainProvider` makes `DefaultAuthConnector.installationId()` return `undefined` so the main provider resolves discovery with `getBaseUrl('auth', undefined)`. But here `getBaseUrl` does `installation || DiscoveryApiClient.getInstallation()`, falling back to the module-static "current installation". `setInstallation(...)` is set during scaffolder (`ScaffolderApiClient.withInstallation`) and agent-platform flows; if a main-provider token refresh fires while one of those is active and that installation has a `backendUrl` override (once the snapshot is loaded), the main sign-in/refresh gets routed to the per-installation backend — reintroducing exactly the scoping this change tried to remove. Consider having the main connector pass an explicit sentinel that bypasses the static fallback.

plugins/gs/src/components/InstallationsConfigLoader/InstallationsConfigLoader.tsx
marians: Fixed

plugins/gs/src/apis/auth/GSAuthProviders.ts
marians: Fixed

plugins/gs/src/components/ClusterAccessStatus/ClusterAccessConnector.tsx
marians: Fixed

plugins/gs/src/apis/kubernetes/KubernetesClient.ts
marians: Fixed

plugins/gs/src/apis/discovery/DiscoveryApiClient.ts
marians: Addressed in 4222faa8. Introduced an exported `NO_INSTALLATION` sentinel: `DefaultAuthConnector.buildUrl` now passes `NO_INSTALLATION` to `getBaseUrl` when `isMainProvider` (instead of `installationId()`), and `getBaseUrl` short-circuits on the sentinel to the default `urlPatternDiscovery` URL — bypassing both the static `getInstallation()` fallback and any per-installation `backendUrl` override. So a main-provider token refresh can no longer be re-routed to a per-installation backend even while a scaffolder / agent-platform flow has set the static current installation. `installationId()` still returns the installation *name* (`string | undefined`) since `mintBrokerSession` relies on it; only the discovery scoping decision moved to the sentinel. Covered by a new `DiscoveryApiClient` test (static installation + override set → sentinel resolves the default URL; a genuine per-installation call still applies the override).

PR #1969, opened 2026-07-21
feat(plans): revamp the Plans page with bui components and shared markdown rendering

marians:
### What does this PR do?

Migrates the internal **Plans** page from Material-UI to the bui design system
(`@backstage/ui`), and extracts a shared markdown renderer.

- **Shared markdown** — new `GSMarkdownContent` in `ui-react` wraps
  core-components' `MarkdownContent` (GFM default) with consistent typography:
  matched `<p>`/`<li>` line-height, spacing between list items, and padded
  non-highlighted code blocks (language-tagged blocks keep their `CodeSnippet`
  styling). The `gs` README/SOUL cards (`CollapsibleMarkdownCard`) and the plans
  document/PR-body/comment rendering all route through it.
- **Header & tabs** — Proposed/Merged move to bui `Tabs`; the repository picker
  becomes a bui `Select` placed in the plugin header via the shared
  page-header-actions slot (instead of a body toolbar).
- **Lists** — proposed, merged-folder, and review-nav lists rebuilt on bui
  `List`/`ListRow`.
- **Merged tab** — each document renders in a bui `Accordion` (expanded by
  default); dot files/folders and loose repository-root documents are hidden;
  the redundant folder prefix is stripped from titles; well-known files get
  friendly labels (e.g. `PRD.md` → "Product Requirements Document (PRD.md)",
  `index.html` → "Web page (index.html)").
- Comments, chips (→ `Badge`), alerts, and the Rendered/Diff toggle move to bui
  equivalents.

The Plans page adopts the bui look-and-feel: the repository selector sits in the
page header, tabs and lists match the rest of the modernised UI, and merged
plans are browsable as a stack of collapsible documents with friendlier titles.

### How does it look like?

Verified in the running app via the DOM (header Select, bui tabs, List rows with
hover affordances, Accordion group expanded by default, code-block styling for
bare vs. highlighted blocks).

The `plans` plugin was the last Plans surface still fully on MUI v4. Markdown
rendering was previously duplicated across plugins with ad-hoc `<p>` fixes; this
consolidates the common case into `ui-react` (follow-ups: adopt
`GSMarkdownContent` in muster/roadmap and migrate EpicChip's remaining MUI
`Tooltip`).

No.

### DIFF
--- a/plugins/plans/src/components/MergedTab/MergedTab.tsx
+++ b/plugins/plans/src/components/MergedTab/MergedTab.tsx
@@ -1,46 +1,53 @@
 import { useMemo } from 'react';
 import { useSearchParams } from 'react-router-dom';
+import type { Selection } from 'react-aria-components';
 import {
-  Box,
-  Grid,
+  Accordion,
+  AccordionGroup,
+  AccordionPanel,
+  AccordionTrigger,
+  Alert,
+  Flex,
   List,
-  ListItem,
-  ListItemSecondaryAction,
-  ListItemText,
-  Paper,
-  Typography,
-  makeStyles,
-  Theme,
-} from '@material-ui/core';
-import { Alert } from '@material-ui/lab';
+  ListRow,
+} from '@backstage/ui';
+import { makeStyles, Theme } from '@material-ui/core';
 import { EmptyState, Progress } from '@backstage/core-components';
 import { useApi } from '@backstage/frontend-plugin-api';
 import { useQuery } from '@tanstack/react-query';
 import { plansApiRef } from '../../apis';
-import { compareDisplayPaths, isRenderableFile } from '../../lib/files';
+import {
+  compareDisplayPaths,
+  friendlyFileName,
+  isDotPath,
+  isRenderableFile,
+  stripFolderPrefix,
+} from '../../lib/files';
 import { EpicChip } from '../EpicChip';
 import { PlanFileContent } from '../PlanFileContent';
 
-const ROOT_GROUP = '(repository root)';
-
 const useStyles = makeStyles((theme: Theme) => ({
-  listPanel: {
-    padding: 0,
-  },
-  detailPanel: {
-    padding: theme.spacing(2),
-  },
-  fileSection: {
-    marginTop: theme.spacing(2),
-    '&:first-child': {
-      marginTop: 0,
+  layout: {
+    display: 'grid',
+    gridTemplateColumns: '1fr',
+    gap: theme.spacing(2),
+    alignItems: 'start',
+    [theme.breakpoints.up('md')]: {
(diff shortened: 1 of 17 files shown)


plugins/plans/src/components/MergedTab/MergedTab.tsx
```diff
@@ -94,27 +101,37 @@ export function MergedTab({ repo }: { repo: string }) {
...
         continue;
       }
+      // Hide dot files/folders (e.g. `.agents/…`) and loose repository-root
+      // documents — a plan is a top-level folder of documents.
+      if (isDotPath(entry.path)) {
+        continue;
+      }
       const slash = entry.path.indexOf('/');
-      const folder = slash === -1 ? ROOT_GROUP : entry.path.slice(0, slash);
+      if (slash === -1) {
```
marians, line 110: Since this now **skips** loose repository-root documents (`slash === -1 → continue`) and `ROOT_GROUP` is gone, the function's docstring above ("loose root documents are grouped under a synthetic root entry") is now stale and describes removed behavior. Worth updating so the next reader isn't misled.

plugins/plans/src/lib/files.test.ts
```diff
@@ -86,3 +89,56 @@ describe('compareDisplayPaths', () => {
...
+  it('removes the folder prefix from a path within it', () => {
+    expect(stripFolderPrefix('my-plan/README.md', 'my-plan')).toBe('README.md');
+    expect(stripFolderPrefix('my-plan/sub/deep.md', 'my-plan')).toBe(
+      'sub/deep.md',
+    );
+  });
+
+  it('leaves paths outside the folder unchanged', () => {
+    // Root-level documents grouped under a synthetic folder name.
+    expect(stripFolderPrefix('README.md', '(repository root)')).toBe(
```
marians: Minor: this test case covers the `(repository root)` synthetic folder, but `MergedTab` no longer produces that group — loose root documents are now skipped entirely (`slash === -1 → continue`) and `ROOT_GROUP` was removed. `stripFolderPrefix` is still correct and the test is harmless, but it now documents a code path that production can't reach. Not blocking.

plugins/plans/src/components/MergedTab/MergedTab.tsx
marians: Good catch — updated the docstring in 370241ec to say loose repository-root documents and dot files/folders are hidden.

plugins/plans/src/lib/files.test.ts
marians: Reworded the test comment in 370241ec to describe the generic "path not under the given folder" case instead of the removed synthetic root group. The assertion stays (stripFolderPrefix is still used generally).

PR #1971, opened 2026-07-21
feat(agent-platform): show agent avatar icons in creation preview and list

marians:
### What does this PR do?

Wires the Backstage frontend to the self-hosted DiceBear avatar service (now deployed on the fleet) so agents show a deterministic identifying icon in the two places the epic prototype shows:

1. **Agent creation** (`/agents/new`) — a debounced live preview next to the name field, seeded by the **technical name (slug)** and served from the no-cache `/v1/preview/…` route, so throwaway seeds typed while naming never enter any cache. It renders as soon as a name is typed, using the selected installation if picked, else the first reachable one (DiceBear output is identical across installations).
2. **Agent list** (`/agents`) — the avatar next to each agent's name and description, seeded by the technical (resource) name.

The avatar URL is built client-side from the installation's base domain (`avatars.<baseDomain>/v1/[preview/][<size>/]<name>.png`), mirroring the existing Grafana link pattern (`grafana.<baseDomain>`). No manifest/`spec.iconUrl` changes — that's the `giantswarm/agent` chart's job.

Agents are now visually identifiable by a distinct icon in the agent list and while creating an agent, instead of being plain rows of text.

### How does it look like?

- **Creation**: a ~72px avatar preview sits to the left of the Name/Slug fields and updates (debounced) as you type.
- **List**: a ~40px avatar (≈two lines of text tall), top-aligned, next to each agent's name + description.

image

image

- Epic: giantswarm/giantswarm#37211
- The renderer + canonical URL contract (incl. the `/preview` no-cache route and the `{48,96,128,512}` size allowlist) shipped separately via the [`giantswarm/dicebear`](https://github.com/giantswarm/dicebear) chart and the `agentic-platform` bundle.

**CSP note:** the images load directly from the public per-installation endpoint, so each deployment must allowlist its own `https://avatars.<baseDomain>` in `backend.csp.img-src` (where `gs.installations` is configured). Rather than relaxing `img-src` broadly, this requirement is documented in `app-config.yaml` and `app-config.local.yaml.example`. Customer base domains are intentionally kept out of this checked-in repo.

No.

No changeset — the `agent-platform` plugin is `private: true` and not published to npm.


plugins/agent-platform/src/components/AgentAvatarPreview/AgentAvatarPreview.tsx
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function AgentAvatarPreview() {
+  const { state } = useNewAgentForm();
+  const { installations } = useInstallations();
+  const buildAvatarUrl = useAgentAvatarUrl();
+
+  const { installations: reachable } = useReachableInstallations(
+    installations.map(i => i.name),
+  );
+  const previewInstallation = state.installation ?? reachable[0];
```
marians: The preview installation is gated on `useReachableInstallations`, but the avatar it drives is installation-independent by design (the doc comment says "DiceBear output is identical across installations"). `useReachableInstallations` only keeps `healthy` installations; on a cold, direct load, once the access probe seeds `connecting` entries the reachable set is *empty* until the first `healthy` result arrives (see the hook's own tradeoff note). During that window `reachable[0]` is `undefined`, so `buildAvatarUrl` returns `undefined` and the preview shows the initials fallback instead of the avatar — even though any configured installation with a `baseDomain` would produce the identical image. This couples a purely cosmetic preview to cluster reachability and to a `clusterAccessStatusApi` subscription it does not need. Consider falling back to the first configured installation that has a `baseDomain` instead of the first *reachable* one.

plugins/agent-platform/src/components/AgentsTable/AgentsTable.tsx
```diff
@@ -48,9 +98,11 @@ export type AgentsTableProps = {
  * "no agents" empty state.
  */
 export function AgentsTable({ rows }: AgentsTableProps) {
+  const buildAvatarUrl = useAgentAvatarUrl();
+
   return (
     <Table<AgentRow>
-      columnConfig={columnConfig}
+      columnConfig={getColumnConfig(buildAvatarUrl)}
```
marians: `getColumnConfig(buildAvatarUrl)` is called in the render body, so the entire column-config array and all five `cell` render closures are rebuilt on every `AgentsTable` render (`buildAvatarUrl` itself is stable via `useCallback`). Passing a fresh `columnConfig` identity to `<Table>` on each render defeats any memoization the bui Table does on its columns. Wrap it in `const columnConfig = useMemo(() => getColumnConfig(buildAvatarUrl), [buildAvatarUrl])`.

plugins/agent-platform/src/components/AgentsTable/AgentsTable.tsx
```diff
@@ -1,41 +1,91 @@
...
+              purpose="decoration"
+              name={row.name}
+              src={
+                buildAvatarUrl(row.installation, row.technicalName, {
+                  size: LIST_AVATAR_SIZE,
+                }) ?? ''
+              }
+            />
+            <Flex direction="column" gap="1">
+              <Text as="p" variant="body-medium">
```
marians: Replacing `CellText` with a hand-rolled `Flex`/`Text` layout for the name column drops the overflow/truncation handling `CellText` provides (the other columns still use `CellText`). A long agent name or description will now wrap/overflow differently from the rest of the table. If the larger avatar is the only reason `CellText` could not be used, consider whether `CellText`'s title/description text handling can be preserved alongside the custom avatar rather than re-implementing the text rendering.

app-config.yaml
```diff
@@ -80,6 +80,15 @@ backend:
...
+    # Agent avatar icons are served per installation from
+    # `avatars.<baseDomain>` (see plugins/agent-platform). Each deployment must
+    # allowlist its own installations' avatars hosts here — e.g.
+    # `https://avatars.<baseDomain>` — in the environment config where
+    # `gs.installations` is set (customer base domains are intentionally kept
+    # out of this checked-in file). Backstage replaces (does not merge) config
+    # arrays, so an environment override of `img-src` must re-list these base
+    # entries. For local dev, add your test installation's avatars host to
+    # app-config.local.yaml.
     img-src:
```
marians, line 92: Because `img-src` here does not (and by design cannot) list any `avatars.<baseDomain>` host, the checked-in default blocks every avatar image in all deployments until each environment overrides `img-src` — and since Backstage *replaces* rather than merges these arrays, that override must also re-list `'self'`, `data:`, and `avatars.githubusercontent.com` or those break too. The failure mode is silent (avatars degrade to initials, no error), so it's easy to ship a deployment where the whole feature is invisibly off. Worth confirming the per-installation env configs are updated in lockstep with this PR so the feature isn't dark on merge.

PR #1973, opened 2026-07-21
feat(agent-platform): set agent icon URL from technical name at creation

marians:
### What does this PR do?

Populates `agent.iconUrl` in the composed agent manifest at creation time, so the created Agent's `spec.iconUrl` (surfaced on the A2A AgentCard) matches the avatar Backstage already renders.

The review/deploy page now sets `agent.iconUrl` in the HelmRelease values to the **size-agnostic** canonical avatar URL (`https://avatars.<baseDomain>/v1/<name>.png`), derived from the agent's technical name via the same `useAgentAvatarUrl` builder used for display (added in #1971). The field is omitted when the installation has no configured base domain, so the chart keeps its default rather than persisting an empty URL.

Agents created through the Agent Platform now carry a stable avatar icon on the Agent resource / A2A AgentCard, consistent with the icon shown in the Backstage agent list and creation preview — and reusable by other consumers (e.g. Slack) that read `spec.iconUrl`.

Follows giantswarm/agent#8 (which added the `agent.iconUrl` chart value → `spec.iconUrl`) and giantswarm/giantswarm#37211 (comment 5035146327). This is the Backstage side of populating that field.

No.

### DIFF
--- a/plugins/agent-platform/src/components/NewAgentReviewPage/NewAgentReviewPage.tsx
+++ b/plugins/agent-platform/src/components/NewAgentReviewPage/NewAgentReviewPage.tsx
@@ -9,6 +9,7 @@ import { useProvidePageHeaderActions } from '@giantswarm/backstage-plugin-ui-rea
 
 import { CHART_DEFAULTS } from '../../lib/agentDefaults';
 import { CHART_NAME, composeManifests } from '../../lib/composeManifests';
+import { useAgentAvatarUrl } from '../../hooks/useAgentAvatarUrl';
 import { useAgentChart } from '../../hooks/useAgentChart';
 import { useDeployAgent } from '../../hooks/useDeployAgent';
 import { newAgentRouteRef } from '../../routes';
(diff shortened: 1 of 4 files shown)


plugins/agent-platform/src/components/NewAgentReviewPage/NewAgentReviewPage.tsx
```diff
@@ -127,6 +134,7 @@ export function NewAgentReviewPage() {
           description: state.description,
           modelConfigName: state.modelConfigName ?? '',
           systemMessage: state.systemMessage,
+          iconUrl: buildAvatarUrl(state.installation, state.slug) ?? '',
```
marians, line 137: **Low severity — persisted `spec.iconUrl` has no graceful fallback for external consumers.**

The Backstage UI (`useAgentAvatarUrl` → `Avatar`/`CellProfile`) degrades to initials when the avatars endpoint is unreachable. But the URL baked into the Agent resource here is unconditional whenever a `baseDomain` exists — so if the self-hosted DiceBear/avatars service is not actually serving on that installation, external consumers that read `spec.iconUrl` (A2A AgentCard, Slack) get a permanently broken image with no initials fallback, rather than the empty/default the chart would otherwise keep.

This is largely mitigated by design (`agentAvatar.ts` documents the avatars service as shipping with the agent-platform bundle), so likely an acceptable trade-off — flagging only so it is a conscious decision rather than a blocker.

PR #1984, opened 2026-07-23
refactor(flux): migrate Flux details panel to bui (@backstage/ui)

marians:
### What does this PR do?

Migrates the Flux **details panel** — the body rendered inside `DetailsPane` when a resource is selected in the tree/list view — and the full `ResourceCard` component tree from legacy `@material-ui/core` + `makeStyles` to bui (`@backstage/ui`). This is the first Flux → bui PR.

- Rebuilds the panel frame (`Section`, `Details`, and the Kustomization/HelmRelease/Repository/ImageAutomation detail components) on bui `Flex`/`Box`/`Text`.
- Migrates the `ResourceCard` tree (`ResourceCard`/`ResourceWrapper`, `ResourceInfo`, `ResourceHeading`, `ResourceStatus`, `ResourceChips`, `ResourceMetadata`, `CopyCommandMenu`). `CopyCommandMenu` now uses the bui `MenuTrigger`/`Menu` with a `ButtonIcon` trigger.
- Resource cards are now collapsible bui `Accordion`s (expanded by default): the name/kind/status row is the trigger, metadata/actions are the revealable panel. The card keeps its rounded corners and status-colored background (including the not-ready tint), the resource name is a bold heading, the status and type/namespace indicators are bui `Badge`s, and the caret is aligned top-right.
- The card's custom theme-aware background colors and the per-kind resource-type color coding are intentionally kept in colocated `makeStyles`, as bui has no equivalent tokens.

Visual refresh of the Flux details panel; no behavioural change. Because `ResourceWrapper`/`ResourceInfo` are shared with the tree-view node, panel-only styling is gated behind an `emphasized` flag so the tree view keeps its original appearance (its full migration is a follow-up PR).

### How does it look like?

Verified live in the running app (light + dark) via DevTools. Collapsible cards with rounded corners, not-ready rose tint, bold heading, color-coded type badge, and neutral namespace/status badges.

First step of the incremental Flux → bui migration. `DetailsPane` shell (in `ui-react`) and the tree-view node are out of scope here.

No.


plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
```diff
@@ -176,65 +203,94 @@ export const ResourceCard = ({
...
-                ) : (
-                  <NotAvailable />
-                ),
-              }}
-              fixedKeyColumnWidth="120px"
-            />
+      <Accordion
+        id={`${kind}-${namespace ?? ''}-${name}`}
+        className={classes.accordion}
+        defaultExpanded
```
marians, line 260: The card is an uncontrolled `Accordion` (`defaultExpanded`, no `isExpanded`), and `defaultExpanded` only applies on mount. `DetailsPane`/`FluxResourceDetails`/`Details` render this card in place without a `key` tied to the resource identity, so navigating between two resources **of the same kind** (e.g. Kustomization → Kustomization) keeps the same `Accordion` instance mounted. If the user collapsed the previous card, the new resources card stays collapsed instead of expanding by default. Consider keying the card (or `Details`/the detail component) by the resource ref so it re-mounts and re-applies `defaultExpanded`.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceHeading/ResourceHeading.tsx
```diff
@@ -1,50 +1,37 @@
...
   name,
   inactive,
   nowrap = false,
+  emphasized = false,
 }: ResourceHeadingProps) => {
-  const classes = useStyles();
+  const color = inactive ? 'secondary' : 'primary';
+  // The details panel (emphasized) gets a larger, bold heading; the tree-view
+  // node keeps its original, smaller size until it is migrated separately.
+  const as = emphasized ? 'h4' : 'div';
```
marians: In emphasized mode `ResourceHeading` renders `<Text as="h4">`, and this heading is placed inside `AccordionTrigger`, which (react-aria `Disclosure`) wraps its children in a `<button>` nested in a `<Heading>`. That yields `<h3><button>…<h4>name</h4>…</button></h3>` — a heading nested in a button nested in a heading. This is invalid HTML (a `<button>` may only contain phrasing content, and headings should not nest) and can confuse screen readers. Browsers tolerate it, but consider `as="div"` (or a non-heading element) for the emphasized title inside the trigger.

plugins/flux-react/src/components/FluxOverview/ResourceCard/CopyCommandMenu/CopyCommandMenu.tsx
```diff
@@ -164,56 +134,41 @@ export const CopyCommandMenu = ({ resource }: CopyCommandMenuProps) => {
...
+      <MenuTrigger>
+        <ButtonIcon
+          icon={
+            copiedId ? (
+              <CheckIcon fontSize="small" />
+            ) : (
+              <TerminalIcon fontSize="small" />
+            )
+          }
           aria-label="Copy CLI command"
```
marians: The copy button lost its tooltip in the migration. The old code wrapped the button in a MUI `<Tooltip title="Copy CLI command">`; now there is only `aria-label`, which is not shown on hover. `ButtonIcon` does not render a visible tooltip from `aria-label` on its own, so sighted users lose the affordance label. Consider wrapping in bui `TooltipTrigger`/`Tooltip`.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
```diff
@@ -176,65 +203,94 @@ export const ResourceCard = ({
...
-                    relative
-                  />
-                ) : (
-                  <NotAvailable />
-                ),
-              }}
-              fixedKeyColumnWidth="120px"
-            />
+      <Accordion
+        id={`${kind}-${namespace ?? ''}-${name}`}
```
marians: The `Accordion` `id` is derived from `kind`/`namespace`/`name` only (not `cluster`). Each `ResourceCard` is a standalone `Disclosure` whose `id` drives the panel `id` and the trigger `aria-controls`. If the same kind+namespace+name identity appears in two cards on one panel (e.g. the primary "This X" card and a related/dependency card referring to the same object), the DOM `id`s collide — invalid HTML and mis-wired `aria-controls`. Including `cluster` (already part of the React `key` used for dependency cards) would make it unique.

plugins/flux-react/src/components/FluxOverview/ResourceCard/CopyCommandMenu/CopyCommandMenu.tsx
marians: Fixed in 31d86547 — wrapped the `ButtonIcon` in bui `TooltipTrigger`/`Tooltip`. Verified live: hovering shows the "Copy CLI command" tooltip and the button now has a matching `aria-describedby`, and opening the menu still works with `TooltipTrigger` nested inside `MenuTrigger`.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceHeading/ResourceHeading.tsx
marians: Fixed in 31d86547 — the emphasized title now renders as a `div` (keeping the bold weight for the visual heading look) instead of `h4`, so there is no longer a heading nested inside the trigger `<button>`. Verified in the a11y tree that the AccordionTrigger button contains no heading elements.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
marians: Fixed in 31d86547 — the `Accordion` is now keyed by the resource identity (`kind-namespace-name-cluster`), so it re-mounts and re-applies `defaultExpanded` when the panel switches to a different resource of the same kind. Verified live: collapsed a HelmRelease card, navigated to another HelmRelease, and the new card came up expanded again.

plugins/flux-react/src/components/FluxOverview/ResourceCard/ResourceCard.tsx
marians: Fixed in 31d86547 — the `Accordion` `id` now includes `cluster` (`kind-namespace-name-cluster`), matching the React key already used for dependency cards, so the ids stay unique even when the same kind+namespace+name identity appears in more than one card on a panel.

PR #1988, opened 2026-07-23
refactor(flux): migrate Flux Tree view to bui (@backstage/ui)

marians:
### What does this PR do?

Continues the migration of the Flux UI off Material-UI onto bui (`@backstage/ui`), following the details-panel migration (#1984). This slice covers the **Tree view**:

- **FluxOverview** — MUI `Box` → bui `Flex`.
- **ResourceNode** — MUI `Box`/`Button`/`makeStyles`/`PlayArrowIcon` → bui `Flex`/`Box`/`ButtonIcon`.
- **TreeSearch** — MUI `TextField`/`IconButton`/`Typography` → bui `SearchField`/`Text`/`ButtonIcon` (`SearchField`'s built-in clear button replaces the manual one).
- **ResourceCard** — add a `style` passthrough to `ResourceWrapper` so the tree node sets its width directly (an extra wrapper would break the `a:hover > &` hover rule).
- Delete the dead `UI/Tree` component (unexported, unimported; the live tree is `react-vtree` in `OverviewTree`).

Intentionally left on `makeStyles`: `ContentContainer` (scroll-fade pseudo-elements + breakpoints, no bui token). Out of scope: the shared `SingleSelect`/`FiltersLayout`/`DetailsPane` (used across gs/muster/flux).

No functional change — the Tree view now renders with bui components. One latent bug in the expand/collapse caret is fixed along the way (see below).

### How does it look like?

Unchanged visually. Verified live on `/flux/tree`: tree renders, expand/collapse caret toggles in place, resource selection opens the details pane, and search (typing, clear, counter, prev/next, Ctrl/Cmd+F / +G) works.

Notable fix: the expand caret is wrapped in an `onClickCapture` `preventDefault` handler. The tree row is a react-router `Link` with a real `href`, and bui `ButtonIcon` (react-aria) calls `stopPropagation()` on the bubbling click — so a bubble-phase `onClick` never fires and the native anchor navigation would **reload the page**. Capture runs before react-aria severs the bubble; `preventDefault` (without `stopPropagation`, which would block press activation) cancels the navigation while still toggling the node.

Follow-up to #1984. Adds render/interaction tests for `FluxOverview`, `ResourceNode` (incl. a regression test for the navigation guard) and `TreeSearch` — no tests existed for these before.

No.


marians:
## Code review (automated, medium depth)

Reviewed the diff for the Flux **Tree view** bui migration. Overall this is a clean, well-scoped continuation of #1984 — the component swaps are faithful, the MUI→bui spacing tokens are mapped consistently (`p="4"`≈16px, `pt="2"`≈8px, `pl="12"`=48px), the dead `UI/Tree` component is genuinely unreferenced, and the new tests give these components their first coverage. No correctness issues found.

The stand-out is the navigation-guard handling in `ResourceNode`: the `onClickCapture` + `preventDefault` (deliberately *without* `stopPropagation`) is correct and the inline comment explains exactly why react-aria's bubble-phase `stopPropagation` forces the capture phase. Good catch on the latent full-page-reload bug, and nice that it has a regression test.

### Minor observations (non-blocking)

- **Nav buttons lost their hover tooltips.** The old `IconButton`s used `title="Previous match"` / `title="Next match"`, which render a native tooltip on hover; the new `ButtonIcon`s carry only `aria-label`, so sighted mouse users no longer get the tooltip. Accessible name is preserved, so this is purely a small UX nicety — flagging in case it was unintentional.

- **`searchFieldRef` reaches through to the input via `querySelector('input')`.** This depends on bui `SearchField` forwarding its ref to a wrapper element rather than the `<input>` itself. It's verified working live, but it's the most upgrade-fragile spot in the diff. Note that `muster`'s `ToolBrowser` takes the more robust route of putting the ref on a wrapping `<Box ref=...>` instead — worth aligning on that pattern for consistency.

- **Keyboard shortcuts (Ctrl/Cmd+F / +G) aren't covered by tests.** The focus/navigation `keydown` handler is the one behaviour in `TreeSearch` without a test; given it hangs on the ref-forwarding detail above, a small test asserting Ctrl+F focuses the input would protect it against a future bui change.

Nothing here blocks merge. 👍

PR #1990, opened 2026-07-24
fix(gs): align and shrink code-block copy button

marians:
### What does this PR do?

Extracts the "code block + copy button" pattern into a reusable `CodeBlock`
component in the shared `ui-react` library and migrates the `gs` plugin to it.

The copy button is now:
- **aligned** to the top-right of its code field (removing the previous
  negative-margin workaround that pushed it out of place),
- **neutral** — a bui `ButtonIcon` (`tertiary` variant) instead of the
  oversized, primary-blue `CopyTextButton`,
- **text-sized** — a 16px icon in a 24px button that fits inside a single-line
  code block, with a "Copy"/"Copied" tooltip and check-icon feedback.

`gs` re-exports `CodeBlock` from `ui-react` (same pattern already used for
`YamlEditorFormField`), so all consumers pick up the fix with no call-site
changes.

The copy button in the "Kubernetes API access" and "SSH access" boxes on the
cluster details page (and other code blocks) is now correctly aligned, smaller,
and neutral-colored instead of a displaced, oversized blue button.

### How does it look like?

image

image

Backstage's `CopyTextButton` (`@backstage/core-components`) now renders a bui
`ButtonIcon` in the default primary variant at a fixed 32px footprint; the old
`gs` `CodeBlock` tried to tame it with negative margins, which produced the
misalignment.

No.


plugins/ui-react/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,109 @@
...
+  root: {
+    position: 'relative',
+    width: '100%',
+  },
+  codeBlock: {
+    margin: 0,
+    // Leave room on the right so long single-line content doesn't run under
+    // the copy button.
+    padding: theme.spacing(1),
+    paddingRight: theme.spacing(5),
```
marians: The `paddingRight: theme.spacing(5)` reserving space for the copy button is applied unconditionally, but the button is only rendered when `copyEnabled` is true. For a `copyEnabled={false}` block that is **not** transparent, this leaves ~40px of empty dead space on the right with no button. (The current `copyEnabled={false}` callers in `ReviewStep` happen to also pass `transparent`, so the background hides it there — but the reserved padding is still unconditional.) Consider gating the extra padding on `copyEnabled`.

plugins/ui-react/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,109 @@
...
+    top: theme.spacing(0.5),
+    right: theme.spacing(0.5),
+    // bui's small ButtonIcon is 32px square — oversized next to a text-sized
+    // icon, and taller than a single-line code block (so its hover background
+    // pokes past the bottom edge). Shrink the whole button so it sits tidily
+    // inside the block; `!important` overrides bui's own height/width rules.
+    width: '1.5rem !important',
+    height: '1.5rem !important',
+    minWidth: 'unset',
+    minHeight: 'unset',
```
marians: `minWidth`/`minHeight: 'unset'` here lack `!important`, yet the comment just above (and the new `ui` skill doc in this same PR) states that bui's `data-size` selectors out-specify a plain `makeStyles` class and need `!important` (or `&&`) to take effect. So these two lines are almost certainly no-ops. They are also moot given `width`/`height` are pinned with `!important`, so they can just be removed.

plugins/ui-react/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const handleCopy = () => {
+    copyToClipboard(text);
+    setCopied(true);
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+    }
+    timeoutRef.current = setTimeout(() => setCopied(false), 1500);
+  };
+
+  const showCopied = copied && !error;
```
marians, line 94: On a copy failure `error` is captured and used to suppress the "Copied" state, but nothing is ever surfaced to the user or logged — the button silently reverts to "Copy" with no indication the copy failed. This also drops the success feedback the old `CopyTextButton` gave (a snackbar). If that reduced feedback is intentional, fine; otherwise consider at least reporting the error case.

plugins/ui-react/src/components/CodeBlock/CodeBlock.tsx
marians: Fixed in 4a409e5d — moved the extra right padding into a `withCopyButton` class that's only applied when `copyEnabled` is true, so a non-transparent block without a copy button no longer has dead space on the right.

plugins/ui-react/src/components/CodeBlock/CodeBlock.tsx
marians: Fixed in 4a409e5d — removed both lines. Correct: they lacked `!important` so were no-ops, and are moot given `width`/`height` are pinned.

plugins/ui-react/src/components/CodeBlock/CodeBlock.tsx
marians: Addressed in 4a409e5d — the copy failure is now posted to `errorApi` (via `useApi(errorApiRef)`), matching Backstage's `CopyTextButton`. The success feedback is intentionally the inline "Copied" tooltip + check icon rather than a snackbar, which reads better inline for a code block.

PR #1992, opened 2026-07-24
docs(ui-react): add Storybook site for shared UI components

marians:
### What does this PR do?

Stands up a **Storybook site** documenting Giant Swarm's shared UI component
library, `@giantswarm/backstage-plugin-ui-react`, and makes the in-repo stories
the canonical usage reference. Implements the PRD for epic #37264.

- **Config:** a single root-level `.storybook/` (Storybook 10 + React/Vite) that
  glob-discovers `plugins/ui-react/**/*.stories.tsx` — adding a component with a
  story needs no config change.
- **Stories for all 22 barrel-exported visual components**, each with Storybook
  autodocs (auto-generated props table + controls), a default story, the key
  variants/states, a short *what / when to use it* blurb, and a
  **migration-status note** (deprecated MUI v4 vs bui).
- **Themes:** a global decorator supplies the *real* GS light/dark themes
  (reusing the app's `buildPalette`), switchable from a toolbar control — a single
  `UnifiedThemeProvider` themes MUI v4, MUI v5, and bui at once.
- **Intro MDX** landing page (the three UI layers; when to reach for `ui-react`
  vs bui vs building new; an agent pointer).
- **Coverage gate:** a pure, unit-tested function
  (`plugins/ui-react/src/storybook/storyCoverage.ts`) + a thin CLI wrapper
  (`yarn storybook:coverage`) that fails when an exported component has no story,
  and surfaces stale allowlist entries. Allowlist:
  `.storybook/story-coverage-allowlist.json` (currently empty — 22/22 documented).
- **CI:** `.github/workflows/storybook.yaml` runs the coverage gate + a full
  Storybook build on every PR (a broken story fails the pipeline) and publishes
  the static site to **GitHub Pages** on merge to `main`.
- **Discoverability:** rewrote `plugins/ui-react/README.md` (agent pointer +
  hooks/utils inventory) and extended the `.claude/skills/ui` skill.

Source edits to the components are limited to what autodocs needs (no runtime or
behavioural changes); the published package output (`dist`) is unchanged.

No end-user-facing change. This is a developer/agent-facing documentation and
tooling deliverable: a browsable, themed, always-up-to-date reference for the
shared UI library, published at a stable GitHub Pages URL.

### How does it look like?

`yarn storybook` serves it locally on http://localhost:6006. Each component page
has a live rendering, interactive controls, an auto-generated props table, its
key states, and a migration note; the **Theme** toolbar control flips between the
real GS light and dark themes.

- Epic: #37264
- PRD: `backstage-component-docs/PRD.md` in `giantswarm/bumblebee-plans`
- Follow-up filed while documenting `ContentRow`: #37265 (two divergent
  `ContentRow` implementations — the `ui-react` one is unused and renders its
  title/value flush; consolidation was out of scope here).

This PR *is* the docs. The `ui-react` README and the `ui` skill are updated in the
same PR.

No changeset: the change is docs + tooling + dev-dependencies only; the published
`@giantswarm/backstage-plugin-ui-react` output is unchanged, so there is nothing
for consumers to release.

Ref: https://github.com/giantswarm/giantswarm/issues/37264


.storybook/decorators.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { ConfigReader } from '@backstage/config';
+import { configApiRef, errorApiRef } from '@backstage/core-plugin-api';
+import { TestApiProvider } from '@backstage/test-utils';
+import { createUnifiedTheme, UnifiedThemeProvider } from '@backstage/theme';
+import type { Decorator } from '@storybook/react';
+// Reuse the *real* app palette-building logic so stories render in the exact
+// theme users see. `buildPalette` is a pure function; with an empty config it
+// returns the built-in GS light/dark palettes (no branding overrides), matching
+// a default deployment.
+import { buildPalette } from '../packages/app/src/modules/app/customThemes';
```
marians, line 12: The Storybook config lives outside `tsconfig.json`'s `include` (`packages/*/src`, `plugins/*/src` only), so `yarn tsc`/`ci:verify` never type-checks `.storybook/*` or `scripts/*`. Type errors here (this deep app-internal import, the `as any` on the errorApi stub, the `context.globals` access in `preview.tsx`, and everything in `scripts/check-story-coverage.mts`) are only surfaced by a full `storybook:build`, and the coverage script is run via tsx (type-stripped, no check) so its type errors are never caught at all. Consider adding these paths to a tsconfig include or a dedicated typecheck step.

scripts/check-story-coverage.mts
```diff
@@ -0,0 +1,110 @@
...
+
+/** All `*.stories.tsx` files under the plugin, as component names. */
+function readStoriedComponents(): string[] {
+  const names = new Set<string>();
+  const walk = (dir: string) => {
+    for (const entry of readdirSync(dir, { withFileTypes: true })) {
+      const full = join(dir, entry.name);
+      if (entry.isDirectory()) {
+        walk(full);
+      } else if (entry.name.endsWith('.stories.tsx')) {
```
marians: Coverage gate only counts `*.stories.tsx`, but `.storybook/main.ts` discovers stories via `*.stories.@(ts|tsx)`. A story authored as `.stories.ts` would render in Storybook yet the gate would still report its component as undocumented and fail CI. Match the two globs (accept `.stories.ts` too).

scripts/check-story-coverage.mts
```diff
@@ -0,0 +1,110 @@
...
+ * Component names barrel-exported by the library: every `export * from './X/Y'`
+ * in components/index.ts contributes its last path segment (the component name).
+ */
+function readExportedComponents(): string[] {
+  const source = readFileSync(componentsIndex, 'utf8');
+  const names = new Set<string>();
+  const re = /export\s+\*\s+from\s+['"]\.\/(.+?)['"]/g;
+  for (const match of source.matchAll(re)) {
+    const segments = match[1].split('/');
+    names.add(segments[segments.length - 1]);
```
marians, line 45: Exported-component names are derived from the barrel *import path segment* (`export * from './X/Y'` -> `Y`), not the actual exported symbol. It works today only because every directory name equals its component name. If a directory ever exports a differently-named component (or more than one), the gate would demand a story file named after the directory and a correctly-named story wouldn't satisfy it — a confusing false failure.

plugins/ui-react/src/storybook/storyCoverage.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function checkStoryCoverage(
+  input: StoryCoverageInput,
+): StoryCoverageResult {
+  const exported = new Set(input.exportedComponents);
+  const storied = new Set(input.storiedComponents);
+  const allowlisted = new Set(input.allowlist);
+
+  const undocumented = [...exported]
```
marians, line 54: The gate flags undocumented exports and stale allowlist entries, but silently ignores a `*.stories.tsx` whose component is no longer exported (a dangling story after a rename/removal) and an allowlist entry that is redundantly also storied. Neither is caught, so the story set can drift out of sync with the barrel without failing CI.

.github/workflows/storybook.yaml
```diff
@@ -0,0 +1,63 @@
+# Builds the ui-react Storybook (docs for @giantswarm/backstage-plugin-ui-react)
+# and enforces the story coverage gate on every PR; publishes the static site to
+# GitHub Pages on merge to main.
+#
+# Hand-authored (not devctl-generated), so it is safe to edit here.
+name: Storybook
+
+on:
+  pull_request:
```
marians, line 13: No path filter on the `pull_request` trigger: every PR (including backend-only or docs-only changes) now pays a full `yarn install --immutable` + `storybook:build`. Consider `paths:` scoping to `plugins/ui-react/**`, `.storybook/**`, `scripts/check-story-coverage.mts`, and this workflow.

plugins/ui-react/.eslintrc.js
```diff
@@ -1 +1,19 @@
...
+    ...(config.overrides ?? []),
+    {
+      // Storybook stories legitimately call hooks (useState, etc.) inside a
+      // `render` callback to make a component interactive. The render fn isn't
+      // recognised as a component by the rules-of-hooks lint, so disable it for
+      // story files only — the hooks are still called unconditionally at the
+      // top of each render.
+      files: ['src/**/*.stories.tsx'],
+      rules: {
+        'react-hooks/rules-of-hooks': 'off',
```
marians: Disabling `react-hooks/rules-of-hooks` for *all* story files also masks genuine mistakes (a hook called conditionally / inside a nested callback in a story). The comment asserts hooks are always called unconditionally at the top of `render`, but nothing now enforces that. A narrower alternative is to keep the rule and structure the interactive stories as small components.

.storybook/decorators.tsx
marians: Fixed in 590bdc6a. Added `tsconfig.storybook.json` (includes `.storybook/**` and `scripts/**`, `noEmit`) plus a `yarn typecheck:storybook` script, wired into this workflow as a step before the build. It immediately paid off — it flagged the untyped bui `*.css` side-effect import (now covered by a scoped ambient declaration) and the `errorApi` stub, which is now `new MockErrorApi()` instead of an `as any` cast.

scripts/check-story-coverage.mts
marians: Fixed in 590bdc6a. The gate now discovers stories with `/\.stories\.tsx?$/`, matching the `*.stories.@(ts|tsx)` glob in `.storybook/main.ts`, so a `.stories.ts` file counts too.

scripts/check-story-coverage.mts
marians: Good catch on the implicit assumption. Documented it in 590bdc6a: the derivation relies on the repo's one-component-per-directory convention (directory name == component name), and I noted that a differently-named export would require resolving real symbols instead. The new `danglingStories` check (see the storyCoverage thread) also guards the inverse — a story with no matching export now fails the gate rather than passing silently.

plugins/ui-react/src/storybook/storyCoverage.ts
marians: Fixed in 590bdc6a. `checkStoryCoverage` now also returns `danglingStories` (a `*.stories.tsx` whose component is no longer exported) and `redundantAllowlist` (an allowlist entry whose component actually has a story), and the gate fails on both. Added unit tests for each case.

.github/workflows/storybook.yaml
marians: Fixed in 590bdc6a. Scoped the `pull_request` and `push` triggers with `paths:` — `plugins/ui-react/**`, `.storybook/**`, `scripts/check-story-coverage.mts`, `tsconfig.storybook.json`, this workflow, and `package.json`/`yarn.lock` (so dependency bumps still trigger it). Left `merge_group` unscoped so the merge queue always gets a signal.

plugins/ui-react/.eslintrc.js
marians: Agreed — went with the narrower alternative in 590bdc6a. The four interactive stories (SingleSelect, MultipleSelect, Autocomplete, YamlEditorFormField) now put their hooks in small named example components, and the blanket `react-hooks/rules-of-hooks` override is removed, so the rule is back to enforcing across all story files.

PR #1997, opened 2026-07-27
feat(agent-platform-backend): add kagent REST proxy

marians:
### What does this PR do?

Adds a new `agent-platform-backend` plugin: a thin REST proxy over the kagent
controller API, one endpoint per installation. This is the transport the
upcoming Agent Platform **Sessions** list needs, split out so it can be reviewed
and CI'd on its own — nothing calls it yet.

kagent sessions live in kagent's Postgres and are served over the controller's
HTTP API. Unlike agents and model configs they are **not** Kubernetes resources,
so the Kubernetes proxy the rest of the plugin uses (`useResources` +
`KubeObject`) cannot reach them at all.

A backend hop is then unavoidable for two reasons:

- **CORS/CSP** — the browser cannot call `kagent.<baseDomain>` directly.
- **Header collision** — kagent runs `controller.auth.mode: trusted-proxy` and
  derives the user from the `sub` claim of an `Authorization: Bearer` JWT. On the
  inbound leg that header already carries the Backstage identity, so the user's
  per-installation Dex ID token is forwarded in a separate
  `backstage-kagent-authorization` header and promoted to `Authorization` here.
  Same approach as `muster-backend`.

Some deliberate design choices worth a reviewer's attention:

- **Verbatim proxy.** The `{error, data, message}` envelope is not unwrapped,
  unknown fields are not stripped, nothing is filtered. Schema tolerance lives in
  the frontend client (next PR), so kagent schema drift never needs a backend
  release. The split is **backend = transport, frontend = schema**.
- **Derived URLs.** `https://kagent.<baseDomain>/api` per installation — the
  oauth2-proxy-fronted host whose nginx sidecar proxies `/api/` to
  `kagent-controller:8083`. Overridable per installation, which doubles as an
  allowlist since kagent is only deployed on some installations.
- **`/version` resolves against the origin, not the base URL.** kagent serves it
  at the server root (`APIPathVersion`), so appending it to `…/api` would 404.
- **`redirect: 'manual'`.** An oauth2-proxy redirect into Dex surfaces as a 401
  instead of being followed into a 200 HTML sign-in page.

None yet — no UI consumes these routes. It unblocks the Sessions list.

### How does it look like?

All routes are under `/api/agent-platform` and require `?installation=<name>`:

| Route | Token | Purpose |
| --- | --- | --- |
| `GET /health` | — | `{ status, configured }` |
| `GET /kagent/installations` | — | Installation names kagent can be proxied for |
| `GET /kagent/sessions` | required | The user's sessions, verbatim |
| `GET /kagent/version` | optional | Version probe for capability negotiation |
| `GET /kagent/me` | optional | Identity probe (diagnoses empty/shared lists) |

> **Correction (2026-07-27, after merge).** This section originally claimed the
> transport was verified end to end against a live installation with a
> Backstage-minted Dex token. That was wrong: the curl behind that claim used a
> token minted for a *different* client, so it validated some token rather than the
> one this code sends — and those differ in exactly the field that matters.
>
> Verified in the browser since: session reads return **401** on every kagent
> installation. kagent's oauth2-proxy runs `skip-jwt-bearer-tokens: true`, which
> accepts a bearer only when its audience matches oauth2-proxy's own `--client-id`,
> and no `--oidc-extra-audience` is set in the charts or any install config.
> Backstage's token carries `aud: [dex-k8s-authenticator, <backstage client>]`
> while kagent's oauth2-proxy uses a separate Dex client.
>
> The proxy code itself is unaffected — it forwards the token and maps the upstream
> 401 correctly. The fix is platform-side (`--oidc-extra-audience` on kagent's
> oauth2-proxy) and is being handled separately. The prerequisite is documented in
> `docs/agent-platform.md`. A second issue is queued behind it: kagent scopes on
> `sub` while stored sessions have been seen keyed by email, so the list may be
> empty even once the audience is accepted — `GET /kagent/me` distinguishes them.

[agent-platform-ui prototype](https://github.com/giantswarm/agent-platform-ui).
The prototype is fully mock-driven, and its `Session` has ~20 fields against
kagent's 7 — so the eventual UI is a deliberately reduced, read-only version.

Two notes for reviewers:

- **Config visibility.** Every `agentPlatform.kagent` key keeps the default
  *backend* visibility; `apiBaseUrl` embeds `baseDomain`, which deanonymizes
  customers (the same reason `gs.installations` is backend-only). Confirmed with
  `config:print --frontend`, and `/kagent/installations` returns names only, with
  a test asserting no URL key leaks. Worth knowing:
  `backstage-cli config:check` does **not** catch schema violations — it passes
  both a wrong-typed value and an unknown key — so it is not a guard here.
- **Unauthenticated agentgateway door.** Independent of this PR: on GS
  installations `https://agentgateway.<baseDomain>/kagent/**` exposes kagent's
  whole REST + A2A surface with no auth enforced. The chart comment claims
  agentgateway validates the JWT there, but that `AgentgatewayPolicy` only
  renders under `oauthMode: validate` and only targets `/mcp`, while GS runs
  `passthrough`. This PR deliberately uses the oauth2-proxy door instead.

The plugin README is included here. The `docs/agent-platform.md` Sessions section
lands with the UI PR, where it is user-visible.


plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * `GET <origin>/version`.
+   *
+   * NOTE: kagent serves `/version` at the **server root**, not under `/api`
+   * (see `APIPathVersion` in kagent's `internal/httpserver/server.go`), so this
+   * resolves against the origin rather than appending to the configured base
+   * URL — otherwise it 404s.
+   */
+  async getVersion(options: KagentRequestOptions): Promise<unknown> {
+    const url = new URL('/version', this.installation.apiBaseUrl).toString();
```
marians: **`/version` never reaches the kagent controller through either supported door**

`new URL('/version', apiBaseUrl)` keeps only the *origin* and drops the base URL's path, and neither ingress routes the server root to the controller:

- **Derived door** (`https://kagent.<baseDomain>/api`): kagent's UI nginx sidecar (`helm/kagent/files/nginx.conf`) proxies only `location /api/` to `kagent-controller:8083` — `location /` goes to the UI on `127.0.0.1:8001`. So `GET https://kagent.<baseDomain>/version` is answered by the UI, which returns HTML. The non-JSON guard below then throws `AuthenticationError`, i.e. the version probe reports 401 *"a sign-in page was served instead"* on a perfectly healthy installation.
- **Documented override** (`https://agentgateway.<baseDomain>/kagent/api`, see `config.d.ts` and the README): the agentgateway HTTPRoute matches on the `/kagent` path prefix (`agentic-platform`, `templates/kagent/controller-route.yaml`). `https://agentgateway.<baseDomain>/version` doesn't match it and falls through to the catch-all MCP route, which answers 406 → `ServiceUnavailableError` (503).

The unit test only asserts the composed URL string, so it can't catch either case. `APIPathVersion = "/version"` is correct *in-process*, but it isn't reachable at the origin through the doors this plugin targets — consider probing something under `/api` (e.g. `/api/me`) for capability negotiation instead, or dropping the route.

Secondary issue on the same line: `new URL()` throws `TypeError: Invalid URL` if a configured `apiBaseUrl` is not absolute (operator omits `https://`), and that throw is outside `request()`'s try/catch, so it surfaces as an unmapped 500 rather than the 503 every other call path produces.

plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -0,0 +1,248 @@
...
+    // A 2xx with a non-JSON body is oauth2-proxy serving its sign-in page
+    // rather than kagent answering.
+    const contentType = response.headers.get('content-type') ?? '';
+    if (!contentType.includes('application/json')) {
+      throw new AuthenticationError(
+        `The kagent API for installation '${this.installation.name}' returned a non-JSON response (content-type '${contentType}'), which usually means a sign-in page was served instead.`,
+      );
+    }
+
+    return response.json();
```
marians: **Body-read failures escape the error mapping and become a 500**

The `try`/`catch` above only covers the `fetch()` call; the content-type check and `response.json()` run outside it. `AbortSignal.timeout(this.timeoutMs)` stays armed after the headers arrive, so a kagent that answers fast but streams a large session list past the 10 s deadline makes `response.json()` reject with an `AbortError`. Same for a mid-body connection reset, or a truncated/invalid JSON body (`SyntaxError`).

None of those are mapped: they bubble out of `request()` untyped, so `/kagent/sessions` returns a generic 500 instead of the 503 the frontend is meant to key off for "kagent unreachable". Wrapping the body read in the same `catch` (mapping to `ServiceUnavailableError`) keeps the documented contract.

plugins/agent-platform-backend/src/KagentClient.ts
marians: Confirmed on both counts, and fixed in 76bc6fd by **removing the version probe entirely**.

I verified each claim against the sources rather than taking it on trust:

- `helm/kagent/files/nginx.conf` does have `location / { proxy_pass http://kagent_ui; }` and only `location /api/ { proxy_pass http://kagent_backend/api/; }`. So the derived door answers `/version` from the UI with HTML, and my own non-JSON guard then reports `AuthenticationError` on a healthy installation — exactly as described.
- The agentgateway HTTPRoute matches `PathPrefix: /kagent`, so a root-relative `/version` doesn't match it.

I also checked whether the probe could simply move under `/api`, since that was the suggested alternative. It can't: `APIPathVersion = "/version"` is the only version route in `server.go`, there is no `/api/version`, and the `Version` fields in `/api/substrate/status` are per-actor/worker (`actorEntryFromPB`, `workerEntryFromPB`), not the controller's. So there is nothing reachable to probe through either door.

Given that, keeping a best-effort probe would be dead weight plus a misleading error path, so it's gone rather than patched. What this costs: the version-derived capability flags (`hasSessionShares`, `canRenameSessionViaPatch`, `hasSessionReadOnly`) in #1998. All three were unused — the Sessions list consumes none of them — so nothing regresses today. Version *tolerance* is untouched: it lives in the permissive parsing layer, which is where it actually does the work; version *detection* was the speculative part. I'll strip those flags from #1998 and keep `isUserScoped`, which comes from `/api/me` and is genuinely reachable.

If we ever do need version gating, the better shape is probing by behaviour (call a version-specific endpoint, treat 404 as "absent") rather than by version string — noted in the code and README so this isn't rediscovered.

Two tests now guard it: one asserts every outgoing request stays under `/api`, one asserts the route 404s so it isn't casually reintroduced.

On the secondary point — also fixed. `apiBaseUrl` is now validated as an absolute http(s) URL in `readKagentInstallationsFromConfig`, so a scheme-less value is skipped at startup with one clear log line instead of failing opaquely per request. Covered by a table test (scheme-less, protocol-relative, bare path, non-http scheme, nonsense).

plugins/agent-platform-backend/src/KagentClient.ts
marians: Correct — fixed in 76bc6fd. The body read is now inside the same mapping, converting to `ServiceUnavailableError`:

```ts
try {
  return await response.json();
} catch (error) {
  this.logger.debug(...);
  throw new ServiceUnavailableError(
    `Could not read the response from the kagent API for installation '${this.installation.name}'.`,
  );
}
```

Your reasoning about the still-armed `AbortSignal.timeout` is the part I'd missed: I'd been thinking of the timeout as bounding connection setup, when it bounds the whole exchange. A large session list is a realistic trigger too, since `GET /api/sessions` has no server-side pagination and returns the user's entire history.

Three tests cover it — mid-stream `AbortError`, `SyntaxError` from truncated JSON, and a connection reset — each asserting `ServiceUnavailableError` rather than an unmapped throw.

Worth noting the fix immediately proved itself: a pre-existing test of mine reused one mocked `Response` across two calls, and since a body can only be read once, the second read now surfaced as a clean 503 instead of an untyped failure. That was a flaw in the test, which I corrected, but it was a nice demonstration that the path is actually reachable.

PR #2000, opened 2026-07-27
feat(agent-platform): add Sessions tab

marians:
### What does this PR do?

Adds the **Sessions** tab at `/agent-platform/sessions` — a read-only list of the
signed-in user's kagent chat sessions across the fleet, built on the client from
#1998.

> **Stacked on #1998.** Base is `agent-platform-kagent-client`; it will retarget to
> `main` automatically when that merges.

Columns: Session, Agent (display name + avatar), Installation, Started, Last
activity. Client-side search across title/agent/installation, sortable columns,
pagination.

Three decisions worth a reviewer's attention:

**The session queries wait for the kagent allowlist.** My first cut fanned out to
every reachable installation immediately and narrowed once
`GET /kagent/installations` resolved. A test caught the cost: kagent runs on only a
couple of installations, so that fires a doomed request for every *other* one — and
each mints that installation's Dex token (potentially a broker exchange) before it
can fail. The allowlist is a single cheap backend call, cached for an hour and
persisted, so paying one round-trip up front beats N wasted ones on every cold
load. If the allowlist itself fails we fall back to the reachable set, so a backend
hiccup doesn't look like "you have no sessions".

**The agent join matches on the encode side.** kagent's `agent_id` is a "python
identifier" (`ns/name` → `ns__NS__name`, every `-` → `_`), so decoding is lossy —
`sre_agent` and `sre-agent` are indistinguishable. Instead of decoding, the already
loaded `Agent` CRs are indexed by their *encoded* id, scoped per installation. A
decoded label is only a display fallback, and a genuine encode collision resolves
deterministically (asserted by reordering the inputs rather than by pinning which
one wins, since that's down to locale collation).

**The not-user-scoped warning fires only on an explicit negative.** kagent's
`unsecure` mode resolves every caller to a shared built-in user, which would
silently present a shared list as the user's own. But `isUserScoped` is tri-state:
`undefined` means the probe hasn't resolved *or* kagent reported no subject — which
is reachable on a healthy deployment, since `/api/me` returns claims verbatim and
an IdP need not emit `sub`. Warning there would flag a working installation, so the
UI stays silent unless the answer is definitively "no".

A new "Sessions" tab in the Agent Platform section, listing your own kagent
sessions across all management clusters that run kagent. Installations without
kagent are silently skipped; ones that genuinely fail to read are reported.

### How does it look like?

Fleet-merged table sorted by last activity, each row tagged with its installation.
Timestamps render as relative dates (`ui-react`'s `DateComponent`), with a dash
where kagent reports none.

Two things about the data that shape the design, both from the live payload:

- **Titles are short and lossy.** kagent derives them from the first message and
  truncates to 20 characters — `"What issues are assi..."` — and the full text is
  unrecoverable. So the Session column is weak by nature and the Agent column
  carries much of a row's meaning. Worth knowing before judging the layout.
- **`source` is absent on every real row**, so the "exclude A2A subagent sessions"
  filter hides nothing today. Kept as forward-compatibility, not load-bearing.

[agent-platform-ui prototype](https://github.com/giantswarm/agent-platform-ui),
deliberately reduced. A kagent `Session` has 7 fields, so the prototype's status,
trigger, duration, cost, tokens, team, linked task, results and evaluation columns
— plus the nine-stat summary band derived from them — have **no backing data at
all**. kagent also lists with `WHERE user_id = <sub>` and exposes no cross-user
endpoint, so Mine/Watched/All collapses to one implicit scope. Session detail,
delete and rename are deferred (rename in particular deviates from the prototype,
where sessions are never user-named).

Full rationale is in `docs/agent-platform.md` under "The Sessions tab".

Drive-by: **the `tables` skill was stale.** It claimed the bui `Table` has no
search, filtering or sorting. As of `@backstage/ui` 0.17.0 it has all three, plus
row selection, pagination and loading/error states via `useTable` — corrected, with
the two gotchas I hit (`data: undefined` while loading, and column `id`s having to
match the fields `sortFn` reads).

Done in this PR: a "Sessions tab" section in `docs/agent-platform.md` covering the
proxy rationale, what the list can and cannot show, fleet behaviour, user scoping,
and version tolerance; new config keys with a note on their backend-only
visibility; and the Open TODOs updated.

### DIFF
--- a/plugins/agent-platform-backend/src/KagentClient.ts
+++ b/plugins/agent-platform-backend/src/KagentClient.ts
@@ -67,6 +67,23 @@ function stripTrailingSlash(url: string): string {
   return url.replace(/\/+$/, '');
 }
 
+/**
+ * kagent answered (or was reachable) and then failed: a 5xx, a 429, a timeout, or
+ * a body that could not be read.
+ *
+ * Deliberately **not** `ServiceUnavailableError`. That maps to HTTP 503, which the
+ * frontend treats as "kagent isn't deployed on this installation" and silences —
+ * correct for a host that doesn't resolve, badly wrong for a deployed-but-degraded
+ * kagent, whose sessions would then vanish from the fleet-merged list with no
+ * alert and nothing logged. An unrecognised error surfaces as a 500, which the
+ * frontend classifies as a genuine read failure and reports.
+ */
+function upstreamError(message: string): Error {
+  const error = new Error(message);
+  error.name = 'UpstreamError';
+  return error;
+}
+
 /** Whether a configured URL is absolute and http(s), so `fetch` can use it. */
 function isAbsoluteHttpUrl(url: string): boolean {
   try {
@@ -213,10 +230,25 @@ export class KagentClient {
         signal: AbortSignal.timeout(this.timeoutMs),
       });
     } catch (error) {
-      // DNS failure, TLS error, connection refused or timeout. On a fleet where
-      // most installations do not run kagent this is the normal, expected
-      // outcome, so it is logged at debug rather than warn (the root logger
-      // forwards warn/error to Sentry).
+      // A timeout means something *is* listening but did not answer in time —
+      // kagent is deployed and unwell, not absent. That must not be swallowed as
+      // "no kagent here", so it becomes an upstream failure the frontend
+      // surfaces. `AbortSignal.timeout` rejects with a TimeoutError.
+      if ((error as Error)?.name === 'TimeoutError') {
+        this.logger.debug(
+          `kagent request timed out for installation '${this.installation.name}'`,
+          { timeoutMs: this.timeoutMs },
+        );
+        throw upstreamError(
+          `The kagent API for installation '${this.installation.name}' did not respond within ${this.timeoutMs}ms.`,
+        );
+      }
+
+      // DNS failure, TLS error or connection refused: nothing is reachable at
+      // that host. On a fleet where most installations do not run kagent this is
+      // the normal, expected outcome, so it is logged at debug rather than warn
+      // (the root logger forwards warn/error to Sentry) and the frontend treats
+      // it as "not deployed here".
       this.logger.debug(
         `kagent request failed for installation '${this.installation.name}'`,
         { error: String(error) },
@@ -251,11 +283,15 @@ export class KagentClient {
     }
 
     if (!response.ok) {
+      // Anything else non-ok (5xx, 429, …) means kagent or its ingress answered
+      // and failed. It is deployed but unwell, so this must NOT share
+      // ServiceUnavailableError with "host unreachable" — the frontend silences
+      // that one, which would make a degraded kagent look like an empty account.
       this.logger.debug(
         `kagent API returned an error status for installation '${this.installation.name}'`,
         { status: response.status },
       );
-      throw new ServiceUnavailableError(
+      throw upstreamError(
         `The kagent API for installation '${this.installation.name}' returned status ${response.status}.`,
       );
     }
@@ -272,8 +308,8 @@ export class KagentClient {
     // Reading the body is a second chance to fail: the abort signal is still
     // armed after the headers arrive, so a slow or large response can abort
     // mid-stream, the connection can reset, or the body can be truncated /
-    // invalid JSON. All of those are transport failures and must map to the
-    // same 503 the frontend keys off — not escape as an unmapped 500.
+    // invalid JSON. kagent answered in all of those cases, so they are upstream
+    // failures worth surfacing — not "kagent isn't deployed here".
     try {
       return await response.json();
     } catch (error) {
@@ -281,7 +317,7 @@ export class KagentClient {
         `Failed to read the kagent API response body for installation '${this.installation.name}'`,
         { error: String(error) },
       );
-      throw new ServiceUnavailableError(
+      throw upstreamError(
         `Could not read the response from the kagent API for installation '${this.installation.name}'.`,
       );
     }
--- a/plugins/agent-platform-backend/src/KagentClient.ts
+++ b/plugins/agent-platform-backend/src/KagentClient.ts
@@ -4,7 +4,6 @@ import {
   AuthenticationError,
   NotAllowedError,
   NotFoundError,
-  ServiceUnavailableError,
 } from '@backstage/errors';
 
 /**
@@ -71,12 +70,15 @@ function stripTrailingSlash(url: string): string {
  * kagent answered (or was reachable) and then failed: a 5xx, a 429, a timeout, or
  * a body that could not be read.
  *
- * Deliberately **not** `ServiceUnavailableError`. That maps to HTTP 503, which the
- * frontend treats as "kagent isn't deployed on this installation" and silences —
- * correct for a host that doesn't resolve, badly wrong for a deployed-but-degraded
- * kagent, whose sessions would then vanish from the fleet-merged list with no
- * alert and nothing logged. An unrecognised error surfaces as a 500, which the
- * frontend classifies as a genuine read failure and reports.
+ * Deliberately **not** the same error as "kagent is absent". That case is a 404
+ * (see the catch block in `request`): silenced by the frontend, and below the
+ * `>= 500` threshold at which `MiddlewareFactory.error()` logs to Sentry, which
+ * matters because it is the expected outcome on most installations.
+ *
+ * This one surfaces as a 500, so the frontend reports it *and* it reaches Sentry —
+ * both correct here. A deployed-but-degraded kagent is rare and genuinely
+ * actionable, and its sessions silently vanishing from the fleet-merged list would
+ * be the worse failure.
  */
 function upstreamError(message: string): Error {
   const error = new Error(message);
@@ -245,16 +247,26 @@ export class KagentClient {
       }
 
       // DNS failure, TLS error or connection refused: nothing is reachable at
-      // that host. On a fleet where most installations do not run kagent this is
-      // the normal, expected outcome, so it is logged at debug rather than warn
-      // (the root logger forwards warn/error to Sentry) and the frontend treats
-      // it as "not deployed here".
+      // that host, i.e. kagent is not deployed on this installation. On a fleet
+      // where only a couple of installations run kagent, this is the *normal,
+      // expected* outcome for most of them on every page view.
+      //
+      // It must therefore be a 404 and not a 503. `MiddlewareFactory.error()`
+      // logs `logger.error` for any status >= 500, and the root logger forwards
+      // warn/error to Sentry — so a 503 here would raise one Sentry event per
+      // kagent-less installation per page view, per user, fanned out into a
+      // separate issue per installation name. Logging the cause at debug does
+      // not prevent that; only not throwing a 5xx does.
+      //
+      // 404 also matches what the frontend already does with kagent's own 404:
+      // treat it as "no kagent API here" and stay silent. The two cases carried
+      // no distinguishable meaning, so collapsing them loses nothing.
       this.logger.debug(
-        `kagent request failed for installation '${this.installation.name}'`,
+        `kagent is not reachable for installation '${this.installation.name}'`,
         { error: String(error) },
       );
-      throw new ServiceUnavailableError(
-        `Could not reach the kagent API for installation '${this.installation.name}'.`,
+      throw new NotFoundError(
+        `The kagent API is not available for installation '${this.installation.name}'.`,
       );
     }
 
--- a/plugins/agent-platform/src/apis/KagentApiClient.ts
+++ b/plugins/agent-platform/src/apis/KagentApiClient.ts
@@ -86,9 +86,30 @@ export class KagentApiClient implements KagentApi {
   async listSessions(installation: string): Promise<KagentSession[]> {
     const body = await this.get<unknown>('/kagent/sessions', installation);
     const { sessions, drift } = normalizeSessionList(body, installation);
+
     if (drift) {
       reportDrift(installation, drift);
     }
+
+    // Two drift kinds mean we cannot claim to have read this installation's
+    // sessions, so they must reach the caller's failure path rather than
+    // resolving as an empty list:
+    //
+    // - `error-envelope`: kagent reported a failure in-band on a 200. Resolving
+    //   with [] would be indistinguishable from "this user has no sessions".
+    // - `data-not-array`: the contract moved and we dropped every row.
+    //
+    // A console warning alone is not enough — nobody is watching it. Throwing
+    // puts the installation into `unreachableInstallations`, which is visible.
+    //
+    // `skipped-rows` deliberately does *not* throw: partial data is still worth
+    // showing, and the warning records what was dropped.
+    if (drift?.kind === 'error-envelope' || drift?.kind === 'data-not-array') {
+      const error = new Error(drift.message);
+      error.name = 'UpstreamError';
+      throw error;
+    }
+
     return sessions;
   }
 
--- a/plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
+++ b/plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
@@ -1,0 +1,243 @@
+import { createContext, useContext, useMemo, type ReactNode } from 'react';
+import { useApi } from '@backstage/core-plugin-api';
+import { useQueries, useQuery } from '@tanstack/react-query';
+import { useInstallations } from '@giantswarm/backstage-plugin-gs';
+import { kagentApiRef } from '../../apis';
+import { useKagentCapabilitiesMap } from '../../hooks/useKagentCapabilities';
+import { useReachableInstallations } from '../../hooks/useReachableInstallations';
+import { useAgents } from '../AgentsDataProvider';
+import {
+  buildAgentIndex,
+  isListableSession,
+  SessionRow,
+  sortSessionRows,
+  toSessionRow,
+} from './helpers';
+
+/** The backend's kagent allowlist changes with config, not with navigation. */
+const INSTALLATIONS_STALE_TIME_MS = 60 * 60 * 1000;
+
+export type SessionsContextValue = {
+  /** Sessions flattened into plain rows, most recent activity first. */
+  rows: SessionRow[];
+  /**
+   * Initial load: no rows yet and the fleet is still being queried. Only true
+   * until the first installation responds — one slow installation must not keep
+   * the whole table in a skeleton.
+   */
+  isLoading: boolean;
+  /** Rows are shown, but more installations are still resolving. */
+  isLoadingMore: boolean;
+  /** Whether any installation is configured at all. */
+  hasInstallations: boolean;
+  /**
+   * Installations we queried but couldn't read. Surfaced instead of silently
+   * dropped, so an empty list is distinguishable from a partial one.
+   *
+   * Excludes installations where kagent simply isn't deployed — that is the
+   * common case across the fleet and not something a user can act on.
+   */
+  unreachableInstallations: string[];
+  /**
+   * Installations that reported a session list which is **not** scoped to the
+   * signed-in user, because their kagent runs in `unsecure` mode. Rows are still
+   * shown; the UI must stop describing them as the user's own.
+   */
+  notUserScopedInstallations: string[];
+};
+
+const SessionsContext = createContext<SessionsContextValue | undefined>(
+  undefined,
+);
+
+/**
+ * Lists the signed-in user's kagent sessions across every reachable installation
+ * that has kagent configured, and exposes them as plain rows.
+ *
+ * Agent names and avatars are resolved against the `Agent` CRs loaded by
+ * {@link AgentsDataProvider}, so this must be mounted inside one.
+ *
+ * Unlike {@link AgentsDataProvider} this keeps **no sticky per-installation
+ * cache**. That machinery exists there because `useResources` returns fresh
+ * arrays every render and a transient error blanks them; react-query already
+ * retains the last successful `data` per installation across failed refetches,
+ * and an installation genuinely dropping out of the queried set *should* prune
+ * its rows. Please don't "restore parity" — it would add state with nothing to
+ * fix.
+ */
+export function SessionsDataProvider({ children }: { children: ReactNode }) {
+  const kagentApi = useApi(kagentApiRef);
+  const { installations } = useInstallations();
+  const allInstallations = installations.map(installation => installation.name);
+
+  // Only query reachable installations, for the same reason as the other
+  // providers: an unreachable cluster otherwise hangs for the full proxy timeout
+  // and retries, dominating the tail.
+  const { installations: reachableInstallations, isProbing } =
+    useReachableInstallations(allInstallations);
+
+  // Which installations the backend can actually reach kagent on.
+  //
+  // We deliberately *wait* for this before querying anything. kagent runs on only
+  // a couple of installations, so fanning out to the whole reachable set first
+  // would fire a doomed request per remaining installation — and each one mints
+  // that installation's Dex token before it can fail, which may involve a broker
+  // exchange. This is one cheap backend call, cached for an hour and persisted, so
+  // paying a single round-trip up front is far cheaper than N wasted ones on every
+  // cold load.
+  const {
+    data: proxiedInstallations,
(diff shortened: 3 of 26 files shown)


fiunchinho:
### Code review

Found 2 issues:

1. `agentIndex` is memoized on a key that omits the agent display name, so an in-place Agent rename leaves stale names in the Sessions table. `agentRowsKey` is built from `agent.id` only, which is `${installation}/${namespace}/${technicalName}` (`AgentsDataProvider/helpers.ts`, `toAgentRow`) and does not change when the display-name annotation is edited. `toSessionRow` resolves `agentName` from `match.name` (the display name), so after a rename `useAgents()` returns updated rows but the memo never recomputes, and the Agent column keeps showing the old name until the set of agent ids changes. `AgentsDataProvider` guards against exactly this by folding `resourceVersion` into its signature ("Include resourceVersion so an in-place edit to an existing Agent (same name, changed model/description/skills/annotation) changes the signature", added in #1952); that safeguard is dropped here.

https://github.com/giantswarm/backstage/blob/3a87016dd9b64b0e76aa7271ae91ef370ec670fe/plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx#L131-L140

2. The docs attribute the unknown-installation `400` to `404` mapping to the backend, but it happens in the frontend client. `plugins/agent-platform-backend/src/router.ts` throws `InputError` for an unknown installation, which Backstage's error handler turns into a plain HTTP 400 — the backend never emits 404. The reclassification is in `KagentApiClient.ts` (`if (response.status === 400) { error.name = 'NotFoundError'; }`), whose own adjacent comment correctly describes it as mapping a 400 received *from* the proxy.

https://github.com/giantswarm/backstage/blob/3a87016dd9b64b0e76aa7271ae91ef370ec670fe/docs/agent-platform.md#L255-L259

marians:
Thanks for the review @fiunchinho ! Actually the PR wasn't even tested locally yet. But I'll make sure your comments are addressed once I get there.

plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
```diff
@@ -0,0 +1,243 @@
...
+      // "kagent isn't deployed here" is the common case across the fleet and
+      // isn't actionable, so it contributes nothing and stays silent:
+      //   404 — the backend reached kagent's host but there is no API there
+      //   503 — the host didn't resolve, or no kagent endpoint is configured
+      //         (the backend also maps an unknown-installation 400 to 404)
+      // Anything else means we genuinely couldn't read it, which is worth saying.
+      const errorName = (query.error as Error | null)?.name;
+      const notDeployed =
+        errorName === 'NotFoundError' ||
+        errorName === 'ServiceUnavailableError';
```
marians, line 200: **Treating `ServiceUnavailableError` as "not deployed" silently swallows a broken-but-deployed kagent.**

The backend's `KagentClient.request` maps far more than "host doesn't resolve" to `ServiceUnavailableError` → HTTP 503: any non-ok status that isn't 401/403/404 (so 429, 500, 502, 504 from kagent or from oauth2-proxy), plus a body-read/`AbortSignal.timeout` failure mid-stream. All of those land here and are dropped without a word, and the `QueryClientProvider` retry predicate also declines to retry `ServiceUnavailableError`, so there is no second attempt either.

Concrete scenario: kagent on `gazelle` is deployed and healthy-ish but its DB is degraded and `/api/sessions` returns 500. The user's gazelle sessions vanish from the fleet-merged list, `unreachableInstallations` stays empty, no alert renders, nothing is logged — indistinguishable from "you have no sessions there". Same for a request that times out at 10s on a slow installation.

Worth splitting the silent path: 404 (and the 400 the proxy maps to 404) really do mean "no kagent API here"; a 503 that came from a non-ok upstream status or a timeout is a read failure and belongs in `unreachableInstallations`. That needs the backend to distinguish them (e.g. keep `ServiceUnavailableError` for connect/DNS failures and use a different error for upstream 5xx/timeouts) — otherwise the frontend has no signal to classify on.

plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
```diff
@@ -0,0 +1,243 @@
...
+    const allowed = new Set(proxiedInstallations);
+    return reachableInstallations.filter(installation =>
+      allowed.has(installation),
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [reachableKey, proxiedKey, allowlistFailed]);
+
+  const sessionQueries = useQueries({
+    queries: targets.map(installation => ({
+      queryKey: ['agent-platform', 'kagent', 'sessions', installation],
```
marians, line 122: **This query key contains no user identity, but the data is user-scoped — and the cache is persisted to `localStorage`.**

`components/QueryClientProvider.tsx` wraps everything in `PersistQueryClientProvider` with an async-storage persister over `window.localStorage`, no `shouldDehydrateQuery` filter, `maxAge`/`gcTime` of one hour. Until now everything cached there was fleet state (Agents, ModelConfigs) that is the same for every user. Sessions are not: the rows are one user's chat titles, and `['agent-platform','kagent','me',installation]` caches their `sub` (an email).

Scenario: user A opens the Sessions tab on a shared workstation, signs out; user B signs in within the hour and opens the tab. `PersistQueryClientProvider` rehydrates A's sessions cache (same origin, same key) and `SessionsDataProvider` renders A's session titles immediately — and with `staleTime: 60_000` from the provider defaults, a rehydrated entry younger than a minute is not even refetched. Independently, the titles and A's email sit on disk after sign-out.

Two possible fixes: include the Backstage user entity ref in the session/identity query keys, or exclude these keys from persistence via `persistOptions.dehydrateOptions.shouldDehydrateQuery`.

plugins/agent-platform/src/apis/KagentApiClient.ts
```diff
@@ -0,0 +1,205 @@
...
+    return (body.installations ?? [])
+      .map(installation => installation.name)
+      .filter((name): name is string => Boolean(name));
+  }
+
+  async listSessions(installation: string): Promise<KagentSession[]> {
+    const body = await this.get<unknown>('/kagent/sessions', installation);
+    const { sessions, drift } = normalizeSessionList(body, installation);
+    if (drift) {
+      reportDrift(installation, drift);
```
marians, line 91: **Drift never leaves the console, so the `error-envelope` check doesn't actually achieve what its comment claims.**

`normalizeSessionList`'s comment says the in-band `error` flag is checked because otherwise `{error: true, data: null}` "would be indistinguishable from 'this user has no sessions'". But the only consumer of `drift` is `reportDrift`, which does a single `console.warn`. `listSessions` still resolves successfully with `sessions: []`, so `SessionsDataProvider` sees `query.data === []` (truthy), takes the success branch, and the installation contributes nothing to `unreachableInstallations`.

Scenario: kagent on `gazelle` answers `200 {error: true, data: null, message: "failed to list sessions: database connection lost"}`. The user sees a list with no gazelle rows and no indication anything failed; the only trace is a browser console line no one is watching. The `data-not-array` and `skipped-rows` kinds have the same problem (a contract change that drops every row looks like an empty account).

Consider surfacing an `error-envelope` (and `data-not-array`) drift as a thrown error so the existing per-installation failure path reports it, or returning the drift from `listSessions` so the provider can put the installation into `unreachableInstallations`.

plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  const capabilitiesFor = useKagentCapabilitiesMap(targets);
+
+  const { rows: agentRows } = useAgents();
+  const agentRowsKey = agentRows.map(agent => agent.id).join('|');
+  const agentIndex = useMemo(
+    () => buildAgentIndex(agentRows),
+    // Rebuild only when the agents actually change, not on every render.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [agentRowsKey],
```
marians, line 146: **`agentRowsKey` is built from ids only, so the index keeps stale `AgentRow` objects when an agent's display name changes.**

`AgentRow.id` is `${installation}/${namespace}/${name}` — it does not change when the agent's display-name annotation, technical metadata, or anything else changes. `AgentsDataProvider` *does* pick that up (its own `readSignature` includes `resourceVersion`) and emits fresh `AgentRow` objects, but here the memo keys on `agentRowsKey` alone, so `agentIndex` keeps holding the previous render's objects and `toSessionRow` reads `match.name` off them.

Scenario: someone edits an Agent's display-name annotation from "SRE agent" to "Platform SRE". The Agents tab updates; the Sessions tab keeps showing "SRE agent" for the rest of the page session (until an agent is added or removed, which is what changes the id set).

Cheap fix: include the display name in the key, e.g. `agentRows.map(a => \`${a.id}@${a.name}\`).join('|')`.

plugins/agent-platform/src/components/SessionsIndexPage/SessionsIndexPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+    );
+  }
+
+  return (
+    <Content>
+      <Flex direction="column" gap="3">
+        <Text color="secondary">
+          Your agent chat sessions across the management clusters. kagent scopes
+          sessions to the signed-in user, so only your own are listed.
+        </Text>
```
marians, line 49: This static blurb asserts "kagent scopes sessions to the signed-in user, so only your own are listed" unconditionally, including when `notUserScopedInstallations` is non-empty — which is precisely the case the `SessionsContextValue` doc says the UI "must stop describing them as the user's own".

Scenario: one installation runs kagent in `unsecure` mode. The page then says "only your own are listed" at the top and, a couple of hundred pixels below the table, "sessions from those installations are shared rather than yours alone". The two statements contradict each other, and the reassuring one is the one a user reads first.

Softening the sentence when `notUserScopedInstallations.length > 0` (e.g. drop the second clause) would keep the page honest.

marians (commented):
Careful read of the whole diff (client + schema/normalization layer + provider + table + plugin wiring). The version-tolerant parsing, the per-installation degradation, the tri-state `isUserScoped` handling and the memo-signature discipline all hold up under scrutiny, and the telemetry case is correctly placed above the generic `/agent-platform` prefixes.

Five findings, none blocking correctness of the happy path; the first two are the ones I'd act on before merge.

plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
marians: Fixed in 5e71e44 — the backend now distinguishes "absent" from "unwell".

You're right that the mapping was far too broad, and right that the frontend had no signal to classify on. `ServiceUnavailableError` is now raised **only** for connection-level failures — DNS, TLS, connection refused — which genuinely mean nothing is deployed at that host. A `5xx`/`429`, a timeout, or a body that can't be read now raise `UpstreamError` (surfacing as a 500), which the frontend reports.

The timeout case is the one I'd have got wrong on my own: I'd been thinking of the abort as "couldn't reach it", but a timeout means something *is* listening and didn't answer — kagent deployed and slow, not absent. It's split on `error.name === 'TimeoutError'` at the fetch-rejection site, since `AbortSignal.timeout` rejects with that specifically while DNS/connect failures don't.

Your degraded-DB scenario is now a reported failure rather than a silent one, and the retry consequence you noted resolves with it: `UpstreamError` isn't in the provider's no-retry list, so a transient 500 gets retried with backoff instead of being dropped on the first attempt.

Covered by a status table (`429/500/502/503/504` → `UpstreamError`; DNS and refused → `ServiceUnavailableError`; timeout → `UpstreamError`), the three body-read cases flipped to `UpstreamError`, and `UpstreamError` added to the provider's surfaced-errors table test. The classification is also now documented as an explicit table in the backend README and in `docs/agent-platform.md`.

plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
marians: Fixed in 5e71e44 — the sessions and identity queries are no longer persisted at all.

This is the finding I'm most glad you caught. I had explicitly reasoned about credentials in this cache and verified in the browser that no token reaches it — and then missed that *the data itself* is user-scoped, which is the actual exposure. Everything cached there before was fleet state, so the persistence was safe by accident rather than by design, and I extended it to something categorically different without re-examining the assumption.

Both halves of your scenario are real: the titles and the cached email outlive sign-out on disk, and rehydration would show user A's sessions to user B under the same origin and key — with `staleTime: 60_000` not even triggering a refetch for a recent entry.

I chose exclusion-from-persistence over adding the user ref to the keys, because it removes the data-at-rest exposure rather than relocating it under a per-user key. `dehydrateOptions.shouldDehydrateQuery` filters both key shapes, **composed with `defaultShouldDehydrateQuery`** rather than replacing it — worth flagging, because overriding it outright would have silently discarded the library's "only persist successful queries" rule and started persisting errored entries.

Residual risk, stated honestly: this fixes anything involving a reload, but an account switch within one page session would still hit an in-memory cache. Including the Backstage user entity ref in the keys would close that too; I've left it out for now because it means gating the queries on an async identity lookup, on top of the allowlist gate they already wait for. Happy to add it if you'd rather not rely on sign-out reloading the app.

The filter predicate is exported and tested: both user-scoped shapes excluded, the installation allowlist and the fleet Agent/ModelConfig queries still persisted, and no over-matching on similar keys from other plugins.

plugins/agent-platform/src/apis/KagentApiClient.ts
marians: Fixed in 5e71e44 — `error-envelope` and `data-not-array` now throw.

Your reading is exactly right, and it's a case of my own comment describing an intent the code didn't implement: I wrote that the check exists so `{error: true, data: null}` isn't "indistinguishable from 'this user has no sessions'", then routed the result to a `console.warn` and resolved with `[]` — which is precisely indistinguishable from that. A warning nobody watches isn't a signal.

Both kinds now throw, so the existing per-installation failure path reports the installation instead of it contributing an innocuous empty list. I kept `skipped-rows` non-throwing on purpose: partial data is still worth showing, and the warning records what was dropped — but I've made that choice explicit in the code rather than incidental.

Three tests: an in-band error rejects with kagent's own message, a non-array `data` rejects, and a partially-unreadable list still resolves with the rows that parsed.

plugins/agent-platform/src/components/SessionsDataProvider/SessionsDataProvider.tsx
marians: Fixed in 5e71e44 — the key is now `${a.id}@${a.name}`.

Correct, and I'd missed that `AgentRow.id` is deliberately stable across content changes: `installation/namespace/name` only changes when the *set* of agents does. `AgentsDataProvider` goes to some trouble to detect in-place edits (its signature includes `resourceVersion`) and emits fresh rows, and my memo then threw that away — so the Agents tab would update while the Sessions tab kept the old name for the rest of the page session.

Took your suggested fix directly. Display name is the only field the join reads for presentation besides `technicalName`, which is already encoded in the id.

plugins/agent-platform/src/components/SessionsIndexPage/SessionsIndexPage.tsx
marians: Fixed in 5e71e44 — the reassuring clause is dropped when any installation reports that it doesn't identify users.

Good catch on the ordering point in particular: the two statements were separated by the whole table, and the one making the stronger promise was the one read first. The blurb now degrades to "Agent chat sessions across the management clusters." in that case, leaving the warning below as the only claim about scoping.

This also lines up the page with the contract the tri-state `isUserScoped` was introduced for — warn on an explicit negative, stay silent on unknown, and don't assert scoping we haven't confirmed.

plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -213,10 +230,25 @@ export class KagentClient {
...
+          `The kagent API for installation '${this.installation.name}' did not respond within ${this.timeoutMs}ms.`,
+        );
+      }
+
+      // DNS failure, TLS error or connection refused: nothing is reachable at
+      // that host. On a fleet where most installations do not run kagent this is
+      // the normal, expected outcome, so it is logged at debug rather than warn
+      // (the root logger forwards warn/error to Sentry) and the frontend treats
+      // it as "not deployed here".
       this.logger.debug(
```
marians, line 264: **The `logger.debug`-not-`warn` reasoning here doesn't hold: the error still reaches Sentry, one event per kagent-less installation per page view.** (My miss in the first pass, not something this commit introduced — but this PR is what turns the code path on, and the comment right above states the opposite of what happens.)

`httpRouterServiceFactory` installs `middleware.error()` on every plugin router, and `MiddlewareFactory.error()` does:

```js
if (options.logAllErrors || statusCode >= 500) {
  logger.error(`Request failed with status ${statusCode}`, error);
}
```

`ServiceUnavailableError` → 503 → `>= 500` → `logger.error` → the winston Sentry transport at `level: 'warn'` in `packages/backend-common/src/rootLogger.ts`, which has no matching `ignoreErrors` entry. Logging the cause at `debug` here doesn't prevent that; the throw does it.

Concrete scenario, with the default config (`app-config.yaml` derives a kagent URL for *every* `gs.installations` entry with a `baseDomain`): on a fleet of ~15 MCs where kagent runs on 2, opening the Sessions tab fires ~13 requests whose expected outcome is "no kagent here" → ~13 backend `logger.error` entries → ~13 Sentry events, per page view, per user. The identity probe (`/kagent/me`) fans out over the same set, so double it. This is precisely the "expected, already-handled outcome logged at a Sentry level" case `CLAUDE.md` warns about, and the per-installation message means one root cause fans out into an issue per installation name.

The frontend can't fix this — it deliberately treats the outcome as silent and normal. Options, roughly in order of how much they fix:

1. Don't route the expected case through the error middleware: have the router catch "kagent absent" from the client and `res.status(404).json({...})` directly (Backstage's error serialization shape is easy to reproduce), so nothing is logged at `error`.
2. Probe once per installation and cache the "absent" verdict server-side, so it costs one Sentry event per installation per backend lifetime rather than per page view.
3. At minimum, add `/^Request failed with status 503$/` — or better, a kagent-specific message pattern — to `ignoreErrors` in `rootLogger.ts`, and document why.

Worth resolving before this ships to a fleet-sized deployment; on a two-installation dev setup it won't show up.

marians (commented):
Follow-up pass on `5e71e44`. All five findings are genuinely fixed — verified against the code as it now stands, not just the replies.

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | 503 conflates "kagent absent" with "kagent unwell" | **Fixed.** `ServiceUnavailableError` is now raised only from the connect-failure branch; 5xx/429, `TimeoutError`, and all three body-read failures raise `UpstreamError`. I checked the mapping end to end: `UpstreamError` is not in `@backstage/errors`' status map, so `getStatusCodeFromError` falls through to 500, the frontend leaves `error.name === 'Error'`, and the provider's `notDeployed` check excludes it → reported *and* retried, since neither name is in the retry predicate's decline list. The `TimeoutError` discriminator is right: `AbortSignal.timeout()`'s abort reason is a `DOMException` named `TimeoutError` and fetch rejects with the signal's reason, while DNS/connect failures don't. Reasoning about the timeout being "deployed and slow, not absent" is the correct call. |
| 2 | User-scoped data persisted to `localStorage` | **Fixed.** `dehydrateOptions.shouldDehydrateQuery` is a real `PersistQueryClientOptions` field, `defaultShouldDehydrateQuery` is exported by the installed `@tanstack/react-query` 5.101.2, and composing with `&&` rather than replacing keeps the "successful queries only" rule — that detail was worth flagging. The predicate matches both user-scoped key shapes, leaves `['agent-platform','kagent','installations']` persisted, and doesn't over-match. The residual you name honestly (in-session account switch still hits the in-memory cache) is a fair trade to defer; the data-at-rest exposure, which was the actual problem, is gone. One transient wrinkle not worth code: an existing pre-fix `localStorage` blob is still rehydrated once on the first load after deploy, then overwritten by the next persist. |
| 3 | Drift never left the console | **Fixed.** `error-envelope` and `data-not-array` now throw, so the installation lands in `unreachableInstallations`; `skipped-rows` still resolves with the rows that parsed. Keeping partial data non-fatal is the right line to draw. Note the consequence you may or may not want: `{error: true, data: [...some rows...]}` now shows none of those rows instead of some — I think that's the better default, just flagging it's a behaviour change beyond the null case. |
| 4 | `agentRowsKey` missing the display name | **Fixed.** `${a.id}@${a.name}` — `technicalName` is already encoded in the id, so that's complete coverage of what the join reads. |
| 5 | Page contradicting its own warning | **Fixed.** Conditional on `notUserScopedInstallations.length > 0`. Only cosmetic leftover: the reassuring wording shows while the probe is unresolved and can flip once it lands, which is unavoidable without blocking the page on the probe. |

Also checked the fix commits on their own merits: no unused `ServiceUnavailableError` import, the new backend status table in the README matches the code exactly, and the new tests assert the behaviour rather than the implementation. The one test that's vacuous-but-harmless is `'UpstreamError'` in the provider's surfaced-errors table — the backend path arrives as `'Error'` (500 isn't renamed in `handleResponse`); the name only occurs for the client's own drift throw. Both are surfaced, so the assertion holds either way.

One new finding, and it's a first-pass miss on my side rather than something the fix introduced — inline on `KagentClient.ts`: the "logged at `debug` so it doesn't reach Sentry" reasoning on the connect-failure branch is wrong, because the plugin's error middleware logs every `>= 500` response at `error`. With the default config deriving a kagent URL for every installation with a `baseDomain`, the *expected* "no kagent here" outcome becomes a Sentry event per installation per page view. Details and options in the inline comment.

plugins/agent-platform-backend/src/KagentClient.ts
marians: Fixed in e485d2f. I verified every link in the chain before touching anything, and all four hold:

- `logAllErrors` is not set anywhere in our config, so the `statusCode >= 500` branch is the only one that logs.
- `MiddlewareFactory.error()` is exactly as you quote it (`node_modules/@backstage/backend-defaults/.../MiddlewareFactory.cjs.js:194`).
- the winston Sentry transport in `rootLogger.ts` is at `level: 'warn'`.
- `ignoreErrors` has two entries (techdocs indexer, PagerDuty legacy config); neither matches.

So the 503 did reach Sentry, and your arithmetic holds. Worse: **my previous commit made this sharper rather than introducing it.** Splitting "absent" from "unwell" left `ServiceUnavailableError` meaning *specifically* the expected not-deployed case — so I concentrated the expected outcome onto the one path that logs, while adding a comment asserting it was quiet.

I took a variant of your option 1, but simpler than bypassing the middleware: connection-level failures (DNS, TLS, refused) now raise `NotFoundError` → **404**. That is below the logging threshold, so nothing is written at `error` and no event is generated — better than filtering one after the fact (your option 3), and it makes the server-side caching in option 2 unnecessary for noise purposes.

It needs no frontend change, because 404 is already the silent "no kagent API here" path — which is the other reason I prefer it to a bespoke status: the two cases (kagent's own 404 vs. host unreachable) never carried a distinguishable meaning for any consumer, so collapsing them removes a distinction rather than adding one.

Your point about the misleading comment is taken directly: I've rewritten it to say that `logger.debug` on the cause does *not* keep this out of Sentry, since the throw is what the middleware logs after the client is done. That was the actual error in my reasoning, and it's the kind that survives review by looking like a justification.

Two things kept deliberately as 5xx:

- **`UpstreamError` (500)** for a 5xx/429, timeout or unreadable body — a deployed-but-degraded kagent is rare and genuinely actionable, so an event is wanted.
- **`ServiceUnavailableError` (503)** for "no installation configured at all" — a real misconfiguration, and unreachable from the UI anyway: an empty allowlist means the frontend queries nothing.

Two router tests now pin the property rather than the implementation: an unreachable kagent must come back `< 500`, and a degraded one must come back `>= 500`. The reasoning is documented in the backend README under "Never return a 5xx for an expected outcome" and in `docs/agent-platform.md`, since this is exactly the sort of constraint that erodes the next time someone adjusts an error mapping.

Agreed on your last point, and worth stating plainly: this would not have shown up in my local two-installation verification. It needed reading the middleware, not running the app.

PR #2001, opened 2026-07-28
feat(flux-react): add Reconcile and Suspend buttons to resource cards

marians:
### What does this PR do?

Adds **Reconcile** and **Suspend**/**Resume** buttons to the footer of the Flux resource cards in the details panel, and removes the now-redundant `flux reconcile` / `flux suspend` / `flux resume` entries from the card's copy-command menu (`kubectl get -o yaml` and `kubectl describe` stay).

The buttons appear on every card in the panel — the selected resource as well as its source and dependency cards — so e.g. a `GitRepository` can be reconciled straight from the `HelmRelease` panel.

All eight kinds the panel renders get the buttons, **`ImagePolicy` included**. An earlier revision of this PR excluded it, claiming Flux supported neither operation for the kind; that was wrong, as [pointed out in review](https://github.com/giantswarm/backstage/pull/2001#discussion_r3665559595). Both `crds.fluxcd.v1.ImagePolicy` and `v1beta2` declare `spec.suspend` _and_ `status.lastHandledReconcileAt`, i.e. image-reflector-controller honours exactly the mechanism this PR uses. What Flux actually lacks is the _CLI_ subcommand — there is no `flux reconcile image policy` — which is why the old copy-command map omitted the kind and why the allowlist inherited from it was wrong.

- **Reconcile** patches the `reconcile.fluxcd.io/requestedAt` annotation, equivalent to plain `flux reconcile <kind> <name>`. No `--with-source`; since source cards have their own button, that is two clicks away.
- **Suspend / Resume** is a single button that patches `spec.suspend`; its label, icon and target value follow the current state. No confirmation dialog — it is immediately reversible, and the result is reported via a toast.

Reconcile is disabled in two cases, each with a tooltip explaining why:

1. While the resource is **suspended**, because a suspended resource ignores the annotation.
2. Until a requested reconciliation has been **picked up by the controller**, determined the same way Flux itself does it: comparing the annotation against `status.lastHandledReconcileAt` (new `FluxObject.isReconcileRequestPending()`). Because that reads off the resource rather than local UI state, the button is also disabled for a request someone else made — via the `flux` CLI, say — and the state survives a page reload.

The two common Flux operations no longer require leaving Backstage for a terminal.

**Users with read-only cluster access see no buttons at all.** This is the first Kubernetes write path from the browser in this repo, so the write affordances are gated on a `SelfSubjectAccessReview` (`verb: patch`) against the target cluster before anything is rendered. The check:

- **Fails closed** — while the review is in flight, or if it fails, nothing is shown; a button never flashes in and then disappears.
- Is **cached per cluster + API group + resource + namespace**, so one review covers every card of that kind in that namespace rather than one call per card.
- Works for read-only users, since the built-in `system:basic-user` ClusterRole grants `create` on `selfsubjectaccessreviews` to every authenticated user.

Known limitation, accepted deliberately: omitting the resource name from the review means a user granted `patch` through an RBAC rule using `resourceNames` will not see the buttons even though the write would succeed. This fails closed, so nothing breaks. If it turns out to affect real users, the cheap fix is a second-tier name-scoped review only when the namespace-wide one is denied.

A rejected patch is still reported as a permission error toast — `SelfSubjectAccessReview` is advisory, and a webhook authorizer can refuse where RBAC would allow.

### How does it look like?

Footer of a resource card, alongside the existing copy-command and AI-chat buttons:

image

No backend or config change was needed. `kubernetesApi.proxy()` already forwards an arbitrary `RequestInit`, and the backend `/proxy` handler is mounted for all HTTP methods — and, importantly, _before_ `express.json()`, so a PATCH body streams through untouched. Authorization is delegated entirely to the target apiserver via the signed-in user's own OIDC token.

Three details in `kubernetes-react` worth a reviewer's attention:

- `patchResource` passes `Content-Type` as a **plain object**. `KubernetesClient.getKubernetesHeaders` object-spreads `init.headers`, and a `Headers` instance spreads to `{}` — which would silently drop the content type. There is a test pinning this.
- New `KubeObject.getResolvedGVK()` reports the API version an object was actually _read_ at. Reads resolve their version through discovery, so building a write from the class-level `getGVK()` could target a version the cluster does not serve, and would produce cache keys that do not match the ones the read hooks registered.
- The access-review and mutation query keys deliberately avoid the `'cluster'` prefix: `useClusterQueries` filters the active react-query cache on `queryKey[0] === 'cluster'` to drive the per-cluster loading/error banners, and a permission probe must not surface there.

Two smaller notes:

- A pending reconcile request now also triggers the fast (3s) refetch interval in `useFluxResources`. Without it the button would have sat on the 15s poll and looked stuck for far longer than the controller actually takes. The existing `isReconciling()` check there covered only four of the eight collections; the new check covers all eight.
- The disabled Reconcile button uses a **MUI v4** `Tooltip` around a `<span>` rather than the bui one: a disabled react-aria button sets the `disabled` attribute and so receives no hover events, meaning a bui `Tooltip` on it would never appear.

`flux-react` gained two **devDependencies** (`@tanstack/react-query`, `@backstage/plugin-kubernetes-react`) that the new tests need — `no-undeclared-imports` correctly rejected them otherwise. Runtime dependencies are unchanged; the component imports neither.

Verified manually in a local instance against a real management cluster, plus `yarn test` (162 tests / 23 suites), `yarn tsc`, `yarn lint` and `yarn prettier:check`.

No.


plugins/flux-react/src/hooks/useFluxResources/useFluxResources.ts
```diff
@@ -229,15 +229,35 @@ export function useFluxResources(clusters: string | string[] | null) {
...
+    const requestPending = [
+      kustomizations,
+      helmReleases,
+      gitRepositories,
+      ociRepositories,
+      helmRepositories,
+      imagePolicies,
+      imageRepositories,
+      imageUpdateAutomations,
+    ].some(resources => resources.some(r => r.isReconcileRequestPending()));
```
marians: `isReconcileRequestPending()` stays true for as long as the annotation differs from `status.lastHandledReconcileAt`, and there are cases where that never converges — the easiest one is reachable with this PR's own buttons: click **Reconcile**, then **Suspend** before the controller picks the request up. Flux's reconcilers return early on `spec.suspend` *without* patching `status`, so `lastHandledReconcileAt` stays behind for as long as the object is suspended. Same for an object no controller is reconciling: a `HelmRepository` with `spec.type: oci` (source-controller treats those as static and never writes a status), or a CRD installed without its controller running.

Because this single `refetchInterval` is shared by all eight `useResources` calls, one such object pins the whole Flux page at the 3s poll indefinitely — 8 list requests per selected cluster every 3 seconds, for every user viewing that installation, and it never falls back to the 15s baseline until someone removes the annotation by hand. Note this also scans `imagePolicies`, which get no button and so can only ever have been annotated out-of-band.

At minimum I'd exclude suspended resources here (`r => !r.isSuspended() && r.isReconcileRequestPending()`), which also fixes the related UI symptom: for a suspended-with-outstanding-request object the Reconcile button is disabled with the "waiting to be picked up" tooltip, which is not what is actually blocking it. For the no-controller case some bound on how long a pending request can hold the fast interval would be safer.

plugins/kubernetes-react/src/hooks/useSelfSubjectAccessReview/useSelfSubjectAccessReview.ts
```diff
@@ -0,0 +1,62 @@
...
+      verb,
+    ],
+    queryFn: () =>
+      createSelfSubjectAccessReview({
+        kubernetesApi,
+        cluster,
+        resourceAttributes,
+      }),
+    enabled: (options.enabled ?? true) && Boolean(cluster),
+    staleTime: STALE_TIME,
```
marians, line 53: This verdict gets persisted to `localStorage`. `plugins/flux/src/components/QueryClientProvider.tsx` wraps everything in `PersistQueryClientProvider` with `createAsyncStoragePersister({ storage: window.localStorage })` and no `dehydrateOptions`, so the default `shouldDehydrateQuery` (any query in `success` state) dehydrates this one too, with `maxAge` = 1h — and nothing ever calls `persister.removeClient()`, including on sign-out.

Concrete scenario: someone with `patch` uses the Flux page, signs out, and a read-only colleague signs in on the same browser (or the same user's grant is revoked). On the flux page the cache rehydrates `allowed: true` for that cluster/group/resource/namespace, so `canPatch` is true on first paint and Reconcile/Suspend render — then vanish a moment later when the background revalidation (stale on mount, `staleTime` 5min) returns `false`. That's exactly the "button flashes in and then disappears" the fail-closed design is meant to prevent, plus a 403 error toast for anyone who clicks fast enough.

Cheapest fix is to keep permission probes out of the persisted cache — e.g. tag the query (`meta: { persist: false }`) and add a `dehydrateOptions.shouldDehydrateQuery` filter in the plugin `QueryClientProvider`s that honours it.

plugins/kubernetes-react/src/lib/k8s/KubeObject.ts
```diff
@@ -143,4 +146,25 @@ export class KubeObject<T extends KubeObjectInterface = any> {
...
+   * so writes and cache invalidations must use this — targeting the static
+   * version can hit a version the cluster does not serve, and produces query
+   * keys that do not match the ones the read hooks registered.
+   */
+  getResolvedGVK(): CustomResourceMatcher {
+    const ctor = this.constructor as typeof KubeObject;
+
+    return {
+      apiVersion: this.getApiVersionSuffix(),
+      group: this.getGroup() ?? ctor.group,
```
marians: `getGroup()` is `apiVersion.split('/')[0]`, so for a core object (`apiVersion: 'v1'`) it returns `'v1'`, not `''` — the `?? ctor.group` fallback only kicks in when `apiVersion` is missing entirely. For any `isCore` resource this therefore yields `{ group: 'v1', apiVersion: 'v1' }`, with two consequences for the two things this method exists for:

- the invalidation keys in `useFluxResourceActions` become `['cluster', c, 'list', 'v1', 'v1', plural]`, while `useListResources`/`useGetResource` registered `['cluster', c, 'list', 'v1', plural]` (their `.filter(Boolean)` drops the empty group) — so `invalidateQueries` matches nothing and the UI keeps rendering pre-mutation state after a successful write;
- an access review built from it asks about `group: 'v1'`, which no authorizer recognises → always `allowed: false` → the write affordances never appear.

Latent today since only Flux CRs use it, but this is a documented general helper on `KubeObject`, so worth pinning down now.

```suggestion
      group: ctor.isCore ? '' : (this.getGroup() ?? ctor.group),
```

plugins/flux-react/src/components/FluxOverview/ResourceCard/utils/fluxActions.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  FluxObject,
+  KubeObject,
+} from '@giantswarm/backstage-plugin-kubernetes-react';
+
+/**
+ * Kinds that Flux can reconcile on demand and suspend.
+ *
+ * `ImagePolicy` is absent on purpose: Flux has neither a reconcile request nor
```
marians: This justification does not hold up against the CRD schemas: both `crds.fluxcd.v1beta2.ImagePolicy` and `crds.fluxcd.v1.ImagePolicy` declare `spec.suspend` *and* `status.lastHandledReconcileAt` (see `node_modules/@giantswarm/k8s-types/dist/types/crds/fluxcd/v1/ImagePolicy.d.ts`, lines 126 and 171), i.e. image-reflector-controller honours both the reconcile-request annotation and suspend for ImagePolicy — the same mechanism this PR uses.

What Flux actually lacks is the *CLI* subcommand (`flux reconcile image policy` / `flux suspend image policy` don't exist), which is why the old copy-command map omitted the kind. Not a bug, but the comment will mislead the next reader, and it means we're withholding two operations that would work. Same wording appears in the changeset and the PR description.

marians (commented):
Careful diff pass. The core mechanics check out — I verified independently that the upstream `/proxy` router is mounted before `express.json()` (so the PATCH body streams through), that both `getKubernetesHeaders` implementations object-spread `init.headers` (hence the plain-object `Content-Type`), that the invalidation keys match what `useListResources`/`useGetResource` register, that `backend.cors.methods` already includes `PATCH`, and that the effect dep list now covers all eight collections. Four things below, one of them worth fixing before merge.

plugins/flux-react/src/components/FluxOverview/ResourceCard/utils/fluxActions.ts
marians: Confirmed and fixed in 73fad4f1 — you are right on both counts.

I checked the schemas directly: `ImagePolicy.d.ts` (v1 and v1beta2) declares `spec.suspend` ("tells the controller to suspend subsequent policy reconciliations") and `status.lastHandledReconcileAt` ("so a change of the annotation value can be detected"). The kind now gets both buttons rather than just a corrected comment, since withholding working operations was the worse half of the problem. `ImagePolicy` already extended `FluxObject`, so no hierarchy change was needed.

The allowlist comment now explains what it actually encodes — CRDs carrying `spec.suspend` + `status.lastHandledReconcileAt` — and calls out explicitly that this is a *wider* set than the `flux` CLI subcommands, so the next reader does not re-derive the wrong rule. Same correction applied to the changeset and the PR description.

Kept it as an explicit allowlist rather than collapsing to `instanceof FluxObject`, so a future Flux resource class without those fields does not silently get buttons that cannot work. New tests cover an ImagePolicy getting both actions (asserting the review asks about `imagepolicies`/`image.toolkit.fluxcd.io`), suspending one, and Resume for a suspended one; the negative case now uses a `ConfigMap` to exercise the non-Flux guard.

This also resolves the sub-point in your `useFluxResources` comment about the pending-request scan covering `imagePolicies` that had no button — they have buttons now.

plugins/kubernetes-react/src/lib/k8s/KubeObject.ts
marians: Confirmed and applied your suggestion in c1eaa222.

Verified the reachability: `ConfigMap`, `Secret` and `Namespace` all set `isCore = true` and none of them declares a `static group`, so `ctor.group` is `undefined` and the fallback could not have rescued it even if it had fired. Both consequences you list are exactly right.

Added `KubeObject.test.ts` covering the custom-resource case, the object-version-over-class-default case, an empty group for a core resource, and — most usefully — that a key built from `getResolvedGVK()` collapses under `.filter(Boolean)` to the same shape `useListResources` registers. That last one is the assertion that would actually have caught this.

plugins/kubernetes-react/src/hooks/useSelfSubjectAccessReview/useSelfSubjectAccessReview.ts
marians: Confirmed and fixed in c1eaa222, taking your suggested approach.

I verified the whole chain: all three persisting providers (`flux`, `gs`, `agent-platform`) pass `persistOptions={{ persister, maxAge }}` with no `dehydrateOptions`, and a repo-wide grep for `removeClient` / `dehydrateOptions` / `shouldDehydrateQuery` returned nothing at all — so there was no cache clearing anywhere, on sign-out or otherwise.

Rather than filter by query key in each provider, `kubernetes-react` now exports the convention: `NON_PERSISTED_QUERY_META` to tag a query and a `shouldDehydrateQuery` that keeps the library default (successful queries only) and additionally drops anything tagged. The access review carries the tag; all three providers pass the filter. One definition, and it generalises to any future identity- or permission-scoped query.

Note this only stops *new* writes — a browser that already has a verdict in localStorage from an earlier build keeps it until `maxAge` (1h) expires. Not worth a migration in my view, but worth knowing if you test the scenario on a browser that has used the current PR build.

Tests cover the filter directly, including that unrelated `meta` does not accidentally opt a query out.

plugins/flux-react/src/hooks/useFluxResources/useFluxResources.ts
marians: Agreed — this was the real bug of the four, and it was mine. Fixed in c1eaa222.

The predicate moved to `awaitsReconcileHandling.ts` (extracted so it is unit-testable without mocking eight `useResources` calls) and is now deliberately stricter than `isReconcileRequestPending()`:

- **Suspended objects excluded**, as you suggested.
- **The acceleration is bounded**: a request only speeds up the poll if its annotation parses as a timestamp within 2 minutes of now. That covers the no-controller case you raise, plus an annotation applied out-of-band long ago, and it degrades safely — the value is an opaque token to Flux, so an unparseable one gets no acceleration at all. Clock skew in either direction is tolerated via `Math.abs`.

I kept the button's own disabled state on the plain `isReconcileRequestPending()`, since "Flux still considers this outstanding" is the honest answer there; only the polling policy needs the narrower test. The tests assert that split explicitly — for a suspended or stale request, `awaitsReconcileHandling` is false while `isReconcileRequestPending` stays true.

One correction on a sub-point: the tooltip is already right. `FluxResourceActions` checks `isSuspended` first, so a suspended-with-outstanding-request object shows "Suspended resources are not reconciled. Resume it first." rather than the waiting-to-be-picked-up text. Only the polling half needed fixing.

The `imagePolicies` observation resolved differently — they now get buttons too, per the ImagePolicy thread.

PR #2002, opened 2026-07-29
feat(agent-platform-backend): proxy kagent session detail and tasks

marians:
### What does this PR do?

Adds the transport the upcoming Agent Platform **session detail page** needs. Two new proxy routes, both requiring `?installation=` and a forwarded user token, both passing kagent's JSON through verbatim like the existing ones:

| Route | Purpose |
| --- | --- |
| `GET /kagent/sessions/:id` | One session plus its stored events |
| `GET /kagent/sessions/:id/tasks` | The session's A2A tasks — conversation, state, token usage |

Nothing calls them yet; the frontend lands in follow-up PRs.

None yet — no UI consumes these routes in this PR.

Three decisions worth reviewing:

**1. The conversation comes from `…/tasks`, not from `…/sessions/:id`'s `events`.** This is what kagent's own UI renders from (`ui/src/components/chat/ChatInterface.tsx` → `extractMessagesFromTasks`), and the two payloads carry genuinely different things:

| | `events[]` | `Task[]` |
| --- | --- | --- |
| Message content | `data` is a JSON **string** holding an A2A message (doubly encoded) | `history` already structured |
| Session state | — | `status.state` |
| Token usage | — | per-message `{adk,kagent}_usage_metadata` |
| Per-item timestamp | `created_at` | only one per task |

So the frontend will fetch both: tasks for the timeline and state, events only to recover per-message timestamps — A2A messages carry none of their own. Hence two routes rather than one.

**2. No `A2A-Version` header is sent.** kagent's `NegotiateA2AWireVersion` treats a missing header as the legacy v0 wire on both v0.9.9 (our pin) and v0.10, which is the shape kagent's UI consumes and therefore the best-tested one. There's a test asserting the header stays absent so it isn't added casually; opting into the v1 wire should be a deliberate migration.

**3. Session ids stay opaque.** Real responses mix 64-character hex strings and UUIDs, so nothing validates a format — only that a path segment is present. Ids are `encodeURIComponent`-ed before interpolation, with a test that a `/` or `?` in an id can't retarget the request.

One consequence to be aware of: a session belonging to **another user** answers 404 exactly as a deleted one does, because kagent scopes the lookup by the token's user id. Both are expected outcomes for a stale deep link, so neither returns a 5xx — anything `>= 500` is logged at `error` by `MiddlewareFactory.error()` and forwarded to Sentry. Both cases are covered by tests asserting `< 500`.

Backend tests: 54 → 72.

Yes, done in this PR — the plugin README gains the two routes, the events-vs-tasks comparison, the wire-version rationale, and the id-opacity note.

### DIFF
--- a/.changeset/agent-platform-session-detail-proxy.md
+++ b/.changeset/agent-platform-session-detail-proxy.md
@@ -1,0 +1,36 @@
+---
+'@giantswarm/backstage-plugin-agent-platform-backend': minor
+---
+
+Proxy kagent's session detail and session tasks, the transport the upcoming
+Agent Platform session detail page needs.
+
+- `GET /kagent/sessions/:id` — one session plus its stored events.
+- `GET /kagent/sessions/:id/tasks` — the session's A2A tasks, which carry the
+  conversation (`history`), its state (`status.state`) and per-message token
+  usage.
+
+Both require `?installation=` and a forwarded user token, and both pass kagent's
+JSON through verbatim, as the existing routes do.
+
+Two details worth knowing:
+
+- **The conversation comes from `…/tasks`, not from `…/sessions/:id`'s `events`.**
+  That is what kagent's own UI renders from, and the two payloads carry different
+  things: an event's `data` is a JSON _string_ holding an A2A message (doubly
+  encoded) and carries no state or token usage, while task history is already
+  structured and carries both. Events are still useful for one thing — A2A
+  messages have no timestamp of their own, so per-message times come from the
+  events' `created_at`.
+- **No `A2A-Version` header is sent.** kagent's `NegotiateA2AWireVersion` treats a
+  missing header as the legacy v0 wire on both v0.9.9 and v0.10, which is the
+  shape kagent's UI consumes and therefore the best-tested one. Opting into the
+  v1 wire is a deliberate future migration.
+
+Session ids stay opaque — real responses mix 64-character hex strings and UUIDs,
+so nothing validates a format; ids are URL-encoded before being interpolated.
+
+A session belonging to another user answers **404**, exactly as a deleted one
+does, because kagent scopes the lookup by the token's user id. Both are expected
+outcomes for a stale deep link, so neither returns a 5xx — which
+`MiddlewareFactory.error()` would log at `error` and forward to Sentry.
--- a/.changeset/agent-platform-session-detail-proxy.md
+++ b/.changeset/agent-platform-session-detail-proxy.md
@@ -28,9 +28,29 @@ Two details worth knowing:
   v1 wire is a deliberate future migration.
 
 Session ids stay opaque — real responses mix 64-character hex strings and UUIDs,
-so nothing validates a format; ids are URL-encoded before being interpolated.
+so nothing validates or normalizes one. In particular they are not trimmed:
+Express hands over the decoded segment, so trimming would re-encode a _different_
+id and 404 in a way indistinguishable from a missing session.
 
 A session belonging to another user answers **404**, exactly as a deleted one
 does, because kagent scopes the lookup by the token's user id. Both are expected
 outcomes for a stale deep link, so neither returns a 5xx — which
 `MiddlewareFactory.error()` would log at `error` and forward to Sentry.
+
+404 messages are now per-endpoint, because three different things arrive as one:
+
+- Nothing listening at the host (`fetch` rejects) — "kagent is not available
+  here", the fleet-wide wording the frontend's silent classification relies on.
+- kagent answering "no such resource" (404 with a JSON body, since its error
+  middleware always answers JSON) — "that session does not exist". Previously
+  this read "The kagent API is not available for installation X", i.e. a
+  bookmarked link to a deleted session reported an outage on a healthy
+  installation.
+- The endpoint not existing (404 with a non-JSON body, because kagent registers
+  no custom `NotFoundHandler` and net/http answers `text/plain`) — "this kagent
+  predates that endpoint". Without this, an installation on an older kagent would
+  report "session not found" for every session forever, and with no version probe
+  there would be nothing else to go on.
+
+kagent's own 404 message is not forwarded: its middleware appends the underlying
+error, so a session 404 reads `Session not found: no rows in result set`.
--- a/.changeset/agent-platform-session-detail-proxy.md
+++ b/.changeset/agent-platform-session-detail-proxy.md
@@ -5,7 +5,7 @@
 Proxy kagent's session detail and session tasks, the transport the upcoming
 Agent Platform session detail page needs.
 
-- `GET /kagent/sessions/:id` — one session plus its stored events.
+- `GET /kagent/sessions/:id` — the session object.
 - `GET /kagent/sessions/:id/tasks` — the session's A2A tasks, which carry the
   conversation (`history`), its state (`status.state`) and per-message token
   usage.
@@ -13,16 +13,18 @@ Agent Platform session detail page needs.
 Both require `?installation=` and a forwarded user token, and both pass kagent's
 JSON through verbatim, as the existing routes do.
 
-Two details worth knowing:
+Three details worth knowing:
 
-`GET /kagent/sessions/:id` asks kagent for `limit=1`, because the caller wants the
-session object and nothing else. kagent bundles the session's stored events into
-that response and they dominate it — on a real 4-turn session, 591 KB of events
-against 261 bytes of session metadata. Nothing reads them (see below), so this
-trims the request by ~99%. Both v0.9.9 and v0.10 parse `limit` with the same
-handler code, and a version that ignored it would simply return everything, which
-is the previous behaviour.
-
+- **`GET /kagent/sessions/:id` asks kagent for `limit=1`.** The caller wants the
+  session object and nothing else, but kagent bundles the session's stored events
+  into that response and they dominate it — on a real 4-turn session, 591 KB of
+  events against 261 bytes of session metadata. Nothing reads them (see below), so
+  this trims the response by ~99%. It has to be `1`, not `0`: kagent's DB layer
+  gates the LIMIT clause on `opts.Limit > 0`, so `limit=0` means _unlimited_ and
+  would quietly restore the full payload. Both v0.9.9 and v0.10 honour the param —
+  v0.9.9 parses it inline in `HandleGetSession`, v0.10 in
+  `eventQueryOptionsFromRequest` — and a version that ignored it would simply
+  return everything, which is the previous behaviour.
 - **The conversation comes from `…/tasks`, not from `…/sessions/:id`'s `events`.**
   That is what kagent's own UI renders from, and only task history is structured
   as A2A messages carrying the session's state and token usage. The `events` array
--- a/.changeset/agent-platform-session-detail-proxy.md
+++ b/.changeset/agent-platform-session-detail-proxy.md
@@ -15,13 +15,22 @@ JSON through verbatim, as the existing routes do.
 
 Two details worth knowing:
 
+`GET /kagent/sessions/:id` asks kagent for `limit=1`, because the caller wants the
+session object and nothing else. kagent bundles the session's stored events into
+that response and they dominate it — on a real 4-turn session, 591 KB of events
+against 261 bytes of session metadata. Nothing reads them (see below), so this
+trims the request by ~99%. Both v0.9.9 and v0.10 parse `limit` with the same
+handler code, and a version that ignored it would simply return everything, which
+is the previous behaviour.
+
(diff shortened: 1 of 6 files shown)


plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -184,6 +184,52 @@ export class KagentClient {
     return this.request(`${this.installation.apiBaseUrl}/sessions`, options);
   }
 
+  /**
+   * `GET <apiBaseUrl>/sessions/<id>` — one session plus its stored events.
+   *
+   * kagent scopes this by the forwarded token's user id, so a session belonging
+   * to somebody else is indistinguishable from one that does not exist: both
+   * answer 404. That is an expected outcome for a stale or shared deep link, and
+   * `request` already maps it to `NotFoundError` (404) rather than a 5xx.
```
marians, line 214: The status code is right, but the message the frontend receives is not. `request()` maps every 404 to a single fixed string:

```
Throws NotFoundError(`The kagent API is not available for installation '${name}'.`)
```

and `MiddlewareFactory.error()` serialises that message into the response body (verified: the body is `{error: {name, message, status}, ...}`). So the concrete scenario is: a user opens a bookmarked link to a session that was deleted, kagent answers 404, and the session detail page renders *"The kagent API is not available for installation 'gazelle'."* — i.e. it reports an outage for a perfectly healthy installation.

The same conflation bites a second way, which matters more for `listSessionTasks`: if an installation runs a kagent build without `/api/sessions/{id}/tasks`, the route-missing 404 is byte-for-byte indistinguishable from a session-missing 404. Since there is deliberately no version probe, "this kagent is too old" will surface to users as "session not found" on every session, on every page load.

Worth letting the caller supply the not-found message (or distinguishing on kagent's `{error, message}` envelope), so the session routes can say "session not found" while `/sessions` and `/me` keep the current wording.

plugins/agent-platform-backend/src/router.ts
```diff
@@ -154,6 +154,23 @@ export async function createRouter(
...
+   * Read the session id from the path.
+   *
+   * Session ids are **opaque**: real kagent responses mix 64-character hex
+   * strings and UUIDs, so nothing may validate a format. Express has already
+   * decoded the segment and cannot match across `/`, so "non-empty" is the only
+   * check left worth making.
+   */
+  const readSessionId = (req: express.Request): string => {
+    const raw = req.params.sessionId;
+    const sessionId = typeof raw === 'string' ? raw.trim() : '';
```
marians: Two small things here, both stemming from the doc comment's premise.

1. `.trim()` is a mutation of a value the surrounding docs declare **opaque** ("nothing may validate a format"). If kagent ever hands out an id with a leading or trailing space, the frontend percent-encodes it, Express decodes it back, this trims it away, and `encodeURIComponent` then sends a *different* id upstream — a silent 404 that looks like a missing session. Trimming is a format assumption, just a quiet one.

2. The `if (!sessionId)` guard is unreachable under Express 5 / path-to-regexp v8: `:sessionId` does not match an empty segment, so `GET /kagent/sessions/` falls through to the list route above (confirmed by running it — `listSessions` is called, `getSession` is not). The `InputError` can never be thrown, which also makes the "non-empty is the only check left worth making" comment misleading for the next reader.

Either drop the helper and read `req.params.sessionId` directly, or keep it as a typing shim without the trim:

```suggestion
    const sessionId = typeof raw === 'string' ? raw : '';
```

plugins/agent-platform-backend/src/KagentClient.ts
marians: Both halves are right, and the second one is the more serious of the two — fixed in ecb9c308.

Rather than only letting the caller supply a message, I could distinguish the two 404s, which addresses the "old kagent" case properly. kagent's error middleware always answers `Content-Type: application/json` (`go/core/internal/httpserver/middleware_error.go` — `w.Header().Set("Content-Type", "application/json")` before encoding `{error: ...}`), while an unrouted path falls through to net/http's `http.NotFound` and answers `text/plain`, because kagent registers no custom `NotFoundHandler` (grepped: no `NotFoundHandler` anywhere in `go/`). So there are three cases, not two:

| Actually happened | How we know | Message |
| --- | --- | --- |
| Nothing listening at the host | `fetch` rejected | kagent is not available here |
| kagent said "no such resource" | 404 + JSON body | that session does not exist |
| Endpoint doesn't exist | 404 + non-JSON body | this kagent predates that endpoint |

`/sessions` and `/me` keep the old wording, since a 404 there genuinely does mean "no kagent here" and the frontend's silent classification is built around it. There's a test pinning that too, so the fleet-wide path can't drift by accident.

One thing I did **not** do: forward kagent's own message. Its middleware appends the underlying error, so the session 404 actually reads `Session not found: no rows in result set` — I'd rather not put `pgx` internals in front of a user. Test asserts the message doesn't contain it.

Status stays 404 in all three cases; only the wording varies. Asserted separately so the Sentry guarantee doesn't get lost in the message work.

plugins/agent-platform-backend/src/router.ts
marians: Both correct — I verified each before changing anything, since they contradicted my own comment.

Ran a throwaway Express 5.2.1 app with the same two routes:

```
/kagent/sessions               200 -> list
/kagent/sessions/              200 -> list        <- your point 2
/kagent/sessions/abc           200 -> detail:"abc"
/kagent/sessions/%20abc%20     200 -> detail:" abc "   <- your point 1
```

So the guard was unreachable, and `.trim()` really would have re-encoded `%20abc%20` as `abc` and sent a different id upstream. Fixed in ecb9c308: trim dropped, guard dropped, helper reduced to the typing shim you suggested, and the doc comment now says *why* there's no normalization rather than implying validation happened.

Both behaviours are now pinned by tests — one asserting `/kagent/sessions/` reaches `listSessions` and not `getSession` (so the shim's premise can't silently stop holding), and one asserting a padded id arrives at the client as `' abc '` unchanged.

plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -184,6 +205,71 @@ export class KagentClient {
...
+   *
+   * The conversation comes from {@link listSessionTasks}, not from this response's
+   * `events` array — which is ignored entirely, hence the `limit=1` below.
+   */
+  async getSession(
+    sessionId: string,
+    options: KagentRequestOptions,
+  ): Promise<unknown> {
+    return this.request(
+      // `limit=1` because the caller wants the session object and nothing else.
```
marians: **The comment explains everything except the one thing that would prevent a regression: `limit=0` means *unlimited*.**

kagent's DB layer gates the LIMIT'd query on `opts.Limit > 0` (`go/core/internal/database/client_postgres.go`, the `case opts.Limit > 0:` arms), and `eventQueryOptionsFromRequest` leaves `Limit` at its zero value when the param is absent. So `1` is the smallest value that actually limits anything, and "ask for zero events, we don't read them" — the obvious-looking simplification — silently restores the full 591 KB payload.

The URL assertion in `KagentClient.test.ts` does catch that edit, but the failure just says the URL differs; it doesn't say why the value matters. One clause here closes it for good:

```suggestion
      // `limit=1` — not `limit=0`, which kagent reads as *unlimited* (its DB layer
      // gates the LIMIT clause on `opts.Limit > 0`). The caller wants the session
      // object and nothing else.
```

plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -184,6 +205,71 @@ export class KagentClient {
...
+      // kagent bundles the session's stored events into this response and they
+      // dominate it — on a real 4-turn session, 591 KB of events against 261
+      // bytes of session metadata. They are not the conversation (that comes from
+      // `/tasks`) and, despite kagent's Go doc comment, not A2A messages either,
+      // so there is nothing in them we can use.
+      //
+      // Both v0.9.9 and v0.10 parse `limit` with the same handler code. A version
+      // that ignored it would simply return everything, which is exactly today's
+      // behaviour — so this can only help.
```
marians: **"the same handler code" is the one part of this that isn't literally true.**

The semantics are identical — I checked both — but the code isn't shared: v0.9.9 parses `limit` inline inside `HandleGetSession`, while v0.10 (and main) parse it in the extracted `eventQueryOptionsFromRequest` helper. v0.9.9 also answers `400` on an unparseable `limit` from inside the handler rather than from the helper's error return.

That difference changes nothing for us, which is exactly why the sentence should not claim more than it needs to — the point being made is "both honour the param", and that stands on its own:

```suggestion
      // Both v0.9.9 and v0.10 honour `limit` on this endpoint — v0.9.9 parses it
      // inline in `HandleGetSession`, v0.10 in `eventQueryOptionsFromRequest`. A
      // version that ignored it would simply return everything, which is exactly
      // today's behaviour — so this can only help.
```

The changeset repeats the same phrase (`Both v0.9.9 and v0.10 parse \`limit\` with the same handler code`) and is worth the same edit.

.changeset/agent-platform-session-detail-proxy.md
```diff
@@ -0,0 +1,65 @@
...
+
+Two details worth knowing:
+
+`GET /kagent/sessions/:id` asks kagent for `limit=1`, because the caller wants the
+session object and nothing else. kagent bundles the session's stored events into
+that response and they dominate it — on a real 4-turn session, 591 KB of events
+against 261 bytes of session metadata. Nothing reads them (see below), so this
+trims the request by ~99%. Both v0.9.9 and v0.10 parse `limit` with the same
+handler code, and a version that ignored it would simply return everything, which
+is the previous behaviour.
```
marians: **"Two details worth knowing:" is now followed by three, and the new one isn't in the list.**

The `limit=1` paragraph sits between the lead-in and the bullets, so it reads as neither an introduction to them nor a member of them — and the count is off by one. This ships verbatim as the package's changelog entry, so it's worth the 30 seconds.

Folding it in as the first bullet fixes both at once:

```suggestion
Three details worth knowing:

- **`GET /kagent/sessions/:id` asks kagent for `limit=1`.** The caller wants the
  session object and nothing else, but kagent bundles the session's stored events
  into that response and they dominate it — on a real 4-turn session, 591 KB of
  events against 261 bytes of session metadata. Nothing reads them (see below), so
  this trims the response by ~99%. Both v0.9.9 and v0.10 honour `limit` on this
  endpoint, and a version that ignored it would simply return everything, which is
  the previous behaviour.
```

(Unrelated to the ordering, but in the same block: it trims the *response*, not the request — the request is a URL either way.)

plugins/agent-platform-backend/src/KagentClient.ts
marians: This is the best kind of comment — the code was right and the reasoning behind it wasn't written down. Verified and fixed in 647a1f58:

```
go/core/internal/database/client_postgres.go:243:  case opts.OrderAsc && opts.Limit > 0:
go/core/internal/database/client_postgres.go:251:  case opts.Limit > 0:
```

So `limit=0` falls through to the unlimited arm, and the tempting simplification would have quietly undone the whole change. Took your suggested wording at the call site, added the same warning to the README, and — to your point that the assertion catches the edit but not the reason — put it in the test too:

```
// If this assertion fails because someone changed the value: `limit=0` does
// NOT mean "no events". kagent gates its LIMIT clause on `opts.Limit > 0`, so
// zero reads as *unlimited* and restores the full payload. `1` is the
// smallest value that limits anything.
```

plugins/agent-platform-backend/src/KagentClient.ts
marians: You're right and I should have checked before writing it — I'd read v0.9.9's inline parsing earlier in this work (`limit := r.URL.Query().Get("limit")` straight inside `HandleGetSession`) and still wrote "the same handler code". Confirmed v0.10 has the extracted helper:

```
go/core/internal/httpserver/handlers/sessions.go:263:func eventQueryOptionsFromRequest(r *http.Request) (database.QueryOptions, error) {
```

Took your wording verbatim in 647a1f58, and applied the same edit to the changeset, which repeated the phrase. The claim only ever needed to be "both honour the param", so it now says only that.

.changeset/agent-platform-session-detail-proxy.md
marians: Both right — fixed in 647a1f58. Folded the paragraph in as the first bullet with your wording, so the count matches and it stops sitting in no-man's-land between the lead-in and the list. Also fixed request → response; it's the payload coming back that shrinks.

I added one clause to your suggested text, since the other thread on this file established it: the limit has to be `1` rather than `0`, because kagent reads zero as unlimited. Worth having in the changelog entry too, not just the source comment.

PR #2003, opened 2026-07-29
feat(agent-platform): parse kagent session tasks into a timeline

marians:
### What does this PR do?

Turns kagent's A2A task payloads into a renderable timeline. Groundwork for the session detail page — **nothing renders these yet**, so there is no visible change.

> **Stacked on #2002.** Base is `agent-platform-session-detail-proxy` so the diff stays clean; GitHub will retarget to `main` when #2002 merges.

A session is a list of A2A **tasks** (turns) → `history` of **messages** → **parts**. kagent discriminates a part's meaning via prefixed keys in `metadata` rather than by wire type, so the new files split along that seam:

| File | Responsibility |
| --- | --- |
| `lib/kagentMetadata.ts` | the `adk_` / `kagent_` prefix reader |
| `lib/kagentParts.ts` | part predicates (`function_call`, `function_response`, thought, agent-tool) |
| `lib/kagentTimeline.ts` | `buildTimeline(tasks, timestamps)` → flat `TimelineItem[]` + token totals |
| `lib/kagentSessionState.ts` | A2A `TaskState` → label/tone |
| `lib/kagentEventTimestamps.ts` | recovers per-message times from the doubly-encoded events |
| `lib/kagentTaskSchema.ts`, `lib/kagentSessionDetail.ts` | permissive zod + normalizers |

Decisions I'd most like a second opinion on:

**Metadata is read `adk_<key>` first, then `kagent_<key>`.** Not a guess — it's kagent's own interop mechanism (`getMetadataValue`, `ui/src/lib/messageHandlers.ts:479`): upstream ADK writes one prefix, kagent writes the other, and a single session can contain both. There are two fixtures with identical content under different prefixes, and a test asserting they normalize identically.

**A tool call and its result are one item**, matched on the function-call id, with open calls scoped **per task** so a repeatedly-used tool can't get a result attached to the wrong call. An orphan response (call in a task we don't have, or no id) still renders — hiding a result seemed worse than showing one without its request.

**Delegations to other agents are their own kind.** They look like tool calls on the wire (the tell is `__NS__` in the name, same check kagent's UI makes), but a subagent runs in its own session, so its messages never appear here and the response is the only place its token usage surfaces. That usage is counted once toward the session total, keyed on the **call** rather than the response — responses don't always repeat the name, and keying on them would silently under-count.

**An unrecognised approval verdict leaves the verdict unset** rather than defaulting to "approved". The message is still recognised as a decision (so it doesn't render as user prose), but guessing would claim the user consented to an action they may have refused.

**Session state comes from the last task.** kagent returns tasks `ORDER BY created_at ASC` (verified in `tasks.sql.go`), so the array is chronological; an earlier turn having completed says nothing about whether the session is working now. An unknown A2A state renders as itself and counts as *inactive*, so it can't produce a spinner that never resolves. "No tasks at all" stays its own condition rather than being flattened into a state.

**Per-message timestamps** come from joining the session's stored events on `messageId`, because A2A messages carry no time of their own. I've treated that join as **decoration** everywhere: whether the two storage paths agree on ids isn't guaranteed anywhere, so a miss falls back to the task's timestamp and then to showing no time. Tests cover all three levels.

### How does it look like?

Nothing visual yet. The shape it produces:

```
user-message   You: Which GitHub issues are assigned to me?
reasoning      (2 adjacent thought parts, merged into one item)
tool-call      github_search_issues  args + result, isPending: false
agent-message  issue-tracker: You have two open issues...
user-message   You: Ask the SRE agent to check cluster health.
agent-call     kagent__NS__sre_agent  tokens: {total: 3100, ...}
```

Not in this PR — the user-facing write-up in `docs/agent-platform.md` lands with the page, where there's something to describe.

**One thing to flag:** the fixtures here are **hand-authored** from kagent's Go/Python source and its own test payloads, not captured from production. PR 2 adds no UI, so nothing in the app would exercise the new endpoints yet. I'll capture a real gazelle payload and add it as the primary fixture in the UI PR, where the page actually calls them — same approach that caught the `z.unknown().transform()` bug in the sessions list. Until then, treat the parsing as verified against the source but not against the wire.


plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -0,0 +1,479 @@
...
+            }
+            return;
+          }
+
+          // An orphan response: its call was in a task we don't have, or the
+          // payload carried no id. Still render it — a result with no visible
+          // request is odd, but hiding it would be worse.
+          if (isInternalToolName(response.name)) {
+            return;
+          }
```
marians: **The approval's own `function_response` leaks into the timeline as a raw tool call.**

`adk_request_confirmation` is deliberately *not* in `INTERNAL_TOOL_NAMES` (only `adk_request_credential` and `ask_user` are), and the approval branch above (line 312) returns **without** pushing the call into `openCalls`. So when kagent's history contains the `function_response` that answers the long-running confirmation call — which is how ADK resumes a long-running tool — it can never match an open call, falls through to this orphan branch, and renders as a `tool-call` item literally named `adk_request_confirmation` whose `result` is the internal confirmation payload, sitting right after the approval card.

Scenario: the approval fixture plus the response ADK sends on resume →
`['user-message', 'approval', 'tool-call: adk_request_confirmation', 'agent-message']`. The fixture stops before that message, so no test catches it.

Related, same root cause: if a confirmation call ever arrives without the `adk_/kagent_is_long_running` flag (or with it as the string `"true"` — `isKagentMetadataFlagSet` is strict `=== true`), the guard at line 312 misses and the call itself renders as a raw `tool-call` named `adk_request_confirmation` with the `originalFunctionCall` wrapper as its args, instead of an approval card.

```suggestion
          if (
            isInternalToolName(response.name) ||
            response.name === CONFIRMATION_TOOL_NAME
          ) {
            return;
          }
```
(and consider adding `CONFIRMATION_TOOL_NAME` to the `isInternalToolName` check on the call side too, after the long-running branch).

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -0,0 +1,479 @@
...
+        skippedMessages += 1;
+        return;
+      }
+
+      if (message.messageId) {
+        if (seenMessageIds.has(message.messageId)) {
+          return;
+        }
+        seenMessageIds.add(message.messageId);
+      }
```
marians, line 183: **Session-wide dedup + per-task `openCalls` can duplicate a tool call and leave one copy pending forever.**

Dedup is global (`seenMessageIds` lives outside the task loop) but `openCalls` is per task. In the "overlapping history window" case this comment cites, a call message can appear in task N *and* again in task N+1, with only task N+1 carrying the response:

- task N: the call is rendered, registered in task N's `openCalls`, ends `isPending: true` (never resolved — the response isn't in that task);
- task N+1: the repeated call message is deduped here, so it is **not** added to task N+1's `openCalls`; the response then finds no open call, hits the orphan branch (line 279) and pushes a *second* item for the same call.

Result: the same tool call rendered twice — once spinning as pending, once as a result with no request. Scoping `seenMessageIds` per task would trade this for a visible duplicate turn; alternatively, register a deduped call's `itemIndex` in the current task's `openCalls` so a later response can still fold into the existing item.

Flip side of the same asymmetry: messages with no `messageId` are never deduped, so in that overlapping-window case their `usage_metadata` is added to the session total once per repeat.

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -0,0 +1,479 @@
...
+            // we don't recognise. The approval is still resolved — we just can't
+            // say which way, and guessing "approved" would assert consent.
+            if (item?.kind === 'approval' && decision.verdict) {
+              items[openApprovalIndex] = { ...item, verdict: decision.verdict };
+            }
+            openApprovalIndex = undefined;
+          }
+          return;
+        }
+      }
```
marians, line 215: **A decision that arrives in a different task than its approval is silently swallowed.**

`openApprovalIndex` is scoped per task, but the `return` on line 182 is unconditional: any user message carrying a `decision_type` data part is dropped from the timeline whether or not there is an approval to attach it to.

Scenario: kagent records the user's "Approve" as a message in a *new* task (the approval call being in the previous one). The approval item from task N keeps `verdict: undefined` and renders as permanently unanswered, and the decision message itself produces no item at all — so the timeline shows an approval the user apparently never answered, even though they did. The fixture keeps both in one task, so nothing covers this.

Given the code already assumes "only one approval can be open at a time", hoisting `openApprovalIndex` out of the `tasks.forEach` (like `seenMessageIds`) would make the two scopes consistent and cost nothing. At minimum, only swallow the message when an approval is actually open.

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -0,0 +1,479 @@
...
+  let tokens = EMPTY_USAGE;
+  let skippedMessages = 0;
+
+  // kagent's history can repeat a message across tasks (a resend or an
+  // overlapping window), and rendering it twice would read as the agent saying
+  // the same thing twice.
+  const seenMessageIds = new Set<string>();
+
+  tasks.forEach((task, taskIndex) => {
+    const taskTimestamp = task.status?.timestamp;
```
marians: **The task-timestamp fallback bypasses `normalizeTimestamp`.**

Every other kagent timestamp in this plugin goes through `normalizeTimestamp` (session `created_at`/`updated_at`, and each event's `created_at` in `buildEventTimestampIndex`) precisely because kagent serializes non-pointer `time.Time`, so an unset value arrives as Go zero time and browsers render it as "Dec 31, 0000". `task.status.timestamp` is read raw here and becomes `item.at` for *every* item in the task whenever the event join misses — which the PR itself treats as the normal case ("the join is decoration").

Scenario: a task whose `status.timestamp` is `0001-01-01T00:00:00Z` (or any unparseable string kagent emits) and no matching events → the whole turn renders "Dec 31, 0000" instead of the intended "no time at all", and `TimelineItem.at` no longer satisfies its own doc comment ("RFC3339 when known").

Fix: `const taskTimestamp = normalizeTimestamp(task.status?.timestamp);` with `normalizeTimestamp` imported from `./kagentSessions` (as `kagentEventTimestamps.ts` already does).

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -0,0 +1,479 @@
...
+    // The approval awaiting a verdict. Only one can be open at a time — kagent
+    // suspends the task until the user answers.
+    let openApprovalIndex: number | undefined;
+
+    history.forEach((entry, entryIndex) => {
+      const message = parseMessage(entry);
+      if (!message) {
+        skippedMessages += 1;
+        return;
+      }
```
marians, line 164: **`skippedMessages` counts ordinary non-message history entries as data loss.**

`parseMessage` returns `undefined` for two very different things: an entry that isn't parseable at all, and a well-formed entry whose `kind` is simply not `'message'` (`kagentParts.ts:75`). Both land here and increment `skippedMessages`, which `SessionTimeline` documents as "Messages dropped because they could not be parsed at all".

Scenario: a session with artifact/status entries in `history` — your own `tasks.malformed.json` includes `{"kind": "artifact-update", ...}` as a *realistic* entry, so this isn't hypothetical. The upcoming UI showing "N messages could not be read" would then warn about a perfectly healthy session. Distinguishing "not an object" from "not a message kind" (e.g. have `parseMessage` return a discriminated result, or only count the schema failure) keeps the counter meaning what it says.

plugins/agent-platform/src/apis/KagentApiClient.ts
```diff
@@ -105,14 +124,68 @@ export class KagentApiClient implements KagentApi {
...
+  ): Promise<KagentSessionDetail | undefined> {
+    const body = await this.get<unknown>(
+      `/kagent/sessions/${encodeURIComponent(sessionId)}`,
+      installation,
+    );
+    const { detail, drift } = normalizeSessionDetail(body, installation);
+
+    if (drift) {
+      reportDrift(installation, drift);
+    }
```
marians, line 157: **The drift dedup key is now shared by three endpoints, so the new warnings are mostly unreachable.**

`reportedDrift` keys on `${installation}:${drift.kind}` in module state, and all of `listSessions`, `getSessionDetail` and `listSessionTasks` now feed it. `skipped-rows` is emitted by all three (`normalizeSessionDetail` even reuses it for "the response carried no readable session"), so whichever fires first permanently silences the other two for that installation.

Scenario: the sessions list drops one unreadable row on `gazelle` (warning printed, key `gazelle:skipped-rows` burned). The user then opens a session on `gazelle` whose task list drops 30 unreadable tasks — the timeline renders half a conversation and **nothing is logged**. Your own test acknowledges this: it has to use a different installation (`golem`) because "`gazelle:skipped-rows` was already reported by the `listSessions` suite above".

Including the endpoint in the key (`${installation}:${endpoint}:${drift.kind}`) keeps the set bounded while making each read reportable on its own. It would also stop `getSessionDetail`'s "no readable session" — a routine outcome, not drift — from consuming the budget, and would fix the message prefix, which currently says "kagent **sessions** response drift" for a task-list read.

plugins/agent-platform/src/lib/kagentSessionState.ts
```diff
@@ -0,0 +1,90 @@
...
+export function describeSessionState(
+  state: string | undefined,
+): SessionState | undefined {
+  if (!state) {
+    return undefined;
+  }
+  const known = KNOWN_STATES[state.toLowerCase()];
+  if (known) {
+    return { raw: state, ...known };
+  }
```
marians: **Prototype-chain lookup defeats the "an unknown state renders as itself" guarantee.**

`KNOWN_STATES[state.toLowerCase()]` also resolves inherited `Object.prototype` members, so a state of `constructor`, `toString`, `valueOf`, `hasOwnProperty`, … is truthy and takes the "known" branch. `{ raw, ...known }` then spreads a function (no own enumerable properties), yielding `{ raw: 'tostring', label: undefined, tone: undefined, isActive: undefined }` — a badge with no label and an undefined tone, rather than the documented fall-through that renders the state verbatim.

This module's whole contract is tolerating any value kagent puts on the wire, so it's worth closing:

```suggestion
  const key = state.toLowerCase();
  const known = Object.prototype.hasOwnProperty.call(KNOWN_STATES, key)
    ? KNOWN_STATES[key]
    : undefined;
  if (known) {
    return { raw: state, ...known };
  }
  return { raw: state, label: state, tone: 'neutral', isActive: false };
```

marians (commented):
Careful pass over the whole diff (parsing layer, timeline builder, client methods, fixtures). The parsing is genuinely defensive and the reasoning in the comments is easy to follow — the findings below are all in the "wire shape we haven't seen yet" category the PR description already flags, plus one observability regression.

Highest-value ones: the `adk_request_confirmation` response leaking through the orphan branch as a raw tool call, and the drift-dedup key now being shared by three endpoints (which makes the new `listSessionTasks` warning effectively unreachable — the test has to dodge it with a different installation).

Nothing here blocks the groundwork landing; most are cheap to close before the UI PR exercises these paths against a real payload.

One thing I could not verify statically and would check against the real gazelle capture: whether kagent repeats the *same* `usage_metadata` bag on several a2a messages derived from one LLM event. If it does, the message-level accumulation in `buildTimeline` over-counts the session total, and the `messageId` dedup only saves you when the ids actually repeat.

marians:
### Validated against a real payload — three corrections

Your curl let me run the parsers over a live gazelle session (4 tasks, 32 messages, 12 tool calls) before this merges. It produced **29 items, no drift, no skipped rows**, and all 12 tool calls paired with their responses. But it also disproved something I'd built on, so the diff has grown:

**1. Per-message timestamps are not obtainable — that code is deleted.**

kagent's Go type says `Data string // JSON-serialized protocol.Message`, which is what made me think the session's stored events could supply the timestamps A2A messages lack. The comment is stale. The decoded value is an **ADK event**:

```
actions, author, avg_logprobs, branch, cache_metadata, citation_metadata,
content, custom_metadata, error_code, ..., invocation_id, partial,
timestamp, turn_complete, usage_metadata
```

No `messageId` anywhere — 36 events yielded zero joinable ids. `invocation_id` *does* correlate (4/4 against task history), but an invocation is a whole turn, so it's no finer than the task timestamp I already had. `kagentEventTimestamps.ts` is gone, `buildTimeline` lost its second parameter, and every item now takes its task's timestamp. Items within a turn share one **by design** — the UI should show it per turn rather than imply per-message precision.

This also had a knock-on effect worth its own commit on #2002: with events unread, `GET /kagent/sessions/:id` now asks for `limit=1`. On that session the events were **591 KB against 261 bytes** of session metadata I actually needed.

**2. kagent repeats each user message verbatim under the same `messageId`**, on every single turn:

```
t0 #0 user text len=103 id=019f892c-1c1e-75e6-b54f-25ab49e3a1c8
t0 #1 user text len=103 id=019f892c-1c1e-75e6-b54f-25ab49e3a1c8   <- same id
```

So the session-wide dedupe is load-bearing on real data, not defensive as I'd written it. Without it every turn opens by saying the same thing twice. Now has its own test and the fixture reproduces it.

**3. One message can carry prose and several tool calls**, always text first — observed `text(-) -> data(function_call) -> data(function_call)`. Walking parts in order and merging only runs of same-kind text is what keeps the narration ahead of the calls it introduces. Also tested now.

### What the real data does *not* exercise

Worth knowing when reviewing the fixtures: that session had no `thought` parts and no `__NS__` delegations, and every metadata key was `kagent_`-prefixed — no `adk_` in sight. So reasoning, agent delegation and the `adk_` fallback are covered **only** by hand-authored fixtures. They're justified by kagent's source (`part_converter.py` sets the thought key; `getMetadataValue` reads `adk_` first), but they are not wire-verified. Please don't let the "validated against real data" framing above extend further than it should.

### On the fixtures

The real payload is **not** committed. A `/tasks` body is the full conversation — prompts, tool arguments, tool results — which isn't ours to put in this repo. The hand-authored fixtures are now structurally faithful to the three findings above instead.

`tasks.adk-prefixed.json` is now *generated* from `tasks.v0-9-9.json` by swapping only the metadata **key** prefix (agent references like `kagent__NS__sre_agent` are values, not keys, and stay). That way the two can't drift in any other dimension, which is exactly what the prefix test asserts. I got this wrong on the first attempt — a naive `"kagent_` → `"adk_` also rewrote the agent reference.

### One thing to flag for the UI PR

Token totals are large in a way that will need careful labelling: that 4-turn session sums to **1.5M prompt tokens** across 14 model calls (min 3,854, max 144,379). That's not a bug — each call re-sends the whole context and muster's tool schemas are big, so it's genuine cumulative billed usage, and kagent's own UI sums it the same way. But "1.5M input tokens" on a short chat will read as alarming or broken unless the label says what it's counting.

fiunchinho:
### Code review

Found 4 issues:

1. Stale doc comment: `a2aMessageWireSchema` says per-item times "come from joining `messageId` against the session's events — see `kagentEventTimestamps.ts`", but that file was deleted by this PR's second commit and the shipped design does the opposite (events are ignored; every item takes its task's timestamp, per `kagentTimeline.ts` and the changeset).

https://github.com/giantswarm/backstage/blob/232a271678fdac9f0ece04c1e1bff15cfbb5eed7/plugins/agent-platform/src/lib/kagentTaskSchema.ts#L40-L47

2. Stale JSDoc on the `KagentApi` interface: `getSessionDetail` claims to return "the message timestamps recovered from its stored events", and `listSessionTasks` claims timestamp combining "happens in the calling hook". `KagentSessionDetail` is just `{ session; readOnly? }` — it carries no timestamps — and no such hook exists. This describes the abandoned event-timestamp join.

https://github.com/giantswarm/backstage/blob/232a271678fdac9f0ece04c1e1bff15cfbb5eed7/plugins/agent-platform/src/apis/types.ts#L37-L58

3. Misleading field comment: `verdict` is documented as "Undefined while the request is still unanswered", but `readDecision` also leaves it undefined for an *answered* approval with unrecognised wording (tests "leaves the verdict unset for wording it does not recognise" and "leaves an unanswered approval pending" both assert `verdict: undefined`). A consumer relying on "undefined = pending" would mislabel answered approvals.

https://github.com/giantswarm/backstage/blob/232a271678fdac9f0ece04c1e1bff15cfbb5eed7/plugins/agent-platform/src/lib/kagentTimeline.ts#L64-L67

4. The changeset body is ~70 lines of design rationale ("Decisions worth knowing", validation history, internal file cross-references) that Changesets will emit verbatim into the published CHANGELOG (CLAUDE.md guidance: "`CHANGELOG.md` entries should say what the consumer gets, not the rationale that usually already lives in the PR body"). Consider keeping the one-line consumer summary and moving the narrative to the PR body.

https://github.com/giantswarm/backstage/blob/232a271678fdac9f0ece04c1e1bff15cfbb5eed7/.changeset/agent-platform-session-timeline.md#L1-L20

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Both halves right, fixed in b8114528 — and kagent's own code corroborates the first one. `extractToolCallRequests` skips **three** names, not the two I put in `INTERNAL_TOOL_NAMES`:

```
ui/src/lib/toolCallExtraction.ts:63:          data.name === "adk_request_confirmation" ||
```

So kagent filters `adk_request_confirmation` out of tool calls for exactly this reason. Took your suggestion for the response side.

On the call side I went further than "add it to `isInternalToolName`": approvals are now discriminated on the **name alone**, dropping the `is_long_running` requirement entirely. Your second point convinced me the flag is a liability rather than a safeguard — it fails open in the ugliest available direction, turning an approval into a raw tool call that exposes ADK's `originalFunctionCall` wrapper as user-visible arguments. The name is sufficient and can't degrade like that. `isLongRunningPart` is deleted, with a comment where it was explaining why the flag isn't consulted.

The fixture stopped before the resume response, as you spotted. It now includes it, and there's a test asserting no item with `confirmation` in its tool name survives.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Confirmed by tracing it, fixed in b8114528.

I took the second option you offered rather than scoping the dedupe per task — the intra-task duplicate is the one I've actually *observed* (kagent repeats the user message under the same id on every turn), so narrowing the dedupe would reintroduce the visible-duplicate bug I'd just fixed. Instead a message records the calls it produced, and a deduped repeat re-arms them in the current task's `openCalls`, so a later response folds into the existing item. Test covers the exact sequence: call in task N, call+response in task N+1, expecting one item, not pending.

Worth noting the cross-task repeat is still **unobserved** — the real gazelle payload only duplicated *within* a task. My dedupe comment asserted cross-task repeats happen, which was speculation dressed as fact; I've rewritten it to say "reliably for the user's message on every turn, and in principle across tasks".

On the flip side you mention: messages with no `messageId` can't be deduped, so a repeat would double-count usage. That's inherent — there's no key to dedupe on — and the real payload had ids on all 32 messages. Leaving it, since the alternative would be content hashing.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Right, and it's the worst-feeling of the seven: the user approves, the timeline says they never answered. Fixed in b8114528 by hoisting `openApprovalIndex` out of `tasks.forEach`, as you suggested — the code already assumed only one approval can be open at a time, so the two scopes were inconsistent for no reason.

Test added that splits the approval fixture across two tasks (call in task 1, decision in task 2) and asserts the verdict still lands.

I kept the `return` unconditional rather than only swallowing when an approval is open. With session scope the bug is gone, and a stray decision message with no approval produces no item anyway — its only part is the data part, so there's no text to render. Happy to tighten it if you'd rather the guard be explicit.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Correct, fixed in b8114528 exactly as suggested.

One thing worth flagging: you reviewed the commit before my last push, which deleted `buildEventTimestampIndex` and the whole event join — the real gazelle payload showed events are ADK events with no `messageId`, so there was nothing to join. That makes this finding **more** serious, not less: the task timestamp is no longer the fallback, it's the *only* source, so a Go zero time would render "Dec 31, 0000" across every item of the turn rather than just when a join missed.

Test asserts a task with `status.timestamp: '0001-01-01T00:00:00Z'` yields `at === undefined` on every item.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Agreed, and your point about my own fixture landed — I'd put `{"kind": "artifact-update"}` in there *as a realistic entry* and then counted it as data loss.

Fixed in b8114528: `parseMessage` became `parseHistoryEntry`, returning a discriminated `{kind: 'message' | 'other' | 'unparseable'}`. Only `unparseable` increments the counter, and `SessionTimeline.skippedMessages` now documents that exclusion explicitly, since the upcoming UI will surface it as "N messages could not be read".

Test asserts the malformed fixture yields exactly 2 — `null` and the bare string — with the `artifact-update` not counted. Writing that test caught me expecting 3, because I'd conflated it with the 3 skipped *tasks* at the envelope level. Two different counters, and the fixture exercises both.

plugins/agent-platform/src/apis/KagentApiClient.ts
marians: This is the one I should have caught myself — my own test had to switch to `golem` and I wrote a comment explaining why instead of treating it as the bug it was signalling. Fixed in b8114528.

Key is now `${installation}:${source}:${drift.kind}`, with `source` one of `sessions` / `session` / `session tasks`, which also fixes the message prefix you spotted (a task-list read was announcing itself as "kagent **sessions** response drift"). The test that needed a different installation now uses `gazelle` again, and there's a new one that drives the exact scenario: drop a row from the sessions list on an installation, then drop tasks on the *same* installation, asserting both warn.

Also took the `getSessionDetail` half: "the response carried no readable session" no longer borrows `skipped-rows`. It has its own `missing-payload` kind, since `skipped-rows`' message counts dropped rows and reusing it reported "no session" as though rows had been lost.

plugins/agent-platform/src/lib/kagentSessionState.ts
marians: Valid and fixed in b8114528 — but the examples don't all hold, and the difference matters for the test, so I checked rather than pasting them in:

```
Object.prototype members:        constructor, __defineGetter__, __defineSetter__,
                                 hasOwnProperty, __lookupGetter__, __lookupSetter__,
                                 isPrototypeOf, propertyIsEnumerable, toString,
                                 valueOf, __proto__, toLocaleString
lowercase-stable (can collide):  constructor, __proto__

KNOWN_STATES["tostring"]        -> undefined
KNOWN_STATES["valueof"]         -> undefined
KNOWN_STATES["hasownproperty"]  -> undefined
KNOWN_STATES["constructor"]     -> function   <-- collides
KNOWN_STATES["__proto__"]       -> object     <-- collides
```

`toLowerCase()` is what saves `toString` / `valueOf` / `hasOwnProperty` — they lower-case to strings that aren't prototype members. Only `constructor` and `__proto__` survive unchanged, and both do break exactly as you describe (spreading a function or the prototype yields no own enumerable properties → undefined label and tone).

Took your suggestion verbatim, and the test is parameterised over just those two so it documents the real surface rather than implying a wider hole. Agreed on the reasoning regardless of how narrow it is — this module's entire contract is tolerating arbitrary wire values, and the guard is one line.

marians:
### Correction: real data *does* use `adk_` prefixes, and the fallback is load-bearing

I said above that the gazelle session I probed had "only `kagent_`-prefixed metadata — no `adk_` in sight", and that the `adk_` fallback was justified by kagent's source but **not wire-verified**. That was wrong, and it matters more than a footnote.

Opening a *different* gazelle session in the running app today, every message carries:

```json
"adk_usage_metadata": { "candidatesTokenCount": 60, "promptTokenCount": 2909 }
```

The session I probed yesterday carried `kagent_usage_metadata`. So **one installation genuinely serves both prefixes**, presumably depending on which kagent/ADK version wrote each message. The dual-prefix reader is therefore load-bearing, not defensive: reading only `kagent_` would silently zero this session's token totals.

Please read my earlier "what the real data does *not* exercise" paragraph with that removed. Reasoning (`thought`) parts and `__NS__` delegations are still fixture-only.

### And a real bug it surfaced

The same session rendered **"Total 0" next to "Input tokens 1.4M"**. `adk_usage_metadata` has no `totalTokenCount` at all, so summing reported totals stayed at zero while the parts ran into millions. kagent's own UI has the identical hole (`total: usage.totalTokenCount ?? 0`) — I faithfully reproduced a kagent bug.

Fixed in c9b274ec on the UI branch (#2005): the total is derived from `prompt + completion` when kagent reports none, while a *reported* total still wins when present — a model billing thinking tokens separately counts them in the total but in neither part, so it isn't always the sum. Both cases have tests.

Verified live afterwards: the strip now reads `Turns 3 · Input 1.4M · Output 2.6k · Total 1.4M`.

PR #2004, opened 2026-07-29
feat(flux-react): disable Suspend when spec.suspend is GitOps-managed

marians:
### What does this PR do?

Follow-up to #2001, from [Flux-team feedback](https://gigantic.slack.com/archives/C02GDJJ68Q1/p1785243811567809) about manipulating `.spec` from a UI. Three changes.

**1. Detect a contested `spec.suspend` and disable the toggle.**

The concern raised was that a manual suspend on a GitOps-managed resource gets overwritten. That is right — but the precise condition is narrower than "the resource is GitOps-managed". Server-side apply is **field-level**: kustomize-controller only touches fields present in the manifest it applies, and removes a field only if it previously owned it and has since dropped it. So what matters is whether **the applied manifest asserts `spec.suspend`**.

When it does, the applying controller force-takes the field back on its next apply — `ssa/manager_apply.go` in fluxcd/pkg always passes `client.ForceOwnership` — and the suspension undoes itself within one of *that* controller's intervals, not the resource's own. Silently: no conflict error, nothing degraded.

We can see this exactly, without guessing, from `metadata.managedFields` (already in our GET responses): any entry with operation `Apply` that owns `f:spec.f:suspend`. When present, the Suspend/Resume toggle is disabled with a tooltip naming the owning manager(s).

Two details worth reviewing:

- **Not restricted to `kustomize-controller`.** A human `kubectl apply --server-side`, or helm-controller's drift correction when `spec.driftDetection` is enabled, has the same effect. Any SSA applier of the field will revert us, so the check is generic.
- **Detection covers SSA appliers only**, and is a best-effort signal. `Update`-operation owners are skipped, which correctly excludes our own merge patches (so our writes don't lock the button out) — but also misses the two common *non-SSA* declarative writers, which do keep a stored desired state and re-assert it: client-side `kubectl apply` (`kubectl-client-side-apply`) and a plain `helm upgrade`, whose three-way merge resets drift on chart-declared fields. So a chart shipping a Kustomization with `spec.suspend` declared still shows an enabled toggle whose change the next upgrade reverts. This is documented on `getApplyFieldOwners` rather than papered over by matching manager names, which would be a guess.
- **Stale ownership is handled.** A `managedFields` entry is only rewritten by a write, so it outlives the applier — see the escape-hatch note below.

Disabling is precisely correct here rather than merely conservative: the toggle always flips *away* from the current value, and the current value is what the apply-owner last asserted — so whenever the field is apply-owned, the action on offer is exactly the one that would be reverted.

**2. Reconcile is deliberately left enabled**, including on managed resources.

`reconcile.fluxcd.io/requestedAt` is never part of an applied manifest, so no apply-owner asserts or prunes it. kustomize-controller's `Cleanup.Annotations` list is explicit and short (`last-applied-configuration`, two legacy `fluxcd.io` checksums) and does not include it. The controller then copies the value into `status.lastHandledReconcileAt` and *leaves the annotation in place* — that pair is what our pending-request check compares. There is no race to lose, so gating Reconcile would be wrong. There's a comment saying so, to stop a future reader "fixing" it.

**3. An explicit field manager on our writes.**

Writes now set `?fieldManager=giantswarm-backstage`. Previously the apiserver derived the manager name from the request's User-Agent, which for a write proxied through the Backstage backend is both unpredictable and useless in an audit trail.

The suggestion in the thread was to *masquerade as `flux`* so the controller would treat our writes like CLI writes. I checked the source, and there is nothing to imitate:

- flux2 never sets a field manager at all — `internal/utils/utils.go` builds the client as `client.NewWithWatch(cfg, client.Options{Scheme: scheme})`. The `manager: flux` visible in `--show-managed-fields` is just the apiserver deriving it from client-go's default User-Agent.
- kustomize-controller's disallowed-manager list (`kustomization_controller.go`, feeding `ApplyOptions.Cleanup.FieldManagers`) covers `kubectl`/Apply, `kubectl`/Update, `before-first-apply`/Update and its own name with Update, plus any `--override-manager` values. `flux` is absent, so a CLI suspend gets no special treatment in either direction.

So impersonation would buy nothing and destroy attribution. An honest name also gives operators a value they can pass to a controller's `--override-manager` if they want these changes force-reverted.

**4. Corrects the ImagePolicy justification** from #2001.

The comment claimed Flux supports neither operation for `ImagePolicy`. That was wrong: both `crds.fluxcd.v1.ImagePolicy` and `v1beta2` declare `spec.suspend` and `status.lastHandledReconcileAt`, and the CLI covers the kind too — `cmd/flux/reconcile_image_policy.go`, `suspend_image_policy.go`, `resume_image_policy.go` all exist. Including the kind was right; the reason given was not. (The claim also appeared in #2001's description and changeset, which are already merged — noted here for the record.)

On a resource whose `spec.suspend` comes from Git, the Suspend/Resume button is now visibly disabled and says why, instead of appearing to work and then quietly undoing itself minutes later. This is the Adidas case from the thread, where Kustomizations in the `default` namespace set `spec.suspend` explicitly.

Everything else is unchanged: Reconcile works everywhere it did before, and Suspend/Resume still works on resources whose suspend field is not declaratively managed.

Accepted trade-off: you can no longer take a deliberate short-lived suspend on a managed resource from the UI, and if such a resource was suspended elsewhere you cannot resume it here. Both were already true of `flux suspend`/`resume` in the sense that the change wouldn't stick; the difference is that now we say so up front rather than letting it fail silently. The fix in that situation is to change it in Git.

### How does it look like?

```
unmanaged spec.suspend:
  [ ⟳ Reconcile ]  [ ⏸ Suspend ]

spec.suspend applied by kustomize-controller:
  [ ⟳ Reconcile ]  [ ⏸ Suspend ]
                    (disabled)
   tooltip: "spec.suspend is applied by kustomize-controller,
             so a change made here would be reverted on the
             next reconciliation. Change it in Git instead."
```

The Slack thread asked whether the Flux controllers handle a CLI-set `spec.suspend` specially — e.g. keying off the field manager. They don't; see point 3 above. The behaviour difference between "suspend sticks" and "suspend evaporates" is entirely about what's in the applied manifest.

Three per-object annotations exclude a resource from apply and therefore make manual changes durable — `kustomize.toolkit.fluxcd.io/reconcile: disabled`, `kustomize.toolkit.fluxcd.io/ssa: Ignore`, and `kustomize.toolkit.fluxcd.io/ssa: IfNotPresent` (from `applyOpts` in `kustomization_controller.go`).

An earlier revision of this PR claimed these "will correctly show an enabled toggle, since no apply-owner claims the field". That was wrong, as [pointed out in review](https://github.com/giantswarm/backstage/pull/2004#discussion_r3672300856): a `managedFields` entry is only ever rewritten by a write, so it survives the applier walking away. An object handed over for manual control keeps its stale `Apply` entry, and an `ssa: IfNotPresent` object carries one from creation onwards despite never being applied again — so both would have shown a permanently disabled toggle. Those three annotations now short-circuit the check.

Verify on a real MC with:

```bash
kubectl -n default get ks pcn01 -o yaml --show-managed-fields | \
  yq '.metadata.managedFields[] | select(.manager == "kustomize-controller")'
```

No.

### DIFF
--- a/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
+++ b/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
@@ -24,6 +24,12 @@ function describe(resource: FluxObject): string {
   return `${resource.getKind()} ${namespace ? `${namespace}/` : ''}${name}`;
 }
 
+function buildManagedSuspendHint(owners: string[]): string {
+  const by = owners.length === 1 ? owners[0] : owners.join(' and ');
+
+  return `spec.suspend is applied by ${by}, so a change made here would be reverted on the next reconciliation. Change it in Git instead.`;
+}
+
 const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
   const alertApi = useApi(alertApiRef);
   const {
@@ -50,6 +56,16 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
   } else if (isReconcileRequestPending) {
     reconcileHint = REQUEST_PENDING_HINT;
   }
+
+  // Reconcile is deliberately *not* gated on declarative management: the
+  // `reconcile.fluxcd.io/requestedAt` annotation is never part of an applied
+  // manifest, so no apply-owner ever asserts or prunes it. Only `spec.suspend`
+  // can be contested.
+  const suspendFieldApplyOwners = resource.getSuspendFieldApplyOwners();
+  const isSuspendFieldManaged = suspendFieldApplyOwners.length > 0;
+  const suspendHint = isSuspendFieldManaged
+    ? buildManagedSuspendHint(suspendFieldApplyOwners)
+    : '';
 
   const notifyFailure = (error: unknown, action: string) => {
     const forbidden = error instanceof Error && error.name === 'ForbiddenError';
@@ -127,21 +143,26 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
             </Button>
           </span>
         </Tooltip>
-        <Button
-          variant="secondary"
-          size="small"
-          iconStart={
-            isSuspended ? (
-              <PlayArrowIcon fontSize="small" />
-            ) : (
-              <PauseIcon fontSize="small" />
-            )
-          }
-          isPending={isSettingSuspended}
-          onPress={handleToggleSuspended}
-        >
-          {isSuspended ? 'Resume' : 'Suspend'}
-        </Button>
+        <Tooltip title={suspendHint}>
+          <span>
+            <Button
+              variant="secondary"
+              size="small"
+              iconStart={
+                isSuspended ? (
+                  <PlayArrowIcon fontSize="small" />
+                ) : (
+                  <PauseIcon fontSize="small" />
+                )
+              }
+              isDisabled={isSuspendFieldManaged}
+              isPending={isSettingSuspended}
+              onPress={handleToggleSuspended}
+            >
+              {isSuspended ? 'Resume' : 'Suspend'}
+            </Button>
+          </span>
+        </Tooltip>
       </Flex>
     </Box>
   );
--- a/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
+++ b/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
@@ -32,10 +32,27 @@ function joinWithAnd(items: string[]): string {
   return `${items.slice(0, -1).join(', ')} and ${items[items.length - 1]}`;
 }
 
+/**
+ * Deliberately says "at the source that applies it" rather than "in Git": the
+ * ownership check is not Git-specific, so the applier may be a chart rendered by
+ * helm-controller or a human's `kubectl apply --server-side` just as easily as a
+ * Kustomization built from a repository.
+ */
 function buildManagedSuspendHint(owners: string[]): string {
   return `spec.suspend is applied by ${joinWithAnd(
     owners,
-  )}, so a change made here would be reverted on the next reconciliation. Change it in Git instead.`;
+  )}, so a change made here would be reverted on the next reconciliation. Change it at the source that applies it.`;
+}
+
+/**
+ * Reconcile's hint when the resource is suspended *and* its suspension is
+ * declaratively managed. Plain {@link SUSPENDED_HINT} would say "Resume it first",
+ * which is a dead end here — we have just disabled Resume for the same reason.
+ */
+function buildManagedSuspendedReconcileHint(owners: string[]): string {
+  return `Suspended resources are not reconciled, and spec.suspend is applied by ${joinWithAnd(
+    owners,
+  )} — resume it at the source that applies it.`;
 }
 
 /**
@@ -69,19 +86,23 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
 
   const isReconcileDisabled = isSuspended || isReconcileRequestPending;
 
-  let reconcileHint = '';
-  if (isSuspended) {
-    reconcileHint = SUSPENDED_HINT;
-  } else if (isReconcileRequestPending) {
-    reconcileHint = REQUEST_PENDING_HINT;
-  }
-
   // Reconcile is deliberately *not* gated on declarative management: the
   // `reconcile.fluxcd.io/requestedAt` annotation is never part of an applied
   // manifest, so no apply-owner ever asserts or prunes it. Only `spec.suspend`
   // can be contested.
   const suspendFieldApplyOwners = resource.getSuspendFieldApplyOwners();
   const isSuspendFieldManaged = suspendFieldApplyOwners.length > 0;
+
+  let reconcileHint = '';
+  if (isSuspended && isSuspendFieldManaged) {
+    // Must precede the plain suspended case: telling someone to resume first
+    // would be a dead end when Resume is disabled for being managed.
+    reconcileHint = buildManagedSuspendedReconcileHint(suspendFieldApplyOwners);
+  } else if (isSuspended) {
+    reconcileHint = SUSPENDED_HINT;
+  } else if (isReconcileRequestPending) {
+    reconcileHint = REQUEST_PENDING_HINT;
+  }
   const suspendHint = isSuspendFieldManaged
     ? buildManagedSuspendHint(suspendFieldApplyOwners)
     : '';
--- a/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
+++ b/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
@@ -24,11 +24,28 @@ function describe(resource: FluxObject): string {
   return `${resource.getKind()} ${namespace ? `${namespace}/` : ''}${name}`;
 }
 
-function buildManagedSuspendHint(owners: string[]): string {
-  const by = owners.length === 1 ? owners[0] : owners.join(' and ');
+function joinWithAnd(items: string[]): string {
+  if (items.length < 2) {
+    return items.join('');
+  }
 
-  return `spec.suspend is applied by ${by}, so a change made here would be reverted on the next reconciliation. Change it in Git instead.`;
+  return `${items.slice(0, -1).join(', ')} and ${items[items.length - 1]}`;
 }
+
+function buildManagedSuspendHint(owners: string[]): string {
+  return `spec.suspend is applied by ${joinWithAnd(
+    owners,
+  )}, so a change made here would be reverted on the next reconciliation. Change it in Git instead.`;
+}
+
+/**
+ * A disabled react-aria button still receives pointer events — bui's disabled
+ * styling only sets `cursor: not-allowed` — so the native `<button>` swallows the
+ * hover and the tooltip on the wrapping span never opens. MUI's own
+ * wrap-in-a-span recipe works only because its `ButtonBase` sets
+ * `pointer-events: none` when disabled; this restores that half of it.
+ */
+const DISABLED_BUTTON_STYLE = { pointerEvents: 'none' } as const;
 
 const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
   const alertApi = useApi(alertApiRef);
@@ -49,6 +66,8 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
   // the resource's own annotation-vs-status comparison covers the gap after it,
   // and survives a reload or a request someone else made.
   const isReconcileRequestPending = resource.isReconcileRequestPending();
+
+  const isReconcileDisabled = isSuspended || isReconcileRequestPending;
 
   let reconcileHint = '';
   if (isSuspended) {
@@ -135,7 +154,8 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
               variant="secondary"
               size="small"
               iconStart={<RefreshIcon fontSize="small" />}
-              isDisabled={isSuspended || isReconcileRequestPending}
+              isDisabled={isReconcileDisabled}
+              style={isReconcileDisabled ? DISABLED_BUTTON_STYLE : undefined}
               isPending={isRequestingReconciliation}
               onPress={handleReconcile}
             >
@@ -156,6 +176,7 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
                 )
               }
               isDisabled={isSuspendFieldManaged}
+              style={isSuspendFieldManaged ? DISABLED_BUTTON_STYLE : undefined}
               isPending={isSettingSuspended}
               onPress={handleToggleSuspended}
             >
@@ -173,8 +194,8 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
  * them, shown only to users whose cluster RBAC allows the write.
  */
 export const FluxResourceActions = ({ resource }: { resource: KubeObject }) => {
-  // Guard before the inner component mounts so unsupported kinds (e.g.
-  // ImagePolicy) issue no access review at all.
+  // Guard before the inner component mounts so unsupported kinds (e.g. a
+  // ConfigMap, or any non-Flux object) issue no access review at all.
   if (!supportsFluxActions(resource)) {
     return null;
   }
--- a/plugins/flux-react/src/components/FluxOverview/ResourceCard/utils/fluxActions.ts
+++ b/plugins/flux-react/src/components/FluxOverview/ResourceCard/utils/fluxActions.ts
@@ -13,10 +13,12 @@ import {
  * resource class that lacks the fields does not silently get buttons that
  * cannot work.
  *
- * Note this is deliberately *not* the same set as the `flux` CLI's reconcile
- * subcommands: there is no `flux reconcile image policy`, but
- * image-reflector-controller does honour the annotation and `spec.suspend` for
- * ImagePolicy, so the UI can offer both.
+ * `ImagePolicy` belongs here like the rest: `crds.fluxcd.v1.ImagePolicy` and
+ * `v1beta2` both declare `spec.suspend` and `status.lastHandledReconcileAt`, and
+ * the CLI covers it too (`flux reconcile|suspend|resume image policy`, see
+ * `cmd/flux/reconcile_image_policy.go` and siblings in fluxcd/flux2). It was
+ * originally left out of this list on the mistaken grounds that Flux supported
+ * neither operation for the kind.
  */
 const ACTIONABLE_KINDS = [
   'Kustomization',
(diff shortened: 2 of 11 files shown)


plugins/kubernetes-react/src/lib/k8s/KubeObject.ts
```diff
@@ -137,6 +168,31 @@ export class KubeObject<T extends KubeObjectInterface = any> {
...
+    return this.jsonData.metadata.managedFields;
+  }
+
+  /**
+   * The managers that *server-side apply* the given field, and will therefore
+   * re-assert it on their next apply — reverting any imperative write we make.
+   *
+   * Only `Apply`-operation entries count. A manager that reached the field with
+   * an imperative `Update` (a PUT or a non-apply PATCH, which is what our own
+   * writes are) holds ownership but has no declared desired state to restore, so
```
marians: The premise that `Update` owners "never revert anything" is too strong, and it produces false negatives for the two most common *non-SSA* declarative writers — both of which keep a stored desired state and re-assert it:

- **client-side `kubectl apply`** (`manager: kubectl-client-side-apply`, `operation: Update`) sends the whole manifest as a strategic-merge patch on every run, so any manifest-declared field is reset to the manifest value. That is exactly why `kubectl apply` undoes a `kubectl edit`.
- **Helm** — `helm upgrade` (and helm-controller's release writes) patch via `CreateThreeWayMergePatch(..., overwrite=true)`, also recorded as `operation: Update`, and drift on a chart-declared field is reverted on the next upgrade. Only helm-controller's *drift-correction* path goes through `fluxcd/pkg/ssa`, so the changeset's "a Flux object deployed by a HelmRelease is applied by `helm-controller`" holds only when `spec.driftDetection` is enabled.

Concrete scenario: an app chart that ships a `Kustomization`/`HelmRelease` with `spec.suspend` declared. The card shows an enabled Suspend, the user clicks it, and the next `helm upgrade` (or the next `kubectl apply -f`) quietly undoes it — the precise failure mode this PR sets out to eliminate, with no warning.

Either narrow the claim in this comment (and in the changeset) to "we only detect SSA appliers", or additionally treat an `Update` owner as declarative when `metadata.annotations["kubectl.kubernetes.io/last-applied-configuration"]` asserts the path (and/or when the manager is `helm`/`helm-controller`).

plugins/kubernetes-react/src/lib/k8s/FluxObject.ts
```diff
@@ -63,6 +63,37 @@ export class FluxObject<
...
+   * Note the button a suspend toggle offers always flips away from the current
+   * value, and the current value is what the apply-owner last asserted — so
+   * whenever this is non-empty, the offered action is precisely the one that
+   * would be reverted.
+   *
+   * Deliberately not restricted to `kustomize-controller`: a HelmRelease-deployed
+   * Flux object is applied by `helm-controller`, and a `kubectl apply
+   * --server-side` by a human has the same effect. Any apply-owner will revert us.
+   */
+  getSuspendFieldApplyOwners(): string[] {
```
marians, line 111: `managedFields` entries are only ever rewritten by a write, so an `Apply` entry survives indefinitely after the applier stops applying the object. That makes this a false positive for exactly the escape hatches the PR description says are handled:

- Take a Kustomization that kustomize-controller has been applying with `spec.suspend` declared (so it owns `f:spec.f:suspend`), then annotate it `kustomize.toolkit.fluxcd.io/reconcile: disabled` or `kustomize.toolkit.fluxcd.io/ssa: Ignore` — the normal way to hand an object over for manual control. The controller stops touching it, its stale `Apply` entry still lists the field, and the toggle is disabled forever even though a suspend from here would now stick.
- `ssa: IfNotPresent` is worse: the controller *creates* the object via SSA and then never applies again, so the entry is present from the start and the toggle is disabled for the object's whole life.

The PR description states these "will correctly show an enabled toggle, since no apply-owner claims the field", which does not hold. Cheap fix: short-circuit `getSuspendFieldApplyOwners()` to `[]` when the object carries `kustomize.toolkit.fluxcd.io/reconcile: disabled` or `kustomize.toolkit.fluxcd.io/ssa: Ignore|IfNotPresent`.

plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
```diff
@@ -24,6 +24,12 @@ function describe(resource: FluxObject): string {
   return `${resource.getKind()} ${namespace ? `${namespace}/` : ''}${name}`;
 }
 
+function buildManagedSuspendHint(owners: string[]): string {
+  const by = owners.length === 1 ? owners[0] : owners.join(' and ');
```
marians: Two nits on the owner list, both reachable because SSA co-ownership is real (two appliers that assert the *same* value both keep ownership, so `kubectl apply --server-side` alongside kustomize-controller genuinely yields two entries — which is what your own multi-owner test models):

- Three or more owners render as "a and b and c". An Oxford-style join (`a, b and c`) reads better.
- The list isn't deduplicated. Entries are keyed by manager + operation + apiVersion + subresource, so the same manager can appear twice (e.g. `kustomize-controller` at `.../v1beta2` and `.../v1` after a CRD version migration, when the applied value didn't change and so ownership wasn't taken from the old entry), giving "applied by kustomize-controller and kustomize-controller". A `[...new Set(owners)]` in `getApplyFieldOwners` covers it.

plugins/flux-react/src/components/FluxOverview/ResourceCard/utils/fluxActions.ts
```diff
@@ -13,10 +13,12 @@ import {
...
  *
- * Note this is deliberately *not* the same set as the `flux` CLI's reconcile
- * subcommands: there is no `flux reconcile image policy`, but
- * image-reflector-controller does honour the annotation and `spec.suspend` for
- * ImagePolicy, so the UI can offer both.
+ * `ImagePolicy` belongs here like the rest: `crds.fluxcd.v1.ImagePolicy` and
+ * `v1beta2` both declare `spec.suspend` and `status.lastHandledReconcileAt`, and
+ * the CLI covers it too (`flux reconcile|suspend|resume image policy`, see
+ * `cmd/flux/reconcile_image_policy.go` and siblings in fluxcd/flux2). It was
+ * originally left out of this list on the mistaken grounds that Flux supported
```
marians: "It was originally left out of this list" describes something that never happened — `ImagePolicy` has been in `ACTIONABLE_KINDS` since #2001, and the old comment used the (wrong) CLI claim to justify *including* it. A future reader will go looking for a removal that isn't in the history. Suggest phrasing it as "the reason previously given here was wrong" instead.

While in this area: `FluxResourceActions.tsx:176` still says `// Guard before the inner component mounts so unsupported kinds (e.g. ImagePolicy) issue no access review at all` — with `ImagePolicy` in this list that example is exactly backwards (and the existing test at `FluxResourceActions.test.tsx:244` patches an imagepolicy). Since this PR is correcting the `ImagePolicy` narrative, worth fixing that comment too, e.g. to `(e.g. ConfigMap)`.

plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
```diff
@@ -127,21 +143,26 @@ const FluxResourceActionsContent = ({ resource }: { resource: FluxObject }) => {
...
-            ) : (
-              <PauseIcon fontSize="small" />
-            )
-          }
-          isPending={isSettingSuspended}
-          onPress={handleToggleSuspended}
-        >
-          {isSuspended ? 'Resume' : 'Suspend'}
-        </Button>
+        <Tooltip title={suspendHint}>
```
marians, line 187: The span wrapper alone doesn't get you the styled tooltip here. MUI's own "wrap a disabled button in a span" recipe works because MUI's `ButtonBase` sets `pointer-events: none` when disabled (`@material-ui/core/ButtonBase/ButtonBase.js:71`), which is what lets the hover land on the span instead. bui's Button only sets `cursor: not-allowed` for `[data-disabled='true']` — no `pointer-events: none` — so the disabled native `<button>` absorbs the pointer and dispatches no `mouseover`, and MUI's `onMouseOver` on the span never fires.

What the user actually gets is the browser's native `title` tooltip that MUI renders on the span while closed: ~1s delay, unstyled, and no `aria-describedby`. The new test only asserts the `title` attribute, so it passes either way — it doesn't cover the tooltip appearing.

Adding `style={{ pointerEvents: 'none' }}` to the disabled Button (or rendering it `aria-disabled` and blocking `onPress` instead of `isDisabled`) restores the real tooltip. Same applies to the pre-existing Reconcile button above.

fiunchinho:
### Code review

Found 2 issues:

1. Contradictory tooltips for a suspended, GitOps-managed resource. When `isSuspended` and `isSuspendFieldManaged` are both true (the exact state exercised by the test "disables Resume too, for an already-suspended managed resource"), the Reconcile button is disabled with the hint "Suspended resources are not reconciled. Resume it first." — but this PR also disables the Resume button, so the advice is a dead end. `reconcileHint` should account for `isSuspendFieldManaged` and point to Git instead.

https://github.com/giantswarm/backstage/blob/ab2e264e6278d5029405e11f353580e2a66af46c/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx#L52-L58

https://github.com/giantswarm/backstage/blob/ab2e264e6278d5029405e11f353580e2a66af46c/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx#L157-L159

2. The tooltip always says "Change it in Git instead", but the ownership detection is documented as deliberately not Git-specific — `getSuspendFieldApplyOwners()`'s doc comment says a `helm-controller` apply or a human's `kubectl apply --server-side` counts too. For those owners the "in Git" advice is wrong; consider neutral wording like "Change it at the source that applies it."

https://github.com/giantswarm/backstage/blob/ab2e264e6278d5029405e11f353580e2a66af46c/plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx#L27-L31

https://github.com/giantswarm/backstage/blob/ab2e264e6278d5029405e11f353580e2a66af46c/plugins/kubernetes-react/src/lib/k8s/FluxObject.ts#L80-L84

plugins/kubernetes-react/src/lib/k8s/FluxObject.ts
marians: Confirmed and fixed in 2d6d27f4 — this was the sharpest of the five, and it contradicted the PR description rather than just the code.

`getSuspendFieldApplyOwners()` now returns `[]` when the object carries any of the three apply opt-outs. I checked the constant values against `api/v1/kustomization_types.go` rather than trusting my memory of the docs: `DisabledValue = "disabled"`, `IgnoreValue = "Ignore"`, `IfNotPresentValue = "IfNotPresent"` (note the capitals — and `MergeValue` is `"Merge"`, not `"merge"`, which I had wrong in an earlier draft).

Your `IfNotPresent` point is the one that would have hurt most: the entry is there from creation, so the toggle would have been dead for the object's entire life with nothing in the UI hinting why. Table-driven tests cover all three, plus two negative cases (`ssa: Merge` and `reconcile: enabled` must not short-circuit).

The PR description is corrected and now links this thread instead of asserting the opposite.

plugins/kubernetes-react/src/lib/k8s/KubeObject.ts
marians: Agreed, and taken as documentation-only (option (a) of the two you offered) after discussing scope.

You are right on both writers. Client-side `kubectl apply` re-sends the manifest as a strategic-merge patch, which is exactly why it undoes a `kubectl edit`; and `helm upgrade`'s three-way merge resets drift on chart-declared fields. Both land as `operation: Update`, so neither is detected. Your concrete scenario — a chart shipping a Kustomization with `spec.suspend` declared — is a real gap.

Rather than infer declarative intent from manager names (`helm`, `kubectl-client-side-apply`), which is a guess that can false-positive on an imperative writer that happens to be one of those, the doc comment now states plainly that this detects **SSA appliers only**, names both misses, and tells callers to treat an empty result as "no SSA applier claims this" rather than "nothing will revert this". The changeset and PR description say the same. I also dropped the incorrect "a HelmRelease-deployed object is applied by `helm-controller`" — that only holds via the drift-correction path, i.e. with `spec.driftDetection` enabled, as you noted.

Happy to add the `last-applied-configuration` check as a follow-up if the Helm/kubectl-apply case turns out to bite in practice — that one is precise rather than heuristic, so it is the part worth doing properly.

plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
marians: Verified and fixed in 2d6d27f4. You were right down to the mechanism, so I checked both halves at the source:

- `@material-ui/core/ButtonBase/ButtonBase.js:71` — `'&$disabled': { pointerEvents: 'none', cursor: 'default' }`, exactly the line you cited.
- bui's `Button.module.css` — the only disabled rule is `&[data-disabled='true'] { cursor: not-allowed; }`, and `pointer-events` appears **zero** times in the whole module.

So the disabled `<button>` was indeed absorbing the pointer and all anyone got was MUI's native `title` (which it renders on the child while closed — hence my test passing). Fixed by applying `pointer-events: none` only while disabled, on Suspend and on the pre-existing Reconcile button.

One honest caveat on the new test, which I found by deliberately reverting the fix to check it had teeth: **only the `toHaveStyle({ pointerEvents: 'none' })` assertion actually fails without it.** The hover assertions pass either way, because jsdom cannot model pointer-events retargeting and `userEvent.hover(span)` dispatches straight at the span rather than at the button. The style assertion is therefore the regression guard, and the test says so in a comment so nobody mistakes the hover for coverage it does not provide.

I went with `pointer-events` rather than `aria-disabled` + a blocked `onPress` to keep the change contained. Worth noting the residual downside you implicitly raise: a truly `disabled` button is not focusable, so keyboard users still cannot reach either tooltip. The `aria-disabled` route would fix that as well — happy to do it as a follow-up if you think the a11y gap is worth the churn across both buttons and their tests.

plugins/flux-react/src/components/FluxOverview/ResourceCard/FluxResourceActions/FluxResourceActions.tsx
marians: Both fixed in 2d6d27f4.

- Dedupe moved into `getApplyFieldOwners` itself (`[...new Set(managers)]`) rather than the caller, since the property "one manager appears once" belongs to the accessor — the docstring now records why duplicates arise at all (entries keyed by manager + operation + apiVersion + subresource). Test covers the CRD-version-migration shape you describe: the same controller at `v1beta2` and `v1`.
- Oxford join via a small `joinWithAnd` helper: one owner renders bare, two as "a and b", three or more as "a, b and c". Test asserts the three-owner case.

And thanks for spelling out that SSA co-ownership makes multi-owner real rather than theoretical — that is what justifies keeping the plural handling at all instead of simplifying to a single name.

plugins/flux-react/src/components/FluxOverview/ResourceCard/utils/fluxActions.ts
marians: Both fixed in 2d6d27f4 — and you are right that the history does not contain the removal I implied: #2001 was squash-merged with `ImagePolicy` already in the list, so "originally left out of this list" would send a reader hunting for a commit that does not exist. Reworded to "the reason previously given here … was wrong", which is the part that actually changed.

Good catch on `FluxResourceActions.tsx` too — that example was not just stale but actively contradicted by the test two files over. Now `(e.g. a ConfigMap, or any non-Flux object)`, which is what the guard genuinely filters and what the test exercises.

While in there I found a third instance of the same wrong claim that your comment led me to: `FluxResourceActions.test.tsx` still said "only the `flux` CLI lacks a subcommand" in the ImagePolicy test. Fixed as well, so all three copies of the bad narrative are gone from the working tree.

marians:
Thanks @fiunchinho — both valid, both fixed in f3e2d935. Neither was resolved by the earlier review round, and I had missed this comment entirely (I enumerated review comments and review summaries, not issue comments).

**1. Contradictory tooltips.** Correct, and the state you name was covered by a test that asserted only that both buttons were disabled — never what they *said*, which is why it slipped through. `reconcileHint` now branches on `isSuspended && isSuspendFieldManaged` first, ahead of the plain suspended case, and gets its own message:

> Suspended resources are not reconciled, and spec.suspend is applied by kustomize-controller — resume it at the source that applies it.

A test asserts that exact string and additionally that `/Resume it first/` is **not** present in that state, so the dead end cannot come back unnoticed. The plain "Resume it first" wording is kept for the unmanaged case, where it is still the right advice, with its own test.

**2. Git-specific wording.** Also correct, and a genuine internal contradiction: my own docstring says the check covers helm-controller drift correction and `kubectl apply --server-side`. Both hints now say "at the source that applies it", with a comment above the builder recording why the wording is deliberately not Git-specific. A test using a `helm-controller` owner asserts the neutral phrasing and that `/in Git/` is absent.

Worth noting the residual imprecision, since I would rather flag it than let it read as fully solved: the tooltip names the *manager* (e.g. `kustomize-controller`), which is only an indirection towards the thing a user can actually edit. Resolving it to a concrete source would mean walking `kustomize.toolkit.fluxcd.io/name`/`-namespace` ownership labels back to the parent Kustomization and its repository. That is a bigger change and it is not in this PR.

214 tests pass, `tsc`/lint/prettier clean.

PR #2005, opened 2026-07-29
feat(agent-platform): add the session detail page

marians:
### What does this PR do?

Makes Sessions rows clickable. `/agent-platform/sessions/<installation>/<id>` shows one session: header, stats strip, and the conversation.

> **Stacked on #2003** (which is stacked on #2002). Base is `agent-platform-session-timeline` so the diff stays clean; GitHub retargets on merge.

**Read-only.** kagent can rename, delete and continue a session and the prototype offers all three — none are wired up, so this ships without a write path.

### How does it look like?

image

Expanding a tool call shows its arguments and result in a `CodeBlock`; reasoning shows the full text.

Decisions worth a second opinion:

**Timestamps are absolute here, relative in the list.** Both ends of a session usually fall on the same day, so the relative form read *"Started 1 day ago · last activity 1 day ago"* for a session that took three minutes, and printed "1 day ago" identically on every turn marker — hiding the progression the timeline exists to show. One marker per A2A task, since a task's timestamp is the finest granularity that exists (see #2003 for why).

**Input tokens are labelled "billed, cumulative".** Every model call re-sends the whole context, so a 4-turn session with a large tool catalogue reaches 1.4M prompt tokens across 14 calls. Genuine billed usage — kagent's own UI sums it identically — but unlabelled it reads as a bug. There is deliberately **no combined total**: input and output are priced differently, so the sum isn't a number anyone acts on.

**Duration is wall-clock** (`updated_at − created_at`), since kagent records no per-turn durations. It includes time the user was away between turns.

**Calls through Muster are unwrapped.** Agents reach most MCP tools via muster's `call_tool`, so untreated every row read `call_tool` with the real tool buried in the arguments — [klaus-gateway#163](https://github.com/giantswarm/klaus-gateway/issues/163). Done in the parser, and only when the payload really has the wrapper shape, so a future `call_tool` change degrades to showing the proxy rather than losing the call. Unwrapped 7 of 17 calls on a real session.

**`ask_user` is not an approval.** ADK wraps both a permission request and a question in the same `adk_request_confirmation`. The parser records which (`asks: 'approval' | 'input'`) so the UI doesn't string-match a tool name; a question reads *User input requested · Responded · Questions* rather than *Approval requested · approved*.

**A missing session is a not-found state, not an error.** kagent answers 404 for deleted, never-existed and belongs-to-someone-else alike. A genuine read failure gets a separate danger alert with its message.

**Row links use a real anchor plus a whole-row click.** Not `rowConfig.getHref`: `BUIProvider` isn't mounted in this app, so react-aria's `RouterProvider` is inactive and a bui `href` does a full page reload.

**`session` and `session-tasks` are excluded from `localStorage`** for two independent reasons: a conversation is user-scoped data that mustn't outlive sign-out on disk, and at ~500 KB per session it would evict the fleet lists the persistence exists for.

The tab's content is now a `SessionsRouter` hoisting the query client and fleet-wide `Agent` list above both screens, so opening a session reuses what the list already loaded.

### Fixes found by testing against real sessions

The page was built against fixtures, then driven against live gazelle sessions. Everything below is a bug that only real data exposed:

| Symptom | Cause |
| --- | --- |
| `Total 0` beside `Input 1.4M` | `adk_usage_metadata` carries no `totalTokenCount` at all; summing reported totals stayed at zero. kagent's own UI has the same hole. Now derived from the parts, with a reported total still winning when present. |
| "Approval requested" accordion opened onto nothing | `CallDetail` handled only tool/agent calls and returned `null` for approvals, while the proposed arguments sat unrendered. Entries with no payload are now plain rows, not accordions. |
| No Details control on an `ask_user` session | Its visibility keyed on *activity*, which excludes approvals. That exclusion is right for **Hidden** (an approval is the user's own action) but shouldn't govern expand/collapse. |
| The user's reply missing from the conversation | The reply rides on the same message as the `decision_type` data part, and the decision branch returned immediately — discarding it. The agent appeared to answer something nobody said. |

Also corrects a claim I made on #2003: real payloads **do** use `adk_`-prefixed metadata. Two gazelle sessions a day apart used `kagent_usage_metadata` and `adk_usage_metadata` respectively, so the dual-prefix reader is load-bearing — reading one prefix silently zeroes a session's tokens.

### What the page cannot show

kagent stores none of this, so the prototype's remaining fields have no data behind them and shouldn't be re-added speculatively: **cost**, **tokens/second**, **context-window usage**, the owning **team**, the **trigger** that started the session, a **linked work item**, produced **results**, and **evaluation**. Delegation entries are inert, since the response doesn't reliably carry the child session's id.

The session id is deliberately not displayed — a 64-character opaque string that told the reader nothing, and it's in the URL already.

Yes, done — `docs/agent-platform.md` gains a "session detail page" section covering the two reads, why events are ignored, per-turn timestamps, the entry taxonomy, Muster unwrapping, cumulative-token labelling, and an explicit list of what the page cannot show.

---

540 tests, clean `tsc` and `lint`. Verified live against gazelle: two requests per load, deep-link reload, garbage id → 404 (not 5xx), `localStorage` at 51 KB with no bearer token and no conversation content, and no console warnings.

### DIFF
--- a/plugins/agent-platform/src/components/SessionTimeline/SessionTimeline.tsx
+++ b/plugins/agent-platform/src/components/SessionTimeline/SessionTimeline.tsx
@@ -1,0 +1,145 @@
+import { useMemo, useState } from 'react';
+import {
+  Alert,
+  Flex,
+  Text,
+  ToggleButton,
+  ToggleButtonGroup,
+} from '@backstage/ui';
+import { makeStyles } from '@material-ui/core';
+import { DateComponent } from '@giantswarm/backstage-plugin-ui-react';
+
+import { SessionTimeline as Timeline } from '../../lib/kagentTimeline';
+import { TimelineEntry } from './TimelineEntry';
+import { ActivityDetail, groupIntoTurns, isActivityItem } from './helpers';
+
+const useStyles = makeStyles(theme => ({
+  turnMarker: {
+    display: 'flex',
+    alignItems: 'center',
+    gap: theme.spacing(1),
+    marginTop: theme.spacing(1),
+  },
+  rule: {
+    flex: 1,
+    height: 1,
+    backgroundColor: theme.palette.divider,
+  },
+}));
+
+export type SessionTimelineProps = {
+  timeline: Timeline;
+  /** Display name of the session's agent, used to label its messages. */
+  agentName?: string;
+};
+
+/**
+ * The session's conversation, grouped into turns.
+ *
+ * **Timestamps are per turn, not per item.** A2A messages carry no time of their
+ * own, and kagent's stored events — which looked like a source for one — turned out
+ * to be ADK events with no `messageId` to correlate against. So a task's timestamp
+ * is the finest granularity that exists, and it is shown once per turn rather than
+ * repeated on every item, which would imply precision we don't have.
+ */
+export function SessionTimeline({ timeline, agentName }: SessionTimelineProps) {
+  const classes = useStyles();
+  // Collapsed by default: the agent's working is why this screen is worth opening,
+  // but a wall of expanded tool payloads is unreadable.
+  const [detail, setDetail] = useState<ActivityDetail>('collapsed');
+
+  const turns = useMemo(() => groupIntoTurns(timeline.items), [timeline.items]);
+
+  const activityCount = useMemo(
+    () => timeline.items.filter(isActivityItem).length,
+    [timeline.items],
+  );
+
+  if (timeline.items.length === 0) {
+    return (
+      <Text variant="body-medium" color="secondary">
+        This session has no messages yet.
+      </Text>
+    );
+  }
+
+  return (
+    <Flex direction="column" gap="3">
+      <Flex
+        align="center"
+        justify="between"
+        gap="2"
+        style={{ flexWrap: 'wrap' }}
+      >
+        <Text variant="title-small">Timeline</Text>
+        {activityCount > 0 && (
+          <Flex align="center" gap="2">
+            <Text variant="body-small" color="secondary">
+              Agent activity
+            </Text>
+            <ToggleButtonGroup
+              selectionMode="single"
+              disallowEmptySelection
+              selectedKeys={[detail]}
+              onSelectionChange={keys => {
+                const next = [...keys][0];
+                if (
+                  next === 'hidden' ||
+                  next === 'collapsed' ||
+                  next === 'expanded'
+                ) {
+                  setDetail(next);
+                }
+              }}
+            >
+              <ToggleButton id="hidden">Hidden</ToggleButton>
+              <ToggleButton id="collapsed">Collapsed</ToggleButton>
+              <ToggleButton id="expanded">Expanded</ToggleButton>
+            </ToggleButtonGroup>
+          </Flex>
+        )}
+      </Flex>
+
+      {timeline.skippedMessages > 0 && (
+        <Alert
+          status="warning"
+          title="Some messages could not be read"
+          description={`${timeline.skippedMessages} ${
+            timeline.skippedMessages === 1 ? 'message' : 'messages'
+          } in this session did not match the shape we expect from kagent and are not shown.`}
+        />
+      )}
+
+      {turns.map(turn => {
+        const visible =
+          detail === 'hidden'
+            ? turn.items.filter(item => !isActivityItem(item))
+            : turn.items;
+        if (visible.length === 0) {
+          return null;
+        }
+        return (
+          <Flex key={turn.taskIndex} direction="column" gap="3">
+            <div className={classes.turnMarker}>
+              <Text variant="body-small" color="secondary">
+                {turn.at ? <DateComponent value={turn.at} relative /> : '—'}
+              </Text>
+              <span className={classes.rule} />
+            </div>
+            {visible.map(item => (
+              <TimelineEntry
+                // Keyed on the detail setting as well as the item: an
+                // AccordionGroup's expanded set is only read on mount, so the
+                // global control takes effect by remounting the entries.
+                key={`${item.id}:${detail}`}
+                item={item}
+                resolvedAgentName={agentName}
+                defaultExpanded={detail === 'expanded'}
+              />
+            ))}
+          </Flex>
+        );
+      })}
+    </Flex>
+  );
+}
--- a/plugins/agent-platform/src/components/SessionTimeline/SessionTimeline.tsx
+++ b/plugins/agent-platform/src/components/SessionTimeline/SessionTimeline.tsx
@@ -120,9 +120,13 @@ export function SessionTimeline({ timeline, agentName }: SessionTimelineProps) {
         }
         return (
           <Flex key={turn.taskIndex} direction="column" gap="3">
+            {/* Absolute, not relative: every turn of a session usually falls on
+                the same day, so the relative form printed "1 day ago" three times
+                and hid the progression entirely. An exact time shows how the
+                session actually unfolded. */}
             <div className={classes.turnMarker}>
               <Text variant="body-small" color="secondary">
-                {turn.at ? <DateComponent value={turn.at} relative /> : '—'}
+                {turn.at ? <DateComponent value={turn.at} /> : '—'}
               </Text>
               <span className={classes.rule} />
             </div>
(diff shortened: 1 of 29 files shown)


plugins/agent-platform/src/components/SessionsTable/SessionsTable.tsx
```diff
@@ -169,6 +207,17 @@ export function SessionsTable({
       <Table<SessionRow>
         {...tableProps}
         columnConfig={columnConfig}
+        rowConfig={{
+          // Whole-row click as a convenience, on top of the anchor in the title
+          // cell. `onClick` + navigate rather than `getHref`, because without
+          // BUIProvider a bui href does a full page reload (see the title cell).
+          onClick: row => {
```
marians, line 214: **Row `onClick` fires *in addition to* the in-cell anchor, so a title click navigates twice.**

I verified this rather than guessed: rendering `SessionsTable` in a test app with `useNavigate` mocked and clicking the title `link` calls `navigate('/agent-platform/sessions/gazelle/abc')` — react-aria's row press handler runs even though the anchor already handled the same click (bui's `Row` passes `rowConfig.onClick` as react-aria `onAction`, and `usePress` has no guard for interactive descendants).

Two concrete consequences:

1. Plain click on the title: react-router's `Link` pushes `/agent-platform/sessions/<inst>/<id>`, then the row handler pushes the *same* path again → two identical history entries, so **Back needs two presses** to get back to the list.
2. With a modifier: in the same test the row handler still ran with Meta held. In a browser react-router leaves modified events to the browser (new tab), so **cmd-click opens the session in a new tab _and_ navigates the current tab away** — exactly what the anchor was added to avoid.

Fix options: drop `rowConfig.onClick` and keep the anchor only; or keep both but make the title cell swallow the press before it reaches the row (`click` alone isn't enough — `usePress` works off `pointerdown`/`pointerup`, so those need `stopPropagation` too).

plugins/agent-platform/src/components/SessionTimeline/SessionTimeline.tsx
```diff
@@ -0,0 +1,171 @@
...
+  );
+
+  // Whether `hidden` would actually remove anything, which decides if that option
+  // is worth offering.
+  const hasActivity = useMemo(
+    () => timeline.items.some(isActivityItem),
+    [timeline.items],
+  );
+
+  if (timeline.items.length === 0) {
```
marians, line 76: **The `skippedMessages` warning is unreachable in the case it exists for.**

This early return sits above the `timeline.skippedMessages > 0` alert further down. When *every* history entry fails `a2aMessageWireSchema` — the schema-drift case `skippedMessages` was introduced to report — `items` is empty and `skippedMessages > 0`, so the user is told "This session has no messages yet." and the "Some messages could not be read" alert never renders. Total data loss is presented as an ordinary empty session.

Either render the alert in this branch too, or gate the early return on `timeline.skippedMessages === 0`.

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -210,7 +228,27 @@ export function buildTimeline(tasks: A2aTaskWire[]): SessionTimeline {
...
+          // text part on this same message, and bailing out discarded it — the
+          // question and the agent's follow-up were shown, but what the user
+          // actually said vanished from the conversation.
+          //
+          // The text parts fall through to the normal handling below. When there
+          // are none, the answers are recovered from the decision payload's
+          // `ask_user_answers` instead: kagent's own UI reads only that field, so it
+          // may be the sole carrier on sessions that did not come through a
+          // gateway that also writes the text part.
+          if (decision.answers.length > 0 && !hasTextPart(parts)) {
```
marians, line 242: **`hasTextPart` and `flushText` disagree about what counts as text, which can re-introduce the "user's reply vanished" bug this PR fixes.**

`hasTextPart` is `parsePart(p)?.text !== undefined`, so a part with `text: ""` (or whitespace only) counts as carrying text. But `flushText` does `chunks.join('').trim()` and returns without pushing when the result is empty. So a decision message that carries `ask_user_answers` *and* a blank text part loses the reply both ways: the fallback here is skipped because `hasTextPart` is true, and the text run is dropped because it trims to nothing — the question and the agent's follow-up show, the user's answer doesn't.

Making the predicate require non-blank text keeps the two in step:
```suggestion
          if (decision.answers.length > 0 && !hasTextPart(parts)) {
```
(i.e. change `hasTextPart` to `parts.some(p => Boolean(parsePart(p)?.text?.trim()))`.)

plugins/agent-platform/src/lib/kagentParts.ts
```diff
@@ -33,11 +33,71 @@ const AGENT_NAMESPACE_SEPARATOR = '__NS__';
...
+  }
+  const args = asRecord(call.args);
+  const innerName = asNonEmptyString(args?.name);
+  if (!innerName) {
+    return call;
+  }
+  return {
+    id: call.id,
+    name: innerName,
+    args: args?.arguments,
```
marians, line 96: **The unwrap guard doesn't match its own doc comment, so a future `call_tool` change loses the arguments instead of degrading to the wrapper.**

The docstring promises it "only unwraps when the payload actually has the wrapper's shape (`name` a non-empty string, **with the inner call's `arguments`**)" and that anything else "degrades to showing the wrapper rather than losing the call". The code only checks `name`.

Concretely: if muster ever renames the inner-args key (`arguments` → `args`, or nests it), every proxied call still unwraps — the row shows the real tool name but `args` becomes `undefined`, so the arguments are silently gone and, while the call is still pending, `hasExpandableDetail` is false and the entry renders as an inert row with nothing to open. That's the exact failure mode the comment says it protects against.

```suggestion
  if (args?.arguments === undefined) {
    return call;
  }
  return {
    id: call.id,
    name: innerName,
    args: args.arguments,
```

plugins/agent-platform/src/components/SessionTimeline/SessionTimeline.tsx
```diff
@@ -0,0 +1,171 @@
...
+      {turns.map(turn => {
+        const visible =
+          detail === 'hidden'
+            ? turn.items.filter(item => !isActivityItem(item))
+            : turn.items;
+        if (visible.length === 0) {
+          return null;
+        }
+        return (
+          <Flex key={turn.taskIndex} direction="column" gap="3">
```
marians, line 144: **Duplicate React key in exactly the case `groupIntoTurns` was written to handle.**

`groupIntoTurns` deliberately groups on *runs* of `taskIndex` rather than a map, so that "items always stay in the order `buildTimeline` produced even if a task index were ever to repeat non-contiguously". When that happens it emits two turns with the same `taskIndex`, and this key then collides: React warns and can reconcile the second group against the first, rendering one turn's entries under the other's timestamp.

Use the position in `turns` (or `${turn.taskIndex}:${turn.items[0].id}`) so the key is as unique as the grouping allows.

plugins/agent-platform/src/hooks/useSessionDetail.ts
```diff
@@ -0,0 +1,107 @@
...
+  const error = isNotFound
+    ? undefined
+    : (((sessionQuery.error ?? tasksQuery.error) as Error | null) ?? undefined);
+
+  return {
+    detail: sessionQuery.data ?? undefined,
+    timeline,
+    state,
+    taskCount: tasks?.length ?? 0,
+    isLoading: sessionQuery.isLoading || tasksQuery.isLoading,
```
marians, line 103: **A known 404 is withheld behind the tasks request.**

`isNotFound` is decided by the session read alone (deliberately, per the comment above), but `isLoading` ORs both queries and `SessionDetailPage` checks `isLoading` first. So when the session read 404s immediately — `NotFoundError` isn't retried, so this is fast — but the tasks read is slow or is being retried, the user keeps staring at `Progress` for the whole tasks lifetime before "Session not found" appears. A non-404 tasks failure retries with the 2s/4s/8s backoff from `QueryClientProvider`, so that's ~15s of spinner for a session we already know doesn't exist; an unreachable installation stretches it to the fetch timeout.

```suggestion
    isLoading: !isNotFound && (sessionQuery.isLoading || tasksQuery.isLoading),
```

marians (commented):
Post-merge review pass (bug hunt only). Six findings, one of them user-visible on every row click; the rest are edge cases and one hypothetical-robustness gap where a doc comment promises more than the code does.

marians:
Related issue: https://github.com/giantswarm/giantswarm/issues/37297

PR #2006, opened 2026-07-29
fix(agent-platform): correct six defects in the session detail page

marians:
### What does this PR do?

Fixes six defects found in a review pass over #2005. Each one has a test that fails without its fix — verified by stashing the fixes and re-running.

| # | Defect | Symptom |
| --- | --- | --- |
| 1 | Row click and title anchor both navigate | Back needs two presses; cmd-click steals the current tab |
| 2 | `skippedMessages` warning unreachable when there are no items | A wholly unreadable session reports "no messages yet" |
| 3 | `hasTextPart` counts blank text as words | An `ask_user` reply can still vanish |
| 4 | `unwrapProxiedCall` keys only on the inner `name` | A `call_tool` shape change would drop a call's arguments |
| 5 | Turns keyed on `taskIndex` | Duplicate React key in the case `groupIntoTurns` exists for |
| 6 | `isLoading` waits for the tasks read | A known 404 sits behind a spinner |

**Clicking a session title no longer navigates twice.** The Sessions table deliberately has two affordances — a real anchor in the title cell, plus a whole-row click — and they were both firing for a single click, because the row's `onClick` is react-aria's `onAction`, which fires for a press anywhere in the row with no exemption for interactive descendants. Two consequences, both user-visible:

- Plain click: react-router's `Link` pushed the path, then the row pushed the *same* path again → two identical history entries, so **Back needed two presses** to get back to the list.
- **cmd-click:** react-router leaves a modified event to the browser, so the session opened in a new tab — and the row handler then navigated the current tab away too. Exactly the opposite of why the anchor was added.

The anchor now swallows the press. `usePress` works off pointer events, so `pointerdown`/`pointerup` are what has to be stopped; stopping propagation doesn't set `defaultPrevented`, so the anchor's own client-side navigation and its analytics event are untouched. The whole-row click keeps working everywhere else in the row.

**A session whose messages can't be read now says so.** The "N messages could not be read" warning sat *below* the early return for an item-less timeline — so the single case the counter exists for, every history entry failing the schema, rendered "This session has no messages yet." and no warning at all. Total data loss presented as an ordinary empty session. The warning now renders in that branch too, and the accompanying line no longer denies the messages existed.

**A known-missing session stops loading immediately.** "Not found" is decided by the session read alone (deliberately — the tasks endpoint also 404s on a kagent too old to serve it), but `isLoading` waited for both reads. A 404 settles at once since `NotFoundError` isn't retried, yet the page stayed on a spinner for however long the *tasks* request took: the full 2s/4s/8s retry ladder for a non-404 failure, or the fetch timeout on an unreachable installation — with the answer already in hand.

Two of the six are cases where a comment promised more than the code delivered, which is worth calling out because both promises are still worth keeping:

**`unwrapProxiedCall`** said it "only unwraps when the payload actually has the wrapper's shape (`name` a non-empty string, with the inner call's `arguments`)" and that anything else "degrades to showing the wrapper rather than losing the call". It checked only `name`. Had muster renamed or nested the inner arguments, every proxied row would have kept naming the real tool while `args` silently became `undefined` — a call with its payload gone, rendered as an entry with nothing to expand.

The fix is not "require `arguments`", which would have regressed genuinely argument-less proxied calls back to reading `call_tool`. It unwraps when the payload is *nothing but* the wrapper: a non-empty `name`, and no key besides `name` and `arguments`. That separates the three cases cleanly — `{ name, arguments }` unwraps, `{ name }` unwraps (nothing to lose), `{ name, somethingElse }` degrades to the proxy with its payload intact.

**`hasTextPart`** decides whether an `ask_user` reply needs recovering from `ask_user_answers`, and counted a `{ text: "" }` part as the user's words — while `flushText` drops a run that trims to nothing. A decision message with a blank text part therefore lost the reply *both* ways: fallback skipped because "there is text", text dropped for being empty. The predicate now agrees with what actually renders. This is the same symptom #2005 fixed from the other direction, which is why it's worth a guard rather than a shrug.

The **turn key** is the mildest of the six and the fix is a one-liner, but it is worth doing precisely because `groupIntoTurns` was written to tolerate a non-contiguous `taskIndex` — it groups on runs specifically so items keep `buildTimeline`'s order in that case. The key then made the two resulting turns ambiguous to React. Its test asserts no duplicate-key warning is emitted, filtered to that message since MUI v4's `findDOMNode` notice fires here regardless.

Yes, done. `docs/agent-platform.md` now states the unwrap rule precisely (and why both halves are load-bearing), and gains a note that an unreadable message is counted and reported rather than hidden — including when *every* entry failed.

---

391 tests in the plugin, clean `tsc`, `lint` and `prettier`. The three navigation assertions run against the real bui `Table`, so they exercise react-aria's press behaviour rather than a stand-in.

### DIFF
--- a/.changeset/agent-platform-session-detail-review-fixes.md
+++ b/.changeset/agent-platform-session-detail-review-fixes.md
@@ -1,0 +1,39 @@
+---
+'@giantswarm/backstage-plugin-agent-platform': patch
+---
+
+Fix six defects found reviewing the session detail page.
+
+- **A click on a session title navigated twice.** The Sessions table has both an
+  anchor in the title cell and a whole-row click, and react-aria's row press fires
+  for a press anywhere in the row — the anchor included. So one click navigated
+  twice: two identical history entries, meaning Back needed two presses to return
+  to the list, and with cmd held the session opened in a new tab _and_ took the
+  current tab with it, which is the opposite of why the anchor is there. The title
+  anchor now swallows the press (`pointerdown`/`pointerup`, which is what
+  `usePress` listens to) so exactly one of the two affordances acts on any click.
+- **A wholly unreadable session claimed to be empty.** The timeline's "N messages
+  could not be read" warning sat below an early return for an item-less timeline —
+  so the one case it exists for, every history entry failing to parse, reported
+  "This session has no messages yet." and never warned at all. The warning now
+  renders in that branch too, with wording that does not deny the messages existed.
+- **A blank text part could still lose an `ask_user` reply.** The reply is recovered
+  from `ask_user_answers` only when the decision message carries no text part, but
+  the check counted a `{ text: "" }` part as words while the renderer drops text
+  that trims to nothing — losing the answer both ways. The check now agrees with
+  what actually renders.
+- **A `call_tool` shape change would have dropped a call's arguments.**
+  `unwrapProxiedCall` promised to degrade to showing muster's proxy rather than
+  losing a call, but keyed only on the inner `name`: had muster renamed or nested
+  the inner arguments, every proxied row would have named the real tool with its
+  arguments silently `undefined`, rendering as an entry with nothing to expand.
+  Unwrapping now also requires that the payload carries no key beyond `name` and
+  `arguments`. An argument-less proxied call (`{ name }` alone) still unwraps.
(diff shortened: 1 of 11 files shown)


marians:
Towards https://github.com/giantswarm/giantswarm/issues/37297

PR #2007, opened 2026-07-30
fix(muster): make the tool explorer's per-server Sign in work

marians:
### What does this PR do?

Fixes the dead **Sign in** button in the muster Tool Explorer's
`Authentication required` alert.

The button called `musterApi.signIn()`, which only resolves the Backstage OAuth
token for muster's own `authProvider` — a token the user necessarily already
holds, since it is what made `list_tools` succeed and report
`servers_requiring_auth` in the first place. `getCredentials` returned
instantly, the mutation "succeeded", queries were invalidated, and the same
alert re-rendered. A guaranteed no-op, not a flake.

The servers in that list need muster's *downstream, per-server* OAuth flow,
which the portal never triggered:

- `core_auth_login { server }` returns a muster sign-in URL (its OAuth proxy
  start endpoint).
- The user completes the flow in a browser; muster's own OAuth callback stores
  the token **and connects the server for that muster session** via its auth
  completion callback. No second tool call is needed.
- muster excludes SSO-managed servers (token forwarding/exchange) from
  `servers_requiring_auth`, so every entry really is user-loginable.

**`muster-backend`**

- `GET /auth/status` — muster's `auth://status` resource, read with a native MCP
  `resources/read`. Cheap per-server auth state, so the servers page doesn't have
  to pull the whole tool catalogue to know. Deliberately *not* the `get_resource`
  meta-tool: that one aggregates the resources of the **downstream** servers and
  never sees the aggregator's own `auth://status` (against gazelle,
  `list_resources` reports "No resources available"). muster's own CLI reads it
  the same way.
- `POST /auth/login` — `core_auth_login`, with muster's free-text answer
  normalised to `{ status, authUrl?, message }` by `parseAuthLoginResult`
  (mirroring how muster's own CLI detects these, `cmd/auth_helpers.go`).
  Muster's refusals — SSO-managed server, rate limit, undiscoverable issuer —
  are expected outcomes for this route, so they return a structured 200 rather
  than a 5xx that `MiddlewareFactory` would ship to Sentry.

**`muster`**

- New shared `ServerSignIn` component and `useServerSignIn` hook. **Sign in**
  asks muster for the server's sign-in URL and offers it as an
  `Open sign-in page ↗` link (new tab, no popup so nothing can be blocked),
  then polls `auth://status` until muster reports the server connected —
  invalidating the muster queries so the hidden tools appear on their own.
  Polling is bounded to 3 minutes so an abandoned flow stops.
- SSO-managed servers get an explanation instead of a button (only an
  administrator can fix those), including when muster's own SSO attempt failed.
- The alert now carries one row per server, matching muster's per-server flow.

Clicking **Sign in** on an auth-required MCP server now actually signs you in to
that server, and its tools appear without a reload. The same affordance is
available per server on the MCP Servers page under
"Authentication / token chain", where `Auth Required` was previously a dead end.

### How does it look like?

```
⚠ Authentication required
  1 server requires authentication and its tools stay hidden until you sign in.
  pro   [ Sign in ]

  ↓ click

⚠ Authentication required
  1 server requires authentication and its tools stay hidden until you sign in.
  pro   [ Open sign-in page ↗ ]  Waiting for you to finish signing in…
```

Fixes https://github.com/giantswarm/giantswarm/issues/37309

The prerequisite on the cluster side is already met: `muster.oauth.mcpClient`
is enabled by the shared-configs default, and gazelle has one `auth.mode: oauth`
server (`pro`).

Two things worth knowing, neither fixable here:

- muster's per-server auth state is scoped to the **MCP session**, which for us
  is the backend's cached MCP client (30 min TTL, keyed per user token). A token
  rotation or a long idle gap drops it and the affordance reappears.
- Optional follow-up on the muster side: setting
  `muster.oauth.mcpClient.postLoginRedirectAllowlist` to the devportal would let
  us append `&redirect=<page>` to the start URL and land the user back in the
  tool explorer instead of muster's static success page.

### Verified live against gazelle

`GET /api/muster/auth/status` returns all 51 aggregated servers, and the flow was
driven through the UI:

- `gazelle-mcp-pro` → `auth_required` with `auth_tool: core_auth_login`; **Sign
  in** produced a fresh `…/oauth/proxy/start?state=…` challenge rendered as the
  link (`target="_blank" rel="noopener noreferrer"`), with the alert switching to
  "Waiting for you to finish signing in…".
- `garm-mcp-prometheus` → `reauth_required` + `token_exchange_enabled` +
  `sso_attempt_failed`, i.e. a real instance of the SSO case: no dead button,
  just the explanation and a pointer at the server's trusted audiences.

Worth noting from that run: muster sets `auth_tool: core_auth_login` on
`reauth_required` servers even when they are SSO-managed, while `handleAuthLogin`
then refuses the manual login ("uses SSO and is connected automatically"). The UI
keys off `token_forwarding_enabled`/`token_exchange_enabled` rather than
`auth_tool`, so it doesn't offer a button muster would reject.

Yes — done: the muster-backend README documents both new endpoints.


plugins/muster-backend/src/router.ts
```diff
@@ -277,6 +282,62 @@ export async function createRouter(
...
+
+  /**
+   * Per-server authentication status for the calling user's muster session
+   * (muster's `auth://status` resource). Read by the "Sign in" affordances to
+   * tell an OAuth-loginable server from an SSO-managed one and to detect when a
+   * browser sign-in has completed.
+   */
+  router.get('/auth/status', async (req, res) => {
+    const { config: installation, client } = resolveInstallation(req);
+    const result = await client.getResource(
```
marians: `/auth/status` has no error handling, so it becomes exactly the Sentry firehose `/auth/login` was carefully written to avoid.

`getResource` throws `ServiceUnavailableError` (→ 503) when there is no text content, and any MCP-level failure (`resources/read` unsupported, `auth://status` not registered on an older muster, resource-not-found, transport error) propagates as a plain `Error` → 500. Both are ≥500, so `MiddlewareFactory` reports them to Sentry regardless of our own log level.

Concrete scenario: one configured installation runs a muster that doesn't register `auth://status`. `useServerSignIn` mounts the query on every MCP Servers page expansion (react-query default 3 retries) and, while a sign-in is outstanding, re-reads it every 5 s per waiting client. That is a continuous stream of 500/503 Sentry issues from an entirely expected "this muster doesn't support it" outcome — and the UI already has a path for it (the hook surfaces `statusError` as text).

Suggest treating an unavailable/unsupported resource the same way as a refused login: catch, log at `info`, and answer `200` with `{ servers: [] }` (or a structured `{ unsupported: true }`) so the affordance just doesn't appear.

plugins/muster/src/components/shared/ServerSignIn.tsx
```diff
@@ -0,0 +1,119 @@
...
+  } = useServerSignIn(serverName, installation);
+
+  const idle = !authUrl && !error;
+  if (onlyWhenRequired && !needsLogin && idle) {
+    return null;
+  }
+
+  if (isSsoManaged) {
+    if (!needsLogin) {
+      return null;
```
marians: The SSO branch re-gates on `needsLogin`, which silently blanks the tool explorer's rows.

`needsLogin` is only `auth_required` / `reauth_required`. So for a server that muster *did* put in `servers_requiring_auth` but whose `auth://status` entry is `{ status: 'failed' | 'disconnected' | 'sso_pending' | 'error' | 'unreachable', token_exchange_enabled: true }`, this returns `null` — and the tool explorer renders `ServerSignIn` *without* `onlyWhenRequired` precisely because muster already said the server is auth-gated. The result is the alert reading

> ⚠ Authentication required — 1 server requires authentication and its tools stay hidden until you sign in.

with no server name, no button, and no explanation. The user is told something is wrong but not what or where. Your own live run hit the neighbouring case (`reauth_required` + `token_exchange_enabled` + `sso_attempt_failed`), which only works because `reauth_required` happens to be in `NEEDS_LOGIN`.

Same guard also discards a just-produced `error`: if the user clicks **Sign in** while `auth://status` still says `auth_required`, muster refuses, and the next poll flips the entry to `failed` + `token_exchange_enabled`, muster's refusal message disappears from the screen.

Reaching this branch at all means we know the server is SSO-managed, so the explanation is always more useful than nothing — consider dropping the inner `if (!needsLogin) return null;` and letting `onlyWhenRequired` be the only suppressor.

plugins/muster/src/components/shared/useServerSignIn.ts
```diff
@@ -0,0 +1,147 @@
...
+    status,
+    isPending: signIn.isPending,
+    authUrl,
+    isWaiting,
+    hasTimedOut,
+    // A failed status read would otherwise leave the UI waiting forever for a
+    // transition it can never observe, so it is surfaced rather than swallowed.
+    error:
+      error ??
+      (statusError ? `Auth status: ${statusError.message}` : undefined),
```
marians: Folding `statusError` into `error` defeats `onlyWhenRequired`, and the MCP Servers page will hit that on a cold visit.

`ServerSignIn` computes `idle = !authUrl && !error`, so as soon as this returns a `statusError` string, `onlyWhenRequired && !needsLogin && idle` is false and the component renders — with a `Sign in` button, since `status` is `undefined`.

Concrete scenario: `AuthChain` is rendered from `StandardServerDisclosure`/`IntegrationServerDisclosure` *outside* the page's `authenticated` gate (only the "Tools" block is gated). Before the user clicks "Connect to muster", `getAuthStatus` gets a 401 from `readCallOptions` — after 3 react-query retries — so every expanded server with `spec.auth` shows

> [ Sign in ]  Auth status: The muster installation 'gazelle' requires a user token for auth provider '…', but the request did not include one.

next to the auth chain, and the button 401s the same way. That duplicates (and contradicts) the page's own "Connect to muster" gate with raw backend prose.

Two things worth doing: keep 401/403 out of `error` (or pass `authenticated` in and leave the query disabled until then), and set `retry: false` on the query so an expected 401 isn't retried three times.

plugins/muster-backend/src/authLogin.ts
```diff
@@ -0,0 +1,62 @@
...
+export function parseAuthLoginResult(payload: unknown): AuthLoginResult {
+  const message =
+    typeof payload === 'string' ? payload : JSON.stringify(payload ?? null);
+
+  const authUrl = message
+    .split('\n')
+    .map(line => line.trim())
+    .find(line => line.startsWith('https://') || line.startsWith('http://'));
+
+  if (authUrl) {
```
marians, line 64: URL detection runs *before* the connected-marker check, so a connected answer that happens to carry a URL line is misclassified.

Any `core_auth_login` answer with a bare `http(s)://…` line wins over `CONNECTED_MARKERS`, even when the markers are present. Muster's success answer already enumerates the server's capabilities ("Successfully connected to 'pro'!\n\nAvailable capabilities: …"); if that block (or a "does not require authentication" answer) ever includes the server endpoint, the issuer URL, or a docs link on its own line, we return `status: 'auth_required'` with that URL as `authUrl`. The UI then swaps the button for an `Open sign-in page ↗` link pointing at a non-challenge URL and starts a 3-minute poll for a transition that already happened.

The markers are the specific signal and the URL is the fallback, so check in that order:

```suggestion
  if (CONNECTED_MARKERS.some(marker => message.includes(marker))) {
    return { status: 'connected', message };
  }

  if (authUrl) {
    return { status: 'auth_required', authUrl, message };
  }
```

plugins/muster-backend/src/MusterMcpClient.ts
```diff
@@ -235,18 +225,46 @@ export class MusterMcpClient {
...
+   * keyed per user token, so a server behind per-user auth gets one MCP session
+   * per user -- which is also what makes muster's per-session downstream auth
+   * state (see getResource / auth://status) stable across requests.
+   */
+  private async connect(options?: { authToken?: string }): Promise<{
+    client: MCPClient;
+    cacheKey: string;
+  }> {
+    const authToken = options?.authToken;
+    const cacheKey = McpClientCache.buildKey(this.installation.name, authToken);
```
marians, line 246: Worth thinking through: when the installation has no `authProvider`, `authToken` is `undefined`, so `McpClientCache.buildKey` collapses to the bare installation name — a single MCP session shared by *every* devportal user.

That was harmless while the session only carried read-only tool discovery. It isn't once `core_auth_login` mutates its downstream auth state: on such an installation, user A completing the browser flow connects the downstream server for the shared session, so user B's `/auth/status` reports it `connected` and B can call that server's tools with A's downstream OAuth grant, without ever authenticating.

Reachable via the legacy single-installation path (`aiChat.mcp` entry with static headers and no `authProvider`) or any `muster.installations[]` entry that omits `authProvider`. Production gazelle sets one, so this is a config-shape hazard rather than a live bug — but it deserves either a guard (refuse `/auth/login` and `/auth/status` when `installation.authProvider` is unset) or an explicit note in the README next to the two new routes, since the failure is silent and cross-user.

plugins/muster-backend/src/router.ts
```diff
@@ -277,6 +282,62 @@ export async function createRouter(
...
+      throw new InputError('server is required in the request body');
+    }
+
+    const callOptions = readCallOptions(req, installation);
+
+    let payload: unknown;
+    try {
+      payload = await client.callTool(AUTH_LOGIN_TOOL, { server }, callOptions);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
```
marians, line 376: This `catch` is unconditional, so infrastructure faults are reported to the user as muster's own refusal.

The comment above justifies the structured 200 for *tool-level* refusals (SSO-managed, rate limit, no issuer), which come back via `unwrapTextContent`'s `isError` path. But the same `catch` also swallows: muster unreachable (`fetch failed`, `ECONNREFUSED`), `Attempted to send a request from a closed client`, a 401/403 from muster's own OAuth proxy, and `ServiceUnavailableError` from `invokeMetaTool`. Each becomes `{ status: 'error', message: 'fetch failed' }` at HTTP 200, which `useServerSignIn` renders verbatim next to the button as if muster had deliberately declined — no Sentry signal, no retry, no way to tell a broken aggregator from a policy decision.

Narrowing the swallow to the tool-error case (e.g. only when `isClosedClientError` is false and the error is not a Backstage `ServiceUnavailableError`/`AuthenticationError`) keeps the anti-Sentry intent while letting genuine outages surface.

plugins/muster/src/components/shared/ServerSignIn.tsx
```diff
@@ -0,0 +1,119 @@
...
+    );
+  }
+
+  return (
+    <div className={classes.root}>
+      {showName ? <code className={classes.name}>{serverName}</code> : null}
+      {authUrl ? (
+        <>
+          <ButtonLink
+            href={authUrl}
```
marians, line 122: Once `hasTimedOut` is set there is no way back except a full page reload, and the link that remains is likely already dead.

`authUrl` is still truthy after the timeout, so this branch keeps rendering the `ButtonLink` and the `Sign in` button never returns — while `isWaiting` is false, so polling has stopped. The copy says "open the link, then reload this page", but the link is the *original* `…/oauth/proxy/start?state=…` challenge, and after 3 minutes that `state` is plausibly expired or already consumed, in which case following it fails and the reload accomplishes nothing.

Concrete scenario: user clicks **Sign in**, gets distracted for 5 minutes, comes back. They see a link that no longer works and no affordance to mint a fresh one.

Restoring the button on timeout (render the `Sign in` button when `hasTimedOut`, keeping the "Still not connected" note) would let a second click get a new challenge, which is the only action that can actually recover.

plugins/muster/src/components/shared/useServerSignIn.ts
```diff
@@ -0,0 +1,147 @@
...
+  // landed yet, so keep polling until the timeout rather than silently dropping
+  // the sign-in link.
+  useEffect(() => {
+    if (!authUrl || status?.status !== 'connected') {
+      return;
+    }
+    setAuthUrl(undefined);
+    setHasTimedOut(false);
+    queryClient.invalidateQueries({ queryKey: ['muster'] });
+  }, [authUrl, status, queryClient]);
```
marians: The completion signal is gated on `authUrl`, which lives in this component's local state — so unmounting the row silently cancels a sign-in that in fact succeeded.

On the MCP Servers page `AuthChain` sits inside `DisclosureAccordion`, whose children unmount when it collapses. Scenario: user expands `garm-mcp-…`, clicks **Sign in**, opens the link in the new tab, and meanwhile collapses the accordion (or clicks through to the Tool Explorer tab and back — the page also background-refetches CRDs every 30 s). `authUrl` is gone, so when they finish the OAuth flow this effect never fires, `invalidateQueries(['muster'])` never runs, and the tools stay hidden behind the stale `servers_requiring_auth` / `Auth Required` state until a manual reload — the same "nothing happened" symptom this PR is fixing.

Making the wait outlive the row would need the pending-server set to be shared (a small context, or a react-query mutation cache keyed per server) rather than per-instance `useState`. At minimum, worth noting that the invalidation should probably also fire when `status` flips to `connected` while `authUrl` is unset but a login was started in this session.

plugins/muster/src/components/shared/useServerSignIn.ts
```diff
@@ -0,0 +1,147 @@
...
+    mutationFn: () => musterApi.signInServer(serverName, installation),
+    onSuccess: result => {
+      setHasTimedOut(false);
+      if (result.status === 'auth_required' && result.authUrl) {
+        setAuthUrl(result.authUrl);
+        setError(undefined);
+        return;
+      }
+      setAuthUrl(undefined);
+      if (result.status === 'connected') {
```
marians, line 175: The `connected` outcome gives the user no feedback at all, which can reproduce the exact no-op click this PR fixes.

`setError(undefined)` + `setAuthUrl(undefined)` + invalidate means the row re-renders as an identical `Sign in` button. That's fine when the invalidation makes the surrounding content unblock — but muster answering "already authenticated and connected" while `auth://status` / `servers_requiring_auth` still lists the server as auth-gated is a real combination (they are computed from different state, and the PR notes muster sets `auth_tool` on `reauth_required` servers it then refuses). In that case: click → brief pending → button returns → alert unchanged → click again → same. Your own test `keeps the button when muster reports the server already connected` asserts precisely this silent shape.

A short confirmation (or surfacing `result.message` as a neutral note rather than only on the error paths) would make the difference between "done" and "nothing happened" visible.

plugins/muster/src/components/shared/ServerSignIn.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+      servers: [{ name: 'pro', status: 'connected' }],
+    });
+
+    await waitFor(
+      () => {
+        expect(screen.queryByRole('link')).not.toBeInTheDocument();
+      },
+      { timeout: 10_000 },
+    );
+  }, 15_000);
```
marians: This test burns real wall-clock time against the 5 s `POLL_INTERVAL_MS`: a `waitFor` window of 10 s inside a 15 s Jest timeout, so it needs at least one full poll tick to land with only ~5 s of slack.

`ci:verify` runs the Jest workers concurrently and has already been memory/CPU constrained on this repo's runners, so a loaded worker can push the first `refetchInterval` tick past the window and flake — and a flake here reads as "the sign-in completion detection is broken", which is expensive to triage.

Injecting the interval (a `pollIntervalMs` option on `useServerSignIn`, or `jest.useFakeTimers()` + `advanceTimersByTime`) would test the same transition deterministically in milliseconds.

plugins/muster/src/components/McpServersPage/serverDetail.tsx
```diff
@@ -139,53 +143,64 @@ export function AuthChain({ server }: { server: MCPServer }) {
...
-          )}
-        </>
-      )}
-      {localMint?.enabled && (
-        <DefRow label="Local mint">
-          audience{' '}
-          <span className={classes.mono}>{localMint.audience ?? '-'}</span>
+    <>
+      <ServerSignIn
+        serverName={server.getName()}
```
marians: For a federated family this signs in to one arbitrary instance only.

`StandardServerDisclosure` renders `<AuthChain server={representative} />`, i.e. one CR chosen out of N per-management-cluster instances, and the block's own caption already warns "Shown for {repMc}; the auth/token chain differs per cluster". `auth://status` and `core_auth_login` are per *server*, so `serverName={server.getName()}` here only ever targets the representative — the other instances of the family keep their tools hidden with no affordance, and the user gets no hint that the button only covered one cluster.

In practice standard servers use forward-token/token-exchange, so most land in the SSO branch and this stays theoretical — but if any federated family is ever `auth.mode: oauth`, the family block silently under-delivers. Either render one `ServerSignIn` per instance (`servers.map`) in this block, or scope it to the integration disclosure where "one CR = one server" actually holds.

Separately: the early `return` above for `!auth || auth.type === 'none'` means a server whose CR has no `spec.auth` but which muster nonetheless reports as `auth_required` gets no sign-in here either.

marians (commented):
Careful pass over the diff. The core diagnosis is convincing and the shape of the fix (per-server `core_auth_login` + `auth://status` polling, native `resources/read`, structured 200 for muster's refusals) looks right — `yarn tsc`, `yarn lint`, and both test suites are clean locally.

Ten findings, none of them blocking the approach. The two I'd most want addressed before merge:

1. **`/auth/status` has no error handling** while `/auth/login` does — a muster without `auth://status` turns into a repeating 500/503 Sentry stream, amplified by the 5 s poll and react-query's default 3 retries.
2. **The SSO branch's inner `needsLogin` guard can blank the tool explorer's rows entirely**, leaving "1 server requires authentication" with no name, button, or explanation.

The rest are error-path/UX robustness: `statusError` defeating `onlyWhenRequired` on a cold MCP Servers page, `parseAuthLoginResult`'s URL-before-markers ordering, the unconditional `catch` on `/auth/login` masking outages as refusals, the shared MCP session when an installation omits `authProvider`, no recovery after `hasTimedOut`, the completion signal being tied to per-row `useState`, the silent `connected` outcome, the federated-family representative, and one real-timer test that's a flake risk in `ci:verify`.

plugins/muster-backend/src/router.ts
marians: Agreed, and it was the same mistake twice — I reasoned about this for `/auth/login` and then left the sibling route bare.

Fixed in 159b0cc2: `/auth/status` catches, logs at `info`, and answers `200 {"servers": []}`. An empty list already carries the right meaning ("no sign-in affordance applies here"), so no new `unsupported` shape was needed. Also set `retry: false` on the query, since retrying an expected 401 or an unsupported resource three times only multiplies it.

plugins/muster/src/components/shared/ServerSignIn.tsx
marians: Correct, and thanks for spotting that the live run only worked by luck of `reauth_required` being in `NEEDS_LOGIN`.

Dropped the inner guard in 159b0cc2 — reaching that branch means we already know the server is SSO-managed, so the explanation is unconditional and `onlyWhenRequired` is the only suppressor. The connected-SSO case on the servers page is still silent because the `onlyWhenRequired` check above it runs first. Covered by a new test with `{ status: "failed", token_forwarding_enabled: true }`.

plugins/muster/src/components/shared/useServerSignIn.ts
marians: Right, and this is a regression I introduced in the follow-up commit rather than the original.

Three changes in 159b0cc2:
- `statusError` is only surfaced while a sign-in is outstanding — that is when it explains a wait that can never end. Outside a flow it is noise, and it can no longer affect `idle`.
- `retry: false` on the query.
- The affordance is now rendered only when the page has `authenticated`, so on a cold visit there is no query at all rather than a 401 three times over.

New test asserts nothing renders on a status error under `onlyWhenRequired`.

plugins/muster-backend/src/authLogin.ts
marians: Applied your suggestion verbatim in 159b0cc2 — the markers are the specific signal and the URL is the fallback, so that ordering is simply correct. Added a test with `Successfully connected` plus a URL line on its own, which fails against the old order.

plugins/muster-backend/src/MusterMcpClient.ts
marians: This one I had not thought about, and the cross-user consequence is worse than the config smell.

Guarded rather than documented, in 159b0cc2: both routes require `installation.authProvider`. Without it `/auth/status` answers `{"servers": []}` and `/auth/login` returns a structured error explaining why, so the affordance simply never appears on such an installation. The reasoning is in a comment on the guard and in the README next to the routes, since the failure is silent and would be easy to reintroduce.

plugins/muster-backend/src/router.ts
marians: Agreed — the comment justified a narrow swallow and the code swallowed everything.

159b0cc2 adds `isInfrastructureError`: closed-client errors, `TypeError` (undici reports connection failures as one), and the Backstage error classes our own client throws. Those rethrow and keep their 5xx; only muster's plain-`Error` tool refusals become a structured 200. Parameterised tests cover both directions.

Remaining fuzziness worth naming: the discriminator is still "plain `Error` means muster declined", so a future transport error that arrives as a bare `Error` would be misfiled. Structured tool results from muster would be the real fix.

plugins/muster/src/components/shared/ServerSignIn.tsx
marians: Good catch — I had written "then reload this page" without checking that the link would still work, and a consumed or expired `state` makes that advice useless.

159b0cc2 restores the `Sign in` button once the deadline passes (with a "Timed out waiting for the sign-in — try again." note), so a second click mints a fresh challenge. The deadline moved into the shared pending entry, so it survives a remount with the flow.

plugins/muster/src/components/shared/useServerSignIn.ts
marians: You are right, and it is the same class of failure this PR exists to fix, which makes it worth the extra machinery rather than a note.

159b0cc2 moves the pending sign-in into the react-query cache under `["muster", "pending-sign-in", installation, server]` instead of `useState`. Collapsing the accordion no longer cancels the wait, and a remounted row resumes the same flow rather than starting a new one — two new hook tests assert exactly that (`signInServer` is still called once after unmount/remount) plus per-server isolation.

Still bounded: while every row is unmounted nothing polls, so completion is observed when a row comes back rather than in the background. Making it fully background would need the wait to live outside the component tree entirely.

plugins/muster/src/components/shared/useServerSignIn.ts
marians: Agreed — and you are right that my own test was asserting the silence rather than catching it.

159b0cc2 adds a `note` to the hook state carrying muster's message on the `connected` path, rendered as a neutral line. The test is renamed and now asserts the message is visible rather than that nothing changed.

plugins/muster/src/components/shared/ServerSignIn.test.tsx
marians: Fair, and the failure mode you describe (a flake here reading as "completion detection is broken") is the expensive part.

159b0cc2 takes the injection route: `useServerSignIn` accepts `{ pollIntervalMs }`, and the transition is now tested at hook level via `renderHook` with a 20 ms interval. The 15 s component test is gone, so nothing in this PR waits on wall-clock time.

plugins/muster/src/components/McpServersPage/serverDetail.tsx
marians: Both halves are real, and the early-return one bothered me more than the representative one — a CR with no `spec.auth` that muster still reports as `auth_required` was silently unreachable.

159b0cc2 moves the affordance out of `AuthChain` entirely and into the disclosures, so neither problem depends on the CR's `spec.auth`:
- integration disclosure: one row for its server;
- standard disclosure: `servers.map(...)`, one row per instance.

`onlyWhenRequired` keeps the family block quiet for the connected majority, so in practice the 25-instance families render nothing extra. `AuthChain`'s doc comment now records why the affordance is not there.

plugins/muster-backend/src/authLogin.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (isClosedClientError(error)) {
+    return true;
+  }
+  // undici surfaces connection failures ('fetch failed', ECONNREFUSED) as a
+  // TypeError rather than a named error class.
+  if (error instanceof TypeError) {
+    return true;
+  }
+  const name = (error as { name?: string } | undefined)?.name;
+  return name !== undefined && INFRASTRUCTURE_ERROR_NAMES.includes(name);
```
marians: `isInfrastructureError` misses the MCP transport's HTTP errors, so the half of the original finding about "401s from muster's OAuth proxy" is still open.

`StreamableHTTPError` (`@modelcontextprotocol/sdk/client/streamableHttp`) is `class StreamableHTTPError extends Error` and sets no custom `name` — it only assigns the HTTP status to `.code`:

```js
class StreamableHTTPError extends Error {
    constructor(code, message) {
        super(`Streamable HTTP error: ${message}`);
        this.code = code;
    }
}
```

So `.name === 'Error'`, it isn't a `TypeError`, and it isn't in `INFRASTRUCTURE_ERROR_NAMES` → `isInfrastructureError` returns `false`. Every non-2xx from muster's endpoint (`throw new StreamableHTTPError(response.status, 'Error POSTing to endpoint: …')`, and the 401/403 paths above it) is thrown out of `connect()`/`tool.execute` inside `/auth/login`'s `try`, so a 502 from the ingress or a 401 from the OAuth proxy still lands on the refusal path: HTTP 200, `{status: 'error'}`, `Streamable HTTP error: Error POSTing to endpoint: …` rendered in red as if muster had deliberately declined, and nothing in Sentry.

`.code` is the discriminator (same `err.code` idiom as the HPM proxy case):

```ts
if (typeof (error as { code?: unknown }).code === 'number') {
  return true;
}
```

`code` is `-1` for `Unexpected content type`, which is also infrastructure, so "has a numeric code" is the right test rather than `code >= 500`.

plugins/muster-backend/src/router.ts
```diff
@@ -277,6 +283,107 @@ export async function createRouter(
...
+    const callOptions = readCallOptions(req, installation);
+
+    try {
+      res.json(await client.getResource(AUTH_STATUS_RESOURCE, callOptions));
+    } catch (error) {
+      logger.info(`Muster ${AUTH_STATUS_RESOURCE} is unavailable`, {
+        installation: installation.name,
+        message: error instanceof Error ? error.message : String(error),
+      });
+      res.json({ servers: [] });
```
marians: This `catch` is unconditional, so it swallows infrastructure faults too — `isInfrastructureError` now exists but isn't used here.

The Sentry-flood argument for not 5xx-ing is sound, but the consequence is that `statusError` in `useServerSignIn` becomes unreachable for exactly the case it was written for. During a muster outage while a sign-in is outstanding, this answers `{servers: []}` at 200 → the frontend's `pending && statusError` message never fires → the row shows *"Waiting for you to finish signing in…"* for the full `POLL_TIMEOUT_MS` (3 min) with no hint that anything is wrong. The same commit removed the old `'Still not connected — open the link, then reload this page.'` copy, so there is now strictly less feedback in this scenario than before.

A discriminated answer keeps the poll cheap and lets the row say something:

```ts
res.json({ servers: [], unavailable: true });
```

(Or keep the 5xx for `isInfrastructureError(error)` only — the frontend's new `retry: false` already removes the 3×-per-mount amplification, leaving one request per 5 s tick.)

plugins/muster/src/components/shared/useServerSignIn.ts
```diff
@@ -0,0 +1,226 @@
...
+  useEffect(() => {
+    if (!pending || hasTimedOut) {
+      return undefined;
+    }
+    const timer = setTimeout(
+      () => setPending({ ...pending }),
+      Math.max(0, pending.deadline - Date.now()),
+    );
+    return () => clearTimeout(timer);
+  }, [pending, hasTimedOut, setPending]);
```
marians: Rewriting the entry (`setPending({...pending})`) rather than clearing it leaves a timed-out sign-in stuck on screen for the rest of the session.

After the deadline `pending` is still truthy, so `authUrl` is still truthy. Only `showLink` is gated on `hasTimedOut` in `ServerSignIn`, so:

- `idle = !authUrl && !error && !note` is `false` → `onlyWhenRequired` can no longer hide the row, on a server that may not need anything.
- `isWaiting` is `false` → `refetchInterval: false`, and the muster `QueryClient` sets `refetchOnWindowFocus: false` (`components/QueryClientProvider.tsx:12`), so `auth://status` is never re-read.
- The self-heal in the completion effect below needs a fresh `status === 'connected'`, which nothing fetches.

Net effect: a user who finishes the IdP round-trip in 3½ minutes comes back to `Sign in` + *"Timed out waiting for the sign-in — try again."* on an already-connected server, and it stays that way until a page reload. Clicking `Sign in` again does recover it, but the row is otherwise undismissable and its state is stale.

Clearing the entry at the deadline restores both `idle` and the status refetch:

```ts
const timer = setTimeout(
  () => setPending(null),
  Math.max(0, pending.deadline - Date.now()),
);
```

That does drop the `hasTimedOut` message, so if the "timed out" wording is worth keeping, it needs to live somewhere that survives clearing `pending` (e.g. a `timedOut: true` field on the entry with `authUrl` unset).

plugins/muster-backend/src/router.ts
```diff
@@ -277,6 +283,107 @@ export async function createRouter(
...
+    const { config: installation, client } = resolveInstallation(req);
+    const { server } = req.body ?? {};
+
+    if (typeof server !== 'string' || server === '') {
+      throw new InputError('server is required in the request body');
+    }
+    if (!hasPerUserSession(installation)) {
+      res.json({
+        status: 'error',
+        message: `The muster installation '${installation.name}' is configured without an authProvider, so it has no per-user session to authenticate a downstream server for.`,
```
marians, line 362: The gate itself is right, but on an installation without `authProvider` the UI now advertises a sign-in whose only possible outcome is this message.

`authProvider` is `getOptionalString` (`MusterMcpClient.ts:143`), and the tool explorer's alert is driven by `list_tools.servers_requiring_auth` (`ToolExplorerPage.tsx:47`), which is independent of it. So on such an installation every listed server renders a `Sign in` button, and clicking it returns this internal config prose — shown to the user in `color="danger"` — once per server. Unactionable for a portal user, and it reads like a fault rather than a deployment property.

The frontend already receives this fact: `requiresAuth: Boolean(installation.authProvider)` (`router.ts:115`), surfaced as `activeInstallationInfo.requiresAuth` and already read on the MCP servers page. Gating the affordance on it and explaining once in the alert would replace N failing buttons with one accurate sentence.

Worth confirming no configured installation actually lacks `authProvider` before merge — for those, this PR's feature is inert rather than degraded.

plugins/muster/src/components/shared/ServerSignIn.tsx
```diff
@@ -0,0 +1,137 @@
...
+    return (
+      <Text as="p" variant="body-small" color="secondary">
+        {serverName} authenticates through SSO from muster's own session, so
+        there is nothing to sign in to here.
+        {status?.sso_attempt_failed
+          ? " muster tried and failed to establish it for your session — check the server's trusted audiences with an administrator."
+          : ''}
+      </Text>
+    );
+  }
```
marians, line 114: Dropping the inner `!needsLogin` gate fixes the blank row, but this branch still discards `error` and `note`.

It returns before the `note`/`error` blocks below, so any message the mutation just produced is lost whenever `auth://status` reports `token_forwarding_enabled`/`token_exchange_enabled`. Concretely: the user clicks `Sign in`, muster refuses with something specific (`Rate limit exceeded. Too many authentication attempts.`), and if the status read reports the server as SSO-managed the row silently swaps that for the generic "authenticates through SSO…" paragraph.

Narrower than the original defect — for the SSO-refusal case the paragraph is genuinely the better explanation — but a rate-limit or issuer-discovery message vanishing is the same discard the commit set out to remove. Rendering `error` under the SSO paragraph rather than instead of it would cover both.

plugins/muster/src/components/shared/ServerSignIn.tsx
```diff
@@ -0,0 +1,137 @@
...
+    hasTimedOut,
+    error,
+    note,
+    isSsoManaged,
+    needsLogin,
+    signIn,
+  } = useServerSignIn(serverName, installation);
+
+  const idle = !authUrl && !error && !note;
+  if (onlyWhenRequired && !needsLogin && idle) {
```
marians: `onlyWhenRequired` suppresses the SSO-failure diagnosis on the one page where an operator would look for it.

For a server reported as `{status: 'failed', token_exchange_enabled: true, sso_attempt_failed: true}`: `needsLogin` is `false` (`failed` isn't in `NEEDS_LOGIN`) and `idle` is `true`, so this returns `null`. The *"muster tried and failed to establish it for your session — check the server's trusted audiences with an administrator"* line therefore only ever renders in the tool explorer, never in the MCP servers page's "Authentication / token chain" block.

That's the more useful of the two placements for this particular message: it names a concrete misconfiguration an operator can fix, and the servers page is where they'd be looking. The new test at `ServerSignIn.test.tsx:209` pins this shape only *without* `onlyWhenRequired`, so the suppressed path isn't covered either way.

Letting `isSsoManaged && status?.sso_attempt_failed` through the `onlyWhenRequired` gate would be enough — it stays quiet for healthy SSO servers, which is what the flag is for.

plugins/muster/src/components/shared/useServerSignIn.ts
```diff
@@ -0,0 +1,226 @@
...
+      (isFailure ? signInResult?.message : undefined) ??
+      signIn.error?.message ??
+      (pending && statusError
+        ? `Auth status: ${statusError.message}`
+        : undefined),
+    // 'connected' with the alert still up is a real combination (the two come
+    // from different muster state), and silence there is indistinguishable from
+    // the no-op click this feature fixes.
+    note:
+      signInResult?.status === 'connected' ? signInResult.message : undefined,
```
marians, line 249: Minor, but inconsistent with the pending-entry design above: `note` and `error` derive from `signIn.data`, which lives in the mutation and therefore in component state.

Two consequences:

- The `DisclosureAccordion` collapse that `PendingSignIn` was moved into the cache to survive still drops these. Collapse a row right after muster answers `connected`, reopen it, and the note is gone — the surviving `pending` entry is `null`, so with `onlyWhenRequired` the row goes back to rendering nothing.
- `signIn.data` is never reset, so in the other direction they never clear. A row that once answered "already connected" keeps `note` set, `idle` stays `false`, and the row is permanently visible under `onlyWhenRequired` for the life of the page.

Not worth a second cache entry on its own, but if the feedback is meant to be durable it belongs next to `PendingSignIn`; if it's meant to be transient, it wants a reset (e.g. `signIn.reset()` once `status` transitions) so it stops pinning the row open.

marians (commented):
Re-review of `159b0cc2` against the previous round. `yarn tsc` clean, `yarn lint` clean, 75 backend + 179 frontend tests pass. Nothing here challenges the approach.

**Closed properly:** marker-before-URL ordering in `parseAuthLoginResult`; the shared-session hazard (both auth routes now require `authProvider`); the SSO branch's blanking `return null`; `statusError` no longer defeating `onlyWhenRequired`; the accordion-collapse cancellation, with the pending entry moved to the query cache and remount + per-server-isolation tests to pin it; the silent `connected` click; per-instance rows for a federated family (`key={instance.getName()}` is safe — a family's CRs all come from the active installation's cluster, so the names are distinct); and the flaky real-timer poll test replaced by hook tests with an injectable interval.

**Seven residual points, in severity order:**

1. `isInfrastructureError` misses the MCP transport's HTTP errors — `StreamableHTTPError` carries the status on `.code` and leaves `.name === 'Error'`, so muster's OAuth-proxy 401s and ingress 5xx still reach the user as muster's own refusal at HTTP 200 and never reach Sentry. This is the part of the earlier finding that's still open.
2. `/auth/status`'s `catch` is unconditional, which makes the frontend's `pending && statusError` path unreachable during a real outage — and the old "open the link, then reload" copy was removed in the same commit, so that scenario now has less feedback than before.
3. A timed-out pending entry is rewritten rather than cleared, so the row stays visible with `Sign in` + "Timed out…" and never re-reads status (`refetchOnWindowFocus` is off plugin-wide).
4. On an installation without `authProvider` the tool explorer still renders a `Sign in` per server whose only outcome is internal config prose in red; `requiresAuth` is already on the wire.
5. The `isSsoManaged` branch still discards a just-produced `error`/`note`.
6. `onlyWhenRequired` suppresses the `sso_attempt_failed` diagnosis on the MCP servers page, which is where an operator would look for it.
7. `note`/`error` still live in component state, so they don't survive the collapse the pending entry was hardened against — and never clear, pinning the row open.

Details inline.

plugins/muster-backend/src/authLogin.ts
marians: The finding is right and it is now fixed in f14b88b8, but the mechanism does not apply to this stack, so it is worth recording why.

We do not use `@modelcontextprotocol/sdk` — `@ai-sdk/mcp` has its own transport and its dependencies are only `pkce-challenge`, `@ai-sdk/provider`, `@ai-sdk/provider-utils`. There is no `StreamableHTTPError` anywhere in it, and `.code` is not the field it sets.

What it actually throws for a non-2xx (`dist/index.js`, the HTTP POST path):

```js
const error2 = new MCPClientError({
  message: `MCP HTTP Transport Error: POSTing to endpoint (HTTP ${response.status}): ${text}`,
  statusCode: response.status,
  url: this.url.href,
  responseBody: text ?? undefined,
});
```

So: `name` defaults to `MCPClientError`, and the status is on **`statusCode`**, not `code`. Your conclusion was exactly right — it was neither a `TypeError` nor in `INFRASTRUCTURE_ERROR_NAMES`, so a 502 from the ingress or a 401 from muster's OAuth proxy reached the user as a refusal at HTTP 200.

`isInfrastructureError` now keys on `name === "MCPClientError"` plus a numeric `statusCode` **or** `code` (keeping your point that a `-1` protocol code is infrastructure too). `name` rather than `isInstance()` deliberately: `MCPClientError` extends `AISDKError`, whose `isInstance()` matches any `AISDKError`.

I checked every `MCPClientError` throw site before widening the predicate — all transport, protocol or config level, none of them a downstream tool's `isError` payload, which reaches us through `unwrapTextContent` as a plain `Error`. So the two classes do not overlap.

plugins/muster-backend/src/router.ts
marians: Agreed, and the second half of that is the part I got wrong: I removed the "reload this page" copy in the same commit that made the failure path silent, so that scenario ended up with less feedback than it started with.

Took the discriminated route in f14b88b8: `{ servers: [], unavailable: true, message }`. The flag is what a waiting row needs to distinguish "nothing needs a sign-in" from "we cannot tell", and it keeps the poll at 200 so there is still no Sentry stream. `useServerSignIn` renders `Cannot read the muster auth status: <message>` while a sign-in is outstanding, and stays quiet otherwise — both directions are now tested.

I kept 200 rather than 5xx-ing only for `isInfrastructureError`: the amplification is per 5 s tick per waiting client, and an unreadable status is not independently actionable — if muster is down, `/auth/login` and every other muster route are already reporting it.

plugins/muster/src/components/shared/useServerSignIn.ts
marians: You are right, and this is a good example of a "recovery" fix creating a worse stuck state than the thing it fixed.

f14b88b8 clears the entry at the deadline (`setPending(null)`), which restores both `idle` and the status refetch, and returns the row to a plain `Sign in` — the only action that can actually recover.

I dropped the "Timed out…" wording rather than keeping it in a `timedOut` field: durable feedback is precisely what pins the row open (your point 7), so re-adding it in another form would reintroduce the same defect. The alert already says the server needs authentication, so a bare button is not confusing.

`timeoutMs` is now injectable alongside `pollIntervalMs`, so there is a test asserting the entry is cleared and the cache holds nothing afterwards.

plugins/muster-backend/src/router.ts
marians: Agreed — the gate was right but the UI in front of it was not.

f14b88b8 reads `activeInstallationInfo.requiresAuth` in the tool explorer's alert. When false, the alert keeps its list of affected servers and replaces the N buttons with one sentence saying this portal talks to muster without per-user credentials and an administrator has to configure an auth provider. The servers page needs no change: `/auth/status` already answers `{servers: []}` there, so `onlyWhenRequired` hides everything.

On your merge question — I could not confirm it. Neither the `giantswarm-configs` nor the `shared-configs` clone I have contains a `muster.installations` block for backstage, so the devportal instances configure it somewhere I cannot see. My local config has `authProvider` on both gazelle and graveler. That is exactly why this now degrades to one accurate sentence rather than relying on the assumption.

Not covered by a test: the muster instance context is not exported, so asserting this conditional would mean mocking the whole page's dependency surface. Flagging it rather than pretending otherwise.

plugins/muster/src/components/shared/ServerSignIn.tsx
marians: Agreed. Fixed in f14b88b8 — `note`/`error` are extracted into a `messages` fragment rendered *under* the SSO paragraph instead of the branch returning before them.

The test drives your exact sequence rather than a synthetic state: status `auth_required` → click → muster refuses with `Rate limit exceeded…` → next status read reports `failed` + `token_exchange_enabled` → assert both the paragraph and the rate-limit message are on screen.

That test also caught a conflict between this fix and your point 7 — see my reply there.

plugins/muster/src/components/shared/ServerSignIn.tsx
marians: Agreed, and the argument that the servers page is the better home for that particular message is the convincing part.

f14b88b8 lets `isSsoManaged && status?.sso_attempt_failed` through the `onlyWhenRequired` gate, exactly as you suggested. Two new tests: the diagnosis renders under `onlyWhenRequired`, and a healthy SSO server (`connected` + `token_exchange_enabled`) still renders nothing.

plugins/muster/src/components/shared/useServerSignIn.ts
marians: Both consequences are real, but writing the fix surfaced a conflict with your point 5, so I did not take either of the two options as stated.

`signIn.reset()` on a status transition was my first attempt, and it defeats the SSO fix: "muster refuses, the next poll reports the server SSO-managed" is precisely the sequence where the message matters, and the reset wipes it before it can be read. My own new test for #5 failed on it, which is how I found out.

So the harmful half is fixed at the render gate instead: `idle` in `ServerSignIn` now ignores `error`/`note`, so they can no longer defeat `onlyWhenRequired` or pin a row open. They are the answer to a click, which by definition happened on a row that was already visible, so they never need to keep it open.

The other half — not surviving a collapse — I am accepting and documenting rather than fixing, on the asymmetry that motivated caching the wait in the first place: a dropped *confirmation* is recoverable (click again and muster re-reports it), a dropped *wait* was not, because the completing event happens in another tab and never comes back. Putting both in the cache would mean inventing an expiry policy for the feedback, which is how #3 went wrong.

PR #2011, opened 2026-08-03
feat(agent-platform): show agent readiness in the agents list

marians:
### What does this PR do?

Adds a **Status** column to the agents list (`/agent-platform/agents`), which previously showed no readiness signal at all, and makes the list poll so that status stays current.

Readiness is derived from the kagent `Agent`'s `status.conditions` — the same data kagent's own REST API projects into its `accepted` / `deploymentReady` flags, so our states and kagent's UI agree on what each state means.

One deliberate divergence: kagent's API additionally requires the `Ready` condition's **reason** to be `DeploymentReady` or `WorkloadReady`, and we key on the status alone. That allowlist's only apparent purpose is to reject a missing Deployment (`Ready=Unknown`/`DeploymentNotFound`), which the status check already rejects — while a kagent that adds a third ready reason would make a healthy fleet render as "not ready" with a tooltip contradicting the label.

Because we already read the `Agent` CRs directly, this needs no new requests and no dependency on the kagent REST transport.

| State | Meaning |
|---|---|
| **Ready** | Accepted, and the backing workload has an available replica |
| **Not ready** | Accepted, but the workload is not up |
| **Not accepted** | The controller rejected the spec |
| **Pending** | Not reconciled yet, or reconciled against an older generation |

**Pending** has no equivalent in kagent's UI. It covers an agent whose `status.observedGeneration` still lags `metadata.generation`, so a just-edited agent reads as "not known yet" instead of presenting stale conditions as fact.

Also in scope:

- A new shared **`StatusLabel`** in `ui-react` (icon + label + intent colour), which this list is the first consumer of. It is built on bui rather than `core-components`, and exists mainly to get two things right once: keeping the label readable by assistive tech, and giving each intent a distinct silhouette so a status does not depend on colour alone.
- **Polling**, two-tier and evaluated per installation: 60s normally (equal to the query client's `staleTime`, so an interval refetch never undercuts data the client considers fresh), dropping to 5s for an installation that has an agent still converging. An agent non-ready for over 3 minutes is treated as durably broken and falls back to the baseline — without that bound a permanently broken agent pins its installation to the fast interval for as long as a tab stays open, which is what kagent's own UI does. Interval refetches only run while the tab is focused.
- **Sortable columns**, matching the sessions list. Status sorts by severity rather than alphabetically, so one click brings problems to the top. The default order is unchanged (installation, then name).
- The **"loading more agents" bar** now means "an installation has not reported its first result yet" rather than "a request is in flight". With polling the old meaning made it flash during steady state, and since it sits above the table it pushed the table down every minute. Its slot is now a fixed height so toggling it cannot shift the table.

Agent health is visible directly in the list, and non-ready agents explain themselves on hover using the controller's own message — the reconcile error, or `"Deployment is not ready, N/M pods are ready"`. In testing, a real broken agent surfaced its full Kyverno admission rejection (`disallow-capabilities-strict`, `disallow-privilege-escalation`, `require-run-as-nonroot`) without needing a `kubectl` round-trip.

The list also keeps itself current instead of needing a manual reload after creating an agent.

Pagination stays off, unlike the sessions list. bui's `useCompletePagination` never resets its offset when the data shrinks, so in a polling list a deletion elsewhere could leave the offset past the end of a shrunken list and show "No agents found." while agents exist. Sorting works without it.

### How does it look like?

image

Icons are the filled Material variants with four distinct silhouettes — circle-with-tick, triangle, circle-with-`!`, hourglass — so the state is legible without relying on colour alone. Colours use bui's current semantic tokens (`--bui-fg-positive` / `-warning` / `-negative` / `-secondary`), so they adapt to light and dark.

| State | Icon | Colour |
|---|---|---|
| Ready | `CheckCircle` | positive (green) |
| Not ready | `ReportProblem` | warning (amber) |
| Not accepted | `Error` | negative (red) |
| Pending | `HourglassEmpty` | secondary (grey) |

All four states were verified in the running app, along with a full poll cycle showing no layout shift.

The behaviour is deliberately aligned with kagent upstream (`kagent-dev/kagent`), specifically `go/core/internal/controller/reconciler/reconciler.go` for which conditions and reasons the controller sets, `go/core/internal/httpserver/handlers/agents.go` for how the REST API projects them, and `ui/src/components/AgentListView.tsx` for the state naming and wording.

Two implementation notes for reviewers:

- The status presentation lives in the new shared `ui-react` `StatusLabel`, not in `@backstage/core-components`' `Status*`. `Status*` puts `aria-hidden="true"` on the span wrapping both its icon *and* its children, so a label passed as its child is hidden from assistive tech and the cell reads as empty. It also hardcodes its icons and applies `position: relative; top: 3px` plus an 8px right margin sized for a 16px label rendered inside it, and its colours come from `theme.palette.status.*` so they cannot be themed with bui tokens. `StatusLabel` owns the markup instead, letting the icon inherit its colour from a single `--bui-fg-*` token via `currentColor`. `agent-platform` keeps only the readiness-to-presentation mapping.
- The readiness derivation lives in `kubernetes-react` as both `Agent` methods and free functions over the raw interface, so the `refetchInterval` callback — which sees `KubeObjectInterface[]`, not hydrated instances — reuses the identical logic rather than reimplementing it.

The remaining `Status*` users — `gs`'s `DeploymentStatus` (15 consumers) and `ClusterStatus`, `muster`'s `ExecutionStatusBadge`, and `flux-react`'s hand-rolled dot — are deliberately left alone. They sit in MUI tables where their current MUI typography matches its surroundings, so converting them now would introduce a font mismatch; they can adopt `StatusLabel` as their pages move to bui.

No.

### DIFF
--- a/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
+++ b/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
@@ -231,16 +231,22 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
     const hasInstallations = allInstallations.length > 0;
     const isBusy = hasInstallations && (isProbing || isLoading);
 
-    // "More rows may still arrive" means an installation has not reported its
-    // first result yet — deliberately *not* "some query is fetching".
+    // "More rows may still arrive", from two independent sources.
     //
-    // Now that the list polls, every installation refetches on an interval, and
-    // the cluster-access probe independently returns installations to
-    // `connecting` every few seconds. Deriving this from fetch/probe activity
-    // therefore flashed the progress bar during steady state, and because the bar
-    // sits above the table in a flex column, each flash pushed the table down.
-    // Once an installation has an entry — even an empty one — it can only ever
-    // *replace* its rows, never contribute the first ones, so it is settled.
+    // `pendingInstallations` covers an installation we already know is healthy
+    // but which has not reported its first result yet. Once it has an entry —
+    // even an empty one — it can only ever *replace* its rows, never contribute
+    // the first ones, so it is settled and a background refetch of it is not
+    // "loading more".
+    //
+    // `isProbing` covers the other half, and is why it cannot be dropped: an
+    // installation still `connecting` is not in `reachableInstallations` at all
+    // (that set is `healthy`-only), so it cannot appear in `pendingInstallations`
+    // — yet it may resolve to healthy and contribute rows seconds later. Without
+    // it the bar switches off during exactly the cold-load fan-in it exists for.
+    // It does not churn: the cluster-access connector only seeds `connecting` for
+    // installations it is not already tracking, so this settles once and stays
+    // settled across re-probes.
     const pendingInstallations = reachableInstallations.filter(
       cluster =>
         !(cluster in agentsByInstallation) &&
@@ -250,7 +256,8 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
     return {
       rows,
       isLoading: isBusy && rows.length === 0,
-      isLoadingMore: rows.length > 0 && pendingInstallations.length > 0,
+      isLoadingMore:
+        rows.length > 0 && (pendingInstallations.length > 0 || isProbing),
       hasInstallations,
       unreachableInstallations,
     };
--- a/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
+++ b/plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
@@ -14,7 +14,12 @@ import {
 import { useInstallations } from '@giantswarm/backstage-plugin-gs';
 import { useReachableInstallations } from '../../hooks/useReachableInstallations';
 import { useModelConfigs } from '../ModelConfigsProvider';
-import { AgentRow, sortAgentRows, toAgentRow } from './helpers';
+import {
+  AgentRow,
+  getAgentsRefetchInterval,
+  sortAgentRows,
+  toAgentRow,
+} from './helpers';
 
 export type AgentsContextValue = {
   /** Agents flattened into plain rows, ordered by installation + name. */
@@ -68,11 +73,15 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
   // round-trips per cluster for no benefit here. `clustersData` is the raw
   // per-cluster list result (present, and possibly empty, only for clusters that
   // responded successfully); `resources` are those hydrated into Agent instances.
+  // The refetch interval is two-tier and evaluated per installation — see
+  // `getAgentsRefetchInterval`. It keeps agent readiness fresh without polling
+  // the whole fleet fast: only installations with an agent that is still
+  // converging get the short interval.
   const { resources, clustersData, isLoading, errors } = useResources(
     reachableInstallations,
     Agent,
     {},
-    { enableDiscovery: false },
+    { enableDiscovery: false, refetchInterval: getAgentsRefetchInterval },
   );
 
   // Model labels resolve progressively as ModelConfigs arrive; we deliberately
@@ -221,10 +230,27 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
     // pin isLoading true and hide the "no installations configured" empty state.
     const hasInstallations = allInstallations.length > 0;
     const isBusy = hasInstallations && (isProbing || isLoading);
+
+    // "More rows may still arrive" means an installation has not reported its
+    // first result yet — deliberately *not* "some query is fetching".
+    //
+    // Now that the list polls, every installation refetches on an interval, and
+    // the cluster-access probe independently returns installations to
+    // `connecting` every few seconds. Deriving this from fetch/probe activity
+    // therefore flashed the progress bar during steady state, and because the bar
+    // sits above the table in a flex column, each flash pushed the table down.
+    // Once an installation has an entry — even an empty one — it can only ever
+    // *replace* its rows, never contribute the first ones, so it is settled.
+    const pendingInstallations = reachableInstallations.filter(
+      cluster =>
+        !(cluster in agentsByInstallation) &&
+        !erroredInstallations.includes(cluster),
+    );
+
     return {
       rows,
       isLoading: isBusy && rows.length === 0,
-      isLoadingMore: isBusy && rows.length > 0,
+      isLoadingMore: rows.length > 0 && pendingInstallations.length > 0,
       hasInstallations,
       unreachableInstallations,
     };
--- a/plugins/agent-platform/src/components/AgentsDataProvider/helpers.ts
+++ b/plugins/agent-platform/src/components/AgentsDataProvider/helpers.ts
@@ -1,7 +1,88 @@
+import type { crds } from '@giantswarm/k8s-types';
+import type { Query } from '@tanstack/react-query';
 import {
   Agent,
+  AgentReadiness,
+  deriveAgentReadiness,
+  getAgentStatusChangedAt,
+  isAgentTransitional,
+  KubeObjectInterface,
   ModelConfig,
 } from '@giantswarm/backstage-plugin-kubernetes-react';
+
+/**
+ * Baseline poll for the fleet-wide agent list. Deliberately equal to the query
+ * client's `staleTime` (see `QueryClientProvider`): interval refetches are *not*
+ * gated by staleness, so a shorter interval would refetch data the client still
+ * considers fresh and reintroduce exactly the background-refetch churn that
+ * `staleTime` was set to avoid.
+ *
+ * This tier only exists to notice agents created, edited or deleted elsewhere
+ * (kubectl, another tab, the scaffolder) — not to watch one converge, which is
+ * what the transitional tier below is for.
+ */
+const BASELINE_REFETCH_INTERVAL_MS = 60_000;
+
+/**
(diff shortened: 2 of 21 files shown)


plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
```diff
@@ -221,10 +230,27 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
...
+      cluster =>
+        !(cluster in agentsByInstallation) &&
+        !erroredInstallations.includes(cluster),
+    );
+
     return {
       rows,
       isLoading: isBusy && rows.length === 0,
-      isLoadingMore: isBusy && rows.length > 0,
+      isLoadingMore: rows.length > 0 && pendingInstallations.length > 0,
```
marians: This drops the "loading more" bar during exactly the phase it exists for, and the premise for the change doesn't seem to hold.

Premise: `isLoading` here is `queries.some(q => q.isLoading)` (`mapQueriesToClusters`), and in react-query v5 `isLoading === isPending && isFetching` — false for a query that already has data. So a background poll does *not* flip it. And `isProbing` doesn't churn either: `ClusterAccessConnector` only calls `recordConnecting` for installations *not already tracked* ("the interval re-probe never seeds"), so `entries.some(state === 'connecting')` is true only until each installation first resolves. The old `isBusy && rows.length > 0` therefore wouldn't have flashed every minute.

Regression: `pendingInstallations` is filtered from `reachableInstallations`, which only contains `healthy` installations (`REACHABLE_STATES = {'healthy'}`). On a cold load, installations still `connecting` are not in that set, so once the first healthy installation reports rows (`rows.length > 0`) and every *currently* healthy installation has an entry, `pendingInstallations` is empty and no bar shows — even though more installations are still being probed and will contribute rows seconds later. That's the fleet fan-in the bar was added for.

Suggest keeping the probe in the signal (the fixed-height slot already prevents the layout shift):

```suggestion
      isLoadingMore:
        rows.length > 0 && (pendingInstallations.length > 0 || isProbing),
```

plugins/kubernetes-react/src/lib/k8s/Agent.ts
```diff
@@ -2,6 +2,131 @@ import { crds } from '@giantswarm/k8s-types';
...
+
+  // The controller stamps `status.observedGeneration` on every status write,
+  // including when reconciliation *fails* — so a spec it genuinely rejects
+  // settles on `notAccepted` rather than sticking at `pending` forever.
+  const { generation } = json.metadata ?? {};
+  const observedGeneration = json.status?.observedGeneration;
+  if (
+    typeof generation === 'number' &&
+    (typeof observedGeneration !== 'number' || observedGeneration < generation)
+  ) {
```
marians, line 94: This fails closed on a field the CRD marks optional. In `crds.kagent.v1alpha2.Agent`, `status.observedGeneration` is `observedGeneration?: number`, while `metadata.generation` is always set by the apiserver for a resource with a status subresource.

Concrete scenario: an installation runs a kagent build that populates `status.conditions` but not the top-level `status.observedGeneration` (it's also easy to lose in an API bump — the per-condition `condition.observedGeneration` is the field that's guaranteed by `metav1.Condition`). Every agent there then reports `pending` forever: the Status column shows "Pending" for healthy and broken agents alike, `getReadinessMessage()` returns `undefined` so the tooltip explaining a broken agent disappears too — i.e. the whole feature silently reads as "unknown" on that installation, with no way to tell it apart from a genuinely unreconciled agent.

Cheap hardening: treat a missing `status.observedGeneration` as "not stale" (or fall back to `max(condition.observedGeneration)`), and only report `pending` when a *present* observedGeneration lags:

```suggestion
    typeof generation === 'number' &&
    typeof observedGeneration === 'number' &&
    observedGeneration < generation
```

plugins/kubernetes-react/src/lib/k8s/Agent.ts
```diff
@@ -2,6 +2,131 @@ import { crds } from '@giantswarm/k8s-types';
...
+  }
+
+  if (
+    findAgentCondition(json, AgentConditionType.Accepted)?.status !== 'True'
+  ) {
+    return 'notAccepted';
+  }
+
+  const ready = findAgentCondition(json, AgentConditionType.Ready);
+  if (ready?.status !== 'True' || !READY_REASONS.includes(ready.reason)) {
```
marians: The reason allowlist produces a self-contradictory cell on version skew. If a kagent version sets `Ready=True` with a reason outside `READY_REASONS` (renamed, or a third workload kind added — e.g. `StatefulSetReady`), the agent renders as **Not ready** while the tooltip shows that condition's own message, i.e. "Deployment is ready". A healthy fleet then reads as broken.

Keying on the reason is right for the `Ready=Unknown`/`DeploymentNotFound` case the test covers, but that case is already excluded by `status !== 'True'`. Consider inverting it — trust `Ready=True` unless the reason is a known *not*-ready reason — or at minimum make the unknown-reason case fall back to `Ready`'s status rather than silently contradicting it.

plugins/agent-platform/src/components/AgentsDataProvider/helpers.ts
```diff
@@ -1,8 +1,89 @@
...
+ * while installations whose agents are all ready never leave the baseline.
+ *
+ * Note that interval refetches only fire while the tab is focused
+ * (`refetchIntervalInBackground` defaults to `false`), which is the same
+ * guard kagent's UI implements by hand with `document.hidden`.
+ */
+export function getAgentsRefetchInterval(
+  query: Query<KubeObjectInterface[]>,
+): number {
+  const items = query.state.data;
```
marians, line 71: The doc above ("the installation an agent was just created on tightens to the fast interval") doesn't hold, because the decision is made from data that predates the new agent.

Scenario: user creates an agent on installation X and lands on the list. X's cached list still contains only the old, all-ready agents, so this returns the 60s baseline; nothing in the plugin calls `invalidateQueries`/`refetch` (grep: no invalidation anywhere in `plugins/agent-platform/src`), and `staleTime: 60_000` suppresses a mount refetch of a cache entry under a minute old. So the new agent can take up to 60s to appear, and only *after* that does the fast tier engage. The fast tier effectively only helps agents that were already visible when they started converging.

Either reword the comment, or invalidate the Agent list query after a successful creation so the first fast poll starts immediately.

plugins/agent-platform/src/components/AgentsTable/AgentsTable.tsx
```diff
@@ -112,11 +125,22 @@ export function AgentsTable({ rows }: AgentsTableProps) {
...
+  // Client-side sorting, same shape as SessionsTable. The initial sort is
+  // installation-then-name, which is the ordering this list had before it was
+  // sortable (see `sortAgentsBy`), so enabling sorting doesn't change the default
+  // view. Sorting by Status once puts the agents needing attention on top.
+  const { tableProps } = useTable<AgentRow>({
+    mode: 'complete',
+    data: rows,
+    sortFn: sortAgentsBy,
+    initialSort: { column: 'installation', direction: 'ascending' },
+    paginationOptions: { pageSize: 25, pageSizeOptions: [25, 50, 100] },
```
marians: Enabling pagination in a list that now polls exposes an unclamped offset. `useCompletePagination` only resets `offset` when the page size or the query (sort/filter/search) changes — never when `data` shrinks: `paginatedData = processedData.slice(offset, offset + pageSize)`.

Scenario: 30 agents, user is on page 2 (`offset = 25`), someone deletes agents elsewhere (or an installation drops out of the reachable set, which prunes its cached rows) so the next poll leaves 24. The slice is empty, so the table swaps to "No agents found." while 24 agents exist — recoverable only by pressing "previous". Worth clamping the offset against `rows.length` (or resetting it when the row count drops below the current offset).

plugins/agent-platform/src/components/AgentsDataProvider/AgentsDataProvider.tsx
```diff
@@ -68,11 +73,15 @@ export function AgentsDataProvider({ children }: { children: ReactNode }) {
...
+  // The refetch interval is two-tier and evaluated per installation — see
+  // `getAgentsRefetchInterval`. It keeps agent readiness fresh without polling
+  // the whole fleet fast: only installations with an agent that is still
+  // converging get the short interval.
   const { resources, clustersData, isLoading, errors } = useResources(
     reachableInstallations,
     Agent,
     {},
-    { enableDiscovery: false },
+    { enableDiscovery: false, refetchInterval: getAgentsRefetchInterval },
```
marians, line 84: Worth checking the interaction with `PersistQueryClientProvider`: every successful poll mutates the query cache, and the persister dehydrates and writes the *whole* agent-platform cache to `window.localStorage` (throttled at 1s by `createAsyncStoragePersister`'s default). So while any installation is in the 5s tier, the full fleet cache is `JSON.stringify`'d and written synchronously on the main thread every 5s for as long as the tab stays open — and at the baseline it's now a write per installation per minute, where before it was essentially one write per page visit. On a large fleet (Agent lists are also the payload the persistence exists for) that's a recurring main-thread hitch. Bumping the persister's `throttleTime` would bound it cheaply.

marians (commented):
Careful single-pass review of the diff (readiness derivation, polling, sortable table, `StatusLabel`). Verified locally: `yarn tsc` clean, `prettier --check` clean, lint clean for the three packages, and the 76 tests in the touched suites pass.

Six findings, all inline — one medium (the `isLoadingMore` rewrite), one medium-ish robustness issue in `deriveAgentReadiness`, and four low. Nothing blocking correctness of the happy path; the two worth acting on are the "loading more" regression and the hard dependency on `status.observedGeneration`.

PR #2013, opened 2026-08-03
feat(agent-platform): add an agent details page

marians:
### What does this PR do?

Adds an agent details page to the Agent Platform plugin. Clicking an agent in the
list opens `/agent-platform/agents/<installation>/<namespace>/<name>` — all three
segments, because an `Agent` name is only unique within a namespace on one
installation.

Seven sections:

- **Header** — avatar, display name, derived readiness, technical name,
  installation/namespace, creation age, description. A kebab in the shared plugin
  header opens a **manifest dialog** with the Agent CR as read-only YAML, minus
  `metadata.managedFields` and the `last-applied-configuration` annotation. That
  dialog is the escape hatch for what the page does not surface (`deployment`,
  `sandbox`, `a2aConfig`, labels).
- **GitOps** — the shared `GitOpsCard`, when the agent's desired state really is in
  Git (see below).
- **Status** — readiness, the controller's own explanation, an `UnsupportedFeatures`
  warning, a note naming both generations when the status is stale, and every
  condition verbatim. This is what makes a broken agent debuggable without
  `kubectl`, so it leads the page.
- **Configuration** — type, model, installation, namespace, created, the owning
  HelmRelease (linked to its deployment page), and the MCP-server / agent tool
  references. A Muster gateway reference links to muster's Tool Explorer with the
  installation preselected.
- **System prompt** — copyable.
- **Skills** — the same card grid the create flow's picker uses, read-only, plus the
  `ref` each skill is pinned to.
- **Recent sessions** — the newest few, over a single-installation query that shares
  the Sessions tab's cache key.

**Read-only.** Editing an agent means changing the values its HelmRelease renders
from, and a delete must remove only that release and never the `OCIRepository` a
namespace's agents share — so neither is a menu item yet, and neither is "Launch
session". `docs/agent-platform.md` records what is still needed for the write path.

Agents in the list become clickable, and one agent now has a page showing what it
is, whether it works, and what it has been used for. The main gain is
debuggability: a rejected spec's admission-webhook output and every status
condition are readable in the UI instead of requiring `kubectl get agent -o yaml`.

Deliberately **not** shown, because there is no data behind it: sessions all-time,
sessions in the last 30 days, and a success rate. kagent keeps no per-agent counters
and scopes its session list to the caller, so each would be a number invented from
one person's history — wrong by orders of magnitude on a shared agent. The sessions
section says whose sessions it is showing, and changes wording on an installation
whose kagent is not user-scoped rather than claiming ownership it cannot.

Two smaller user-visible fixes come along:

- The `SimpleAccordion` used by the cluster access card and the workload details
  pane gains the bottom padding bui's accordion trigger omits, so an expanded header
  no longer sits flush against its content.
- `GitOpsCard` stops claiming GitOps for resources that are reconciled but not in
  Git — see below.

### How does it look like?

image

image

Three pieces of shared work were needed, and each removes a duplicate:

**GitOps provenance is de-duplicated.** `readProvenance` / `isGitOpsManaged` /
`provenanceReleaseId` (from `muster`) and `isManagedByFlux` with the Kustomization
label readers (from `flux-react`) now live in one module in `kubernetes-react`,
alongside new `getHelmReleaseName`/`getHelmReleaseNamespace`. `flux-react` and
`muster` re-export from it, so their public APIs are unchanged.

The distinction mattered: `isManagedByFlux` is **false** for a kagent `Agent`, which
is rendered by a Helm chart and so carries `helm.toolkit.fluxcd.io/*` rather than
Kustomization labels. `GitOpsCard` therefore gained a hop — with no Kustomization
label of its own it resolves the owning `HelmRelease` and follows _its_ labels to
the Kustomization and GitRepository — and now takes any `KubeObject` as `resource`
instead of an `App`/`HelmRelease` as `deployment`.

It also stops equating "reconciled by Flux" with "GitOps-managed": where that chain
ends without a Kustomization it renders **nothing**. An agent created through this
plugin is exactly that case — the create flow applies its `HelmRelease` and
`OCIRepository` through the scaffolder, so Flux reconciles the agent but no file in
Git describes it, and saying otherwise sends the reader looking for something that
does not exist. The gs cluster and deployment pages gate on `isManagedByFlux` and so
always have a Kustomization already; their behaviour is unchanged. `GitOpsCard` had
no tests and now has them for both provenance paths.

**New in `ui-react`: `SimpleAccordion`.** One collapsible section, carrying the
bottom padding bui's accordion trigger lacks. Four places had composed bui's
accordion primitives directly and fixed that locally in three different ways — one
with a selector scoped to the trigger element rather than the button that actually
carries `aria-expanded`, so it matched nothing. gs's own `SimpleAccordion` is
removed in favour of this one and its two call sites import from `ui-react`
directly. `plans`' `MergedTab`, agent-platform's `TimelineEntry` and muster's
`DisclosureAccordion` still carry their own copies and could adopt it.

**New in `ui-react`: `ConditionsList`.** A bui renderer for a resource's status
conditions: one collapsible entry per condition with its type, satisfaction,
relative transition time, reason and message; newest transition first, and the first
failing condition expanded. Takes an `isFailing` override for abnormal-true
conditions (`Stalled`, `UnsupportedFeatures`). gs's MUI v4 `HelmReleaseConditions`
is a candidate to adopt it later; that migration is not in this PR.

Also worth knowing:

- The agent is fetched with a single targeted `useResource`, not read out of the
  list's provider, so a deep link works without the list having loaded. It polls on
  the same two tiers as the list (`isAgentConverging` is now shared): 5 s while
  converging, 60 s once settled or durably broken.
- The model is read by name in the agent's **own namespace**, not through the
  cluster-wide `ModelConfigsProvider` list, which is admin-only — reusing it would
  deny a non-admin the model name on a page they can otherwise read in full. A
  failed read falls back to the bare reference, never to "no model".
- `getTelemetryPageViewPayload` gained an `Agent detail` case, deliberately with no
  `view`: the path segments name an installation and an agent, so echoing them would
  emit a page name per agent and record which installations a user reads.

Verified against real agents on `gazelle` and `graveler`, including a genuinely
_Not accepted_ agent, the manifest dialog, the Muster link, back-navigation, and the
telemetry payload.

Yes — done in this PR. `docs/agent-platform.md` gains an "agent detail page"
section covering the layout, the sections, the targeted model read, the GitOps
semantics and what the page cannot show, and its open-TODO list is narrowed to the
write path that is still missing.

### DIFF
--- a/plugins/agent-platform/src/components/AgentDetailPage/AgentConfigurationCard.tsx
+++ b/plugins/agent-platform/src/components/AgentDetailPage/AgentConfigurationCard.tsx
@@ -136,9 +136,13 @@ function ToolsValue({ agent }: { agent: Agent }) {
 
   return (
     <Flex direction="column" gap="3">
-      {mcpServers.map(serverRef => (
+      {mcpServers.map((serverRef, index) => (
         <McpServerRow
-          key={mcpServerRefId(serverRef)}
+          // `namespace/name` is not unique: nothing stops two entries referencing
+          // the same server with different `toolNames`/`requireApproval`, which is
+          // how you express "these tools need approval, those don't" for one
+          // server. Same reasoning as the skill cards.
+          key={`${mcpServerRefId(serverRef)}#${index}`}
           server={serverRef}
           installation={agent.cluster}
         />
--- a/plugins/agent-platform/src/components/AgentDetailPage/AgentConfigurationCard.tsx
+++ b/plugins/agent-platform/src/components/AgentDetailPage/AgentConfigurationCard.tsx
@@ -1,0 +1,265 @@
+import { ReactNode } from 'react';
+import { Box, Flex, Text } from '@backstage/ui';
+import { Link } from '@backstage/core-components';
+import { useRouteRef } from '@backstage/frontend-plugin-api';
+import {
+  Agent,
+  AgentMcpServerRef,
+  getHelmReleaseName,
+  getHelmReleaseNamespace,
+  ModelConfig,
+} from '@giantswarm/backstage-plugin-kubernetes-react';
+import {
+  DateComponent,
+  InfoCard,
+  NotAvailable,
+  StructuredMetadataList,
+} from '@giantswarm/backstage-plugin-ui-react';
+
+import {
+  agentDetailRouteRef,
+  deploymentDetailsExternalRouteRef,
+  musterToolExplorerExternalRouteRef,
+} from '../../routes';
+import {
+  describeToolScope,
+  isMusterServerRef,
+  mcpServerRefId,
+} from './helpers';
+
+/** Monospace for identifiers the reader may retype into `kubectl`. */
+const MONO: React.CSSProperties = { fontFamily: 'monospace' };
+
+/**
+ * How the model is described: the ModelConfig's friendly name, with the model id
+ * and provider underneath.
+ *
+ * Falls back to the bare reference when the ModelConfig cannot be read — which is
+ * normal for a non-admin, since ModelConfigs live in namespaces they may not have
+ * access to — rather than implying the agent has no model.
+ */
+function ModelValue({
+  modelConfigName,
+  modelConfig,
+  namespace,
+}: {
+  modelConfigName?: string;
+  modelConfig?: ModelConfig;
+  namespace?: string;
+}) {
+  if (!modelConfigName) {
+    return <NotAvailable />;
+  }
+
+  const modelId = modelConfig?.getModel();
+  const provider = modelConfig?.getProvider();
+
+  return (
+    <Flex direction="column" gap="1">
+      <Text variant="body-medium">
+        {modelConfig?.getDisplayName() ?? modelConfigName}
+      </Text>
+      {(modelId || provider) && (
+        <Text variant="body-small" color="secondary" style={MONO}>
+          {[modelId, provider].filter(Boolean).join(' · ')}
+        </Text>
+      )}
+      <Text variant="body-small" color="secondary">
+        ModelConfig{' '}
+        <span style={MONO}>
+          {namespace ? `${namespace}/${modelConfigName}` : modelConfigName}
+        </span>
+      </Text>
+    </Flex>
+  );
+}
+
+/**
+ * One MCP server the agent draws tools from.
+ *
+ * The prop is `server`, not `ref` — React intercepts a `ref` prop rather than
+ * passing it through.
+ */
+function McpServerRow({
+  server: serverRef,
+  installation,
+}: {
+  server: AgentMcpServerRef;
+  installation: string;
+}) {
+  const toolExplorerRoute = useRouteRef(musterToolExplorerExternalRouteRef);
+
+  // Preselect the installation the agent runs on, the way muster's own
+  // cross-links do. Only offered for the Muster gateway: the Tool Explorer talks
+  // to muster, so it can say nothing about any other MCP server.
+  const musterLink =
+    isMusterServerRef(serverRef) && toolExplorerRoute
+      ? `${toolExplorerRoute()}?installation=${encodeURIComponent(installation)}`
+      : undefined;
+
+  return (
+    <Flex direction="column" gap="1">
+      <Flex align="center" gap="2" style={{ flexWrap: 'wrap' }}>
+        <Text variant="body-medium" style={MONO}>
+          {serverRef.kind ?? 'RemoteMCPServer'} {mcpServerRefId(serverRef)}
+        </Text>
+        {musterLink && <Link to={musterLink}>Explore tools</Link>}
+      </Flex>
+      <Text variant="body-small" color="secondary">
+        {describeToolScope(serverRef)}
+      </Text>
+      {/* Rare, and the only per-tool setting that changes what a user will
+          experience mid-session, so it is worth naming. */}
+      {serverRef.requireApproval && serverRef.requireApproval.length > 0 && (
+        <Text variant="body-small" color="secondary">
+          Requires approval: {serverRef.requireApproval.join(', ')}
+        </Text>
+      )}
+    </Flex>
+  );
+}
+
+/** The tools block: MCP servers, then any agents invoked as tools. */
+function ToolsValue({ agent }: { agent: Agent }) {
+  const agentDetailRoute = useRouteRef(agentDetailRouteRef);
+  const mcpServers = agent.getMcpServerRefs();
+  const agentRefs = agent.getAgentRefs();
+
+  if (mcpServers.length === 0 && agentRefs.length === 0) {
+    return (
+      <Text variant="body-small" color="secondary">
+        This agent declares no tool servers, so it has no tools beyond its own
+        reasoning.
+      </Text>
+    );
+  }
+
+  return (
+    <Flex direction="column" gap="3">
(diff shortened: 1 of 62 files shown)


plugins/agent-platform/src/components/AgentDetailPage/AgentDetailPage.tsx
```diff
@@ -0,0 +1,270 @@
...
+          description={`No agent named "${name}" exists in namespace "${namespace}" on ${
+            installation || 'that installation'
+          }. It may have been deleted or renamed, or kagent may not be installed there.`}
+          action={<BackToAgents>Back to agents</BackToAgents>}
+        />
+      </Content>
+    );
+  }
+
+  if (error || !agent) {
```
marians: A transient failure of a **background poll** blanks the whole page even though good data is cached.

`useResource` returns react-query's `error`, and in v5 a failed refetch sets `status: 'error'` while keeping `data`. This page polls every 5 s while converging (`getAgentRefetchInterval`), and the plugin's `QueryClientProvider` explicitly does **not** retry `ServiceUnavailableError` / `UnauthorizedError` / `ForbiddenError`. So: user is watching an agent come up, one poll hits a k8s-proxy hiccup (503) or an expired token → `error` is truthy, `agent` is still the fully-rendered previous object, and the entire page is replaced by the red "Could not load this agent" alert until the *next* successful poll (up to 60 s once it has backed off).

The same applies to the `errors.some(isNotFoundError)` branch above (line 117): one 404 during polling swaps the page for "Agent not found" while the agent is still in hand.

Note the comment on `staleTime` in `QueryClientProvider.tsx` calls out exactly this hazard for the lists ("the refetch is what could error on a single cluster and drop its rows"). Suggest keeping the rendered page and surfacing the failure non-destructively (e.g. via `useShowErrors`, which this page already has an `ErrorsProvider` for) when data is present:

```suggestion
  if (!agent) {
```
…with the `error` message still used for the description, and the not-found branch gated on `!agent` too.

plugins/agent-platform/src/hooks/useAgentSessions.ts
```diff
@@ -0,0 +1,102 @@
...
+        .filter(session => session.agentId === agentId)
+        .map(session => toSessionRow(session, agentIndex)),
+    );
+  }, [sessions, agentId, agentIndex]);
+
+  return {
+    rows,
+    isLoading,
+    isNotUserScoped: capabilities.isUserScoped === false,
+    isUnavailable: isError,
```
marians: Same "error wins over cached data" pattern as the page-level branch: `isError` stays `true` after a *failed refetch* of a query that already holds data, so `rows` is non-empty and correct while `AgentSessionsCard` renders "Sessions could not be read from this installation." instead of the table.

Concrete scenario: the user opens the Sessions tab (populating this exact query key), navigates to an agent 90 s later — the entry is stale, so mount triggers a background refetch; kagent answers 503, which the plugin's `QueryClientProvider` does not retry. `sessions` still holds the previously fetched list, `rows` has the agent's sessions, but the card hides them behind an "unavailable" message.

Suggest only claiming unavailability when there is genuinely nothing to show:

```suggestion
    isUnavailable: isError && !sessions,
```

plugins/flux-react/src/components/GitOpsCard/GitOpsCard.tsx
```diff
@@ -71,15 +126,33 @@ export function GitOpsCard({ deployment, installationName }: GitOpsCardProps) {
...
+      ...helmReleaseErrors,
+      ...kustomizationErrors,
+      ...gitRepositoryErrors,
+    ];
+  }, [gitRepositoryErrors, helmReleaseErrors, kustomizationErrors]);
 
   useShowErrors(errors);
 
   let errorMessage;
+  if (helmReleaseError) {
```
marians: This branch is unreachable, so the HelmRelease-hop failure never gets reported in the card.

If the HelmRelease read fails, `ownerHelmRelease` is `undefined` → `kustomizationOwner` is `undefined` → `kustomizationName` is `undefined` → the new `if (!kustomizationName) return null;` at line 193 fires before anything renders. So `errorMessage` is computed here and then thrown away with the whole component.

Concrete scenario: a user who can read `Agent`s but has no RBAC on `helm.toolkit.fluxcd.io/HelmRelease` in that namespace (or whose owning HelmRelease has been deleted) opens an agent detail page. What they get is a global error notice from `useShowErrors(errors)` — "Failed to load HelmRelease …" — with *no* card and no explanation of why, which is the opposite of the intent of both this block and the `return null` (which is meant to communicate "not in Git", not "we failed to look").

Two things worth deciding explicitly: (a) whether the hop's errors should reach `useShowErrors` at all, given the card is purely informational and the enclosing page already renders fine without it, and (b) whether a *failed* hop should render the card in its error state rather than nothing — right now "we could not tell" is indistinguishable from "definitely not GitOps-managed".

plugins/agent-platform/src/components/AgentDetailPage/AgentConfigurationCard.tsx
```diff
@@ -0,0 +1,265 @@
...
+        reasoning.
+      </Text>
+    );
+  }
+
+  return (
+    <Flex direction="column" gap="3">
+      {mcpServers.map(serverRef => (
+        <McpServerRow
+          key={mcpServerRefId(serverRef)}
```
marians: `mcpServerRefId(serverRef)` is not unique across `spec.declarative.tools` entries — nothing in the CRD stops two entries from referencing the same `namespace/name` with different `toolNames` / `requireApproval`, which is in fact the natural way to express "these tools need approval, those don't" for one server. Both rows then get the same React key: a duplicate-key warning and reconciliation that can mismatch the two rows' content.

`AgentSkillsCard` already guards against exactly this (`key={`${skill.url}#${skill.path ?? ''}#${index}`}` with a comment saying nothing in a gitRef is unique); the same reasoning applies here.

```suggestion
          key={`${mcpServerRefId(serverRef)}#${index}`}
```
(requires taking `index` from the `.map` callback)

marians (commented):
Careful single-pass review of the whole diff (agent detail page, the `provenance` de-duplication, `GitOpsCard`'s HelmRelease hop, `SimpleAccordion`/`ConditionsList` in `ui-react`, the table/route/telemetry changes).

Verified as correct and not commented on: the telemetry case ordering and its regex (three segments, sits above the generic `/agent-platform` cases, cannot swallow `new`/`new/skills`/`new/review`); the router's segment-count reasoning; `getAgentRefetchInterval`'s cast (`useResource`'s `queryInfo.data` really is the raw `KubeObjectInterface`, and `Query<KubeObjectInterface>` matches `QueryOptions`); the `deploymentDetails` external route params and the lowercase `kind: 'helmrelease'` that `useDeploymentFromUrl` compares against; `?installation=` being the param `MusterInstanceProvider` reads; `provenance.ts` being a verbatim move of both the flux-react and muster helpers; `paginationOptions: { type: 'none' }` being valid for bui's `useTable`; `className` on `AccordionGroup` being merged (not replaced) by `useDefinition`, and `defaultExpandedKeys` reaching RAC's `DisclosureGroup`; and `useId()` working as a `Disclosure` key.

Four findings, all inline. The first two are the same shape and are the ones I'd want fixed: an error from a *background* refetch is treated as "we have nothing", so a single un-retried 503/401 hides data that is already loaded and correct.

PR #2017, opened 2026-08-04
fix(muster): keep the MCP Servers index redirect from being clobbered

marians:
### What does this PR do?

Fixes the Agent Platform → **MCP Servers** section landing on a blank view with no second-level tab selected.

`MusterInstanceProvider` writes the active installation into `?installation=` from an effect. That is a *search-only* navigation, so react-router resolves it against the pathname of the render that created it. Once the muster installations query is cached (i.e. from the second visit to the tab within a session), the effect runs in the same commit as the section's index redirect, and its write lands on the pre-redirect path — replacing `/agent-platform/muster/dashboard` back with `/agent-platform/muster`, where no view route matches.

The index redirect and the legacy `workflows/:name/run` redirect are now mounted as siblings of the views, outside `MusterProviders`, so the installation-param write happens in a later commit and can no longer overwrite them. `MusterInstanceProvider` carries a note about the constraint so a redirect isn't re-introduced under it.

Reproduced against production (`v0.179.5`); the history trace on a tab click was:

```
push    /agent-platform/muster
replace /agent-platform/muster/dashboard          <- the index redirect
replace /agent-platform/muster?installation=gazelle  <- clobbers it
```

Clicking the "MCP Servers" tab always lands on the muster Dashboard view with the Dashboard tab highlighted, not on an empty page.

Found while investigating two separate tab-navigation reports. The other one — every entity-page tab href getting the active tab segment prefixed, so "Overview" looks dead — is an upstream bug in `@backstage/plugin-catalog`'s bui entity header (relative tab hrefs + react-router 7 relative-splat resolution) and is not addressed here.

New `MusterSection.test.tsx` covers both redirects; the warm-cache case fails on `main` and passes with this change.

No.

### DIFF
--- a/plugins/muster/src/components/MusterSection/MusterSection.tsx
+++ b/plugins/muster/src/components/MusterSection/MusterSection.tsx
@@ -1,7 +1,15 @@
-import { Navigate, Route, Routes } from 'react-router-dom';
+import {
+  Navigate,
+  Route,
+  Routes,
+  useLocation,
+  useParams,
+} from 'react-router-dom';
 import { Box, Tab, TabList, Tabs } from '@backstage/ui';
+import { useRouteRef } from '@backstage/frontend-plugin-api';
 import { useSplatBasePath } from '@giantswarm/backstage-plugin-ui-react';
 
+import { workflowDetailRouteRef } from '../../routes';
 import { MusterProviders } from '../MusterProviders';
 import { DashboardPage } from '../DashboardPage';
 import { McpServersPage } from '../McpServersPage';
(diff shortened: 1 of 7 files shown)


plugins/muster/src/components/MusterSection/MusterSection.tsx
```diff
@@ -63,3 +83,24 @@ export const MusterSection = () => {
...
+// created in. Mounted alongside the redirect, that write lands on the
+// pre-redirect path and silently replaces `/muster/dashboard` back with
+// `/muster`, leaving the section with no view and no selected tab. It only shows
+// once the installations query is cached (i.e. from the second visit in a
+// session), because a pending query makes the effect bail out and lose the race.
+// Redirecting before the providers mount keeps the two writes in separate
+// commits. Same reason the legacy `workflows/:name/run` redirect lives here.
+export const MusterSection = () => (
+  <Routes>
+    <Route index element={<Navigate to="dashboard" replace />} />
```
marians: The index redirect drops the query string, so a deep link to the section root with an explicit installation loses it. Verified with a throwaway test (two installations `gazelle`, `alpha`): rendering `/agent-platform/muster?installation=alpha` ends up at `/agent-platform/muster/dashboard?installation=gazelle` — the requested `alpha` is silently replaced by the default. Reason: the redirect creates a location without `search`, so by the time the installations query resolves `MusterInstanceProvider` no longer sees `urlInstallation` and falls back to localStorage/first entry.

Pre-existing (main behaves identically), but this PR now owns the redirect, and unlike before it is trivially safe to fix here: the redirect runs before `MusterProviders` mounts, so preserving the search cannot re-introduce the clobber. Note the sibling `LegacyRunRedirect` already preserves `search` for exactly this reason.

```suggestion
    <Route index element={<IndexRedirect />} />
```

with e.g.

```tsx
const IndexRedirect = () => {
  const { search } = useLocation();
  return <Navigate to={{ pathname: 'dashboard', search }} replace />;
};
```

plugins/muster/src/components/MusterSection/MusterSection.tsx
```diff
@@ -19,14 +27,27 @@ const VIEWS = [
...
+ * The bespoke `/workflows/:name/run` route was removed when Run was unified with
+ * the tool explorer; a lingering deep link used to silently resolve to the full
+ * workflows list. Redirect it to the workflow detail (preserving the query
+ * string, e.g. `?installation=`) so the named workflow is not dropped.
+ */
+const LegacyRunRedirect = () => {
+  const { name = '' } = useParams();
+  const { search } = useLocation();
+  const detailLink = useRouteRef(workflowDetailRouteRef);
+  const to = detailLink ? detailLink({ name }) : '..';
```
marians: Minor: moving this component up a level silently changes the meaning of the `'..'` fallback. Previously the redirect was matched at `:name/run` under a `<Routes>` based at `/agent-platform/muster/workflows`, so `..` resolved to the workflows list. Now it is matched at `workflows/:name/run` under a `<Routes>` based at `/agent-platform/muster`, so `..` resolves to the section root — which the new sibling index route immediately redirects to `/dashboard`. So if `useRouteRef(workflowDetailRouteRef)` ever returns undefined (route not bound/mounted), the legacy deep link lands on the Dashboard instead of at least the Workflows list. Consider `'../workflows'` (or `workflowsRouteRef`) for the fallback.

marians (commented):
Reviewed the routing change closely and verified it behaves as claimed. Ran `MusterSection.test.tsx` on the PR head (3/3 pass) and separately checked, with a throwaway test, that:

- tab hrefs are still correct under the extra `<Routes>` layer (`useSplatBasePath` in `MusterViews` sees the same `*` remainder as before — e.g. at `/agent-platform/muster/workflows/my-flow` the hrefs are still `/agent-platform/muster/{dashboard,servers,workflows,tools}`), and
- route ranking works out: `index` and `workflows/:name/run` both win over the `*` catch-all, so `MusterViews` is not remounted per view change.

No correctness problems found with the fix itself. Two small notes inline, both low severity.

PR #2018, opened 2026-08-04
fix(catalog): make entity page tab links work from non-Overview tabs

marians:
### What does this PR do?

Fixes entity page header tabs: from any tab other than **Overview**, every tab link was wrong.

| on `…/component/backstage/circleci` | link before | link after |
| --- | --- | --- |
| Overview | `…/backstage/circleci` (the current URL → click does nothing) | `…/backstage` |
| Deployments | `…/backstage/circleci/deployments` | `…/backstage/deployments` |
| Dependencies | `…/backstage/circleci/dependencies` | `…/backstage/dependencies` |

Clicking one of the others didn't just misnavigate — `…/circleci/dependencies` still matches the `circleci/*` content route, so you land on a bogus URL showing the *previous* tab's content.

**Cause (upstream):** `useEntityTabs`/`buildHeaderTabs` in `@backstage/plugin-catalog` build the tab hrefs relative (`''`, `'deployments'`, …). The bui `Header` renders them through react-aria's `RouterProvider` (mounted by `@backstage/plugin-app`'s `AppRoot`), which resolves them with react-router's `useHref` — and react-router 7 resolves a relative link inside a splat route against the **full current pathname**, not the route's base. The entity page is mounted at `/catalog/:namespace/:kind/:name/*`, so the active tab segment gets prefixed onto every link. Verified on production `v0.179.5` by reading the rendered hrefs and the React props (`tabs` carries `href: 'deployments'`; the anchor gets `…/circleci/deployments`).

**Why a patch and not an override:** the supported override point, `EntityHeaderLayoutBlueprint`, only receives `{ tabs, activeTabId }`, and supplying a `HeaderComponent` makes `EntityLayoutBui` skip `EntityHeaderBui` altogether — we'd re-implement title/tags/lifecycle/owner/part-of metadata and the favourite button, and lose the entity context menu (`contextMenuItems` never reach a custom header, and `EntityContextMenu` isn't exported). The patch is ~40 lines in one dist file, keeps the upstream header intact, and `postinstall` fails loudly if a Backstage bump changes that code.

Upstream `master` still emits relative hrefs, so there is no version to bump to yet. `patches/README.md` (new) records what each of our three patches does and when it can be dropped, plus the note that `patch-package --create` cannot be used in this repo (its pristine temp install uses yarn classic, which can't resolve our `backstage:^` specifiers — patches have to be hand-diffed and verified with an apply run).

Entity page tabs navigate correctly from any tab. In particular "Overview" works instead of appearing dead.

### How does it look like?

Verified against a local dev server on this branch, walking tab → tab without passing through Overview:

```
Overview        -> /catalog/default/component/backstage           (Overview active)
Dependencies    -> /catalog/default/component/backstage/dependencies
Version History -> /catalog/default/component/backstage/version-history
Overview        -> /catalog/default/component/backstage
GitHub Actions  -> /catalog/default/component/backstage/github-actions
Deployments     -> /catalog/default/component/backstage/deployments
```

Also checked from a deep nested sub-route (`…/circleci/12345/step/1`): all tab hrefs still resolve to the entity base and CircleCI stays the active tab.

Second of two tab-navigation bugs reported together; the first (muster's MCP Servers section landing on a blank view) was #2017 and is unrelated.

No — `patches/README.md` is added as part of this change.

- [ ] No changeset: no published `@giantswarm/backstage-plugin-*` package changes, only a dependency patch consumed by the app.

### DIFF
--- a/patches/@backstage+plugin-catalog+2.0.7.patch
+++ b/patches/@backstage+plugin-catalog+2.0.7.patch
@@ -7,7 +7,7 @@ index f3fb7e0..1abb9fa 100644
 +import { useRouteRefParams } from '@backstage/core-plugin-api';
 +import { entityRouteRef, useEntityRefLink } from '@backstage/plugin-catalog-react';
  import { useSelectedSubRoute } from './EntityTabs.esm.js';
- 
+
 +// GIANT SWARM PATCH -- see patches/README.md.
 +//
 +// buildHeaderTabs below builds *relative* tab hrefs ("", "deployments", ...).
--- a/patches/@backstage+plugin-catalog+2.0.7.patch
+++ b/patches/@backstage+plugin-catalog+2.0.7.patch
@@ -1,0 +1,68 @@
+diff --git a/node_modules/@backstage/plugin-catalog/dist/alpha/components/EntityTabs/useEntityTabs.esm.js b/node_modules/@backstage/plugin-catalog/dist/alpha/components/EntityTabs/useEntityTabs.esm.js
+index f3fb7e0..75438d4 100644
+--- a/node_modules/@backstage/plugin-catalog/dist/alpha/components/EntityTabs/useEntityTabs.esm.js
++++ b/node_modules/@backstage/plugin-catalog/dist/alpha/components/EntityTabs/useEntityTabs.esm.js
+@@ -1,6 +1,47 @@
+ import { useMemo } from 'react';
++import { useLocation, useParams } from 'react-router-dom';
+ import { useSelectedSubRoute } from './EntityTabs.esm.js';
+ 
++// GIANT SWARM PATCH -- see patches/README.md.
++//
++// buildHeaderTabs below builds *relative* tab hrefs ("", "deployments", ...).
++// The bui Header renders them through react-aria's RouterProvider, which resolves
++// them with react-router's useHref -- and react-router 7 resolves a relative link
++// inside a splat route (the entity page is mounted at
++// "/catalog/:namespace/:kind/:name/*") against the full current pathname rather
++// than the route's base. So on any tab other than Overview every tab href gets
++// the active tab segment prefixed (".../my-component/circleci/deployments"), and
++// Overview's "" resolves to the current URL -- clicking it does nothing.
++//
++// Resolve the hrefs against the entity's own base path instead. Drop this patch
++// once upstream emits absolute tab hrefs.
++function useEntityBasePath() {
++  const { pathname } = useLocation();
++  const params = useParams();
++  // Compare by segment count: pathname stays percent-encoded while the splat
++  // param is decoded, so lengths would differ on encoded characters -- the "/"
++  // count does not.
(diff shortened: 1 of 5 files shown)


patches/@backstage+plugin-catalog+2.0.7.patch
```diff
@@ -0,0 +1,68 @@
...
++// once upstream emits absolute tab hrefs.
++function useEntityBasePath() {
++  const { pathname } = useLocation();
++  const params = useParams();
++  // Compare by segment count: pathname stays percent-encoded while the splat
++  // param is decoded, so lengths would differ on encoded characters -- the "/"
++  // count does not.
++  const splatSegments = (params["*"] ?? "").split("/").filter(Boolean).length;
++  const segments = pathname.split("/").filter(Boolean);
++  return `/${segments.slice(0, Math.max(0, segments.length - splatSegments)).join("/")}`;
```
marians: `basePath` can be over-truncated when a splat segment contains a percent-encoded slash. The comment above says segment counting is safe because `pathname` stays encoded — but that is exactly the problem for `%2F`: `/catalog/default/component/foo/docs/a%2Fb` has 6 pathname segments, while react-router decodes the splat param to `docs/a/b` (3 segments), so 3 segments are stripped instead of 2 and `basePath` becomes `/catalog/default/component` — every tab href then points at a non-existent entity path. In the extreme (splat with enough encoded slashes that `splatSegments >= segments.length`) `Math.max(0, …)` yields `basePath === "/"`, and `absoluteTabHref` then produces `//deployments`, which `@backstage/ui`'s `isExternalLink` classifies as a protocol-relative external URL (`node_modules/@backstage/ui/dist/utils/linkUtils.esm.js`) — clicking a tab navigates off-site to `http://deployments/`. Worth clamping the strip count to `segments.length - 1` (or bailing out to the unpatched relative behaviour) and guarding the trailing-slash join. Note the same flaw exists in the in-repo `useSplatBasePath` (plugins/ui-react/src/hooks/useSplatBasePath.ts:17), so fixing both keeps them in sync.

patches/README.md
```diff
@@ -0,0 +1,50 @@
+# Dependency patches
+
+Patches in this directory are applied to `node_modules` by
+[`patch-package`](https://github.com/ds300/patch-package) from the root
+`postinstall` script. A patch that no longer applies **fails the install**, so a
```
marians: "A patch that no longer applies **fails the install**" only holds in CI. patch-package deliberately exits 0 on a failed apply outside CI (`node_modules/patch-package/dist/index.js:91`: `shouldExitWithError = --error-on-fail || (isCI && …) || NODE_ENV=test`), so a developer running `yarn install` after a Backstage bump locally gets a red log line but a successful install — and then silently loses the tab fix, which is easy to miss since the symptom is the pre-existing broken navigation. Either soften the wording or make it true by using `yarn patch-package --error-on-fail` in the `postinstall` script.

PR #2024, opened 2026-08-05
feat(agent-platform): delete an agent from the details page

marians:
### What does this PR do?

Adds the companion delete to agent creation. The agent details page's kebab menu
gains **Delete agent…**, which removes the agent's `HelmRelease` — the object that
owns its existence, since helm-controller would otherwise just re-render the `Agent`
CR. Deleting it makes helm-controller uninstall the release and take the agent with
it. The owner is resolved from the Flux provenance labels rather than by assuming the
release is named after the agent, so this also works for agents created outside the
wizard.

**The delete is only offered to users who may perform it.** A
`SelfSubjectAccessReview` for `delete` on that named `HelmRelease` in its namespace
runs before the menu item renders, and the item is withheld while the review is still
in flight, so it never appears and then disappears. The review decides what is
_shown_; authorization stays the apiserver's, since the Kubernetes proxy forwards the
user's own OIDC token — a bypassed menu item still gets a real 403, surfaced in the
dialog. Two further conditions hide it: no resolvable owner (an `Agent` applied by
hand has no release to delete), and a release applied by a Kustomization, whose
desired state is in Git and which would be recreated on the next reconciliation.

**The namespace's shared `OCIRepository` is removed only when provably unused.**
Every agent in a namespace shares one `agent` chart source, so the `HelmRelease`s in
the source's namespace are listed first, and any other release referencing the same
object keeps it. A failed list read keeps it too — an empty list because the read
failed is not the same answer as an empty list because nothing references it. A
failure to delete the source is swallowed: the agent is gone, which is what was asked
for, and an unreferenced chart source is inert and re-applied identically by the next
agent creation. That is also why the permission gate does not require `delete` on
`ocirepositories` — requiring it would hide the action from users whose real deletion
would succeed.

Two new shared primitives, both firsts for this repo:

- **`ConfirmDialog` in `ui-react`** — the first reusable modal. Controlled rather than
  wrapping a `DialogTrigger`, which is the only thing that works when the trigger is a
  `MenuItem` (react-aria unmounts the menu on selection, taking any trigger inside it
  along). Confirming deliberately does **not** close it: an action that can fail needs
  somewhere to report that, and a dialog that dismisses itself on confirm has thrown
  away the only place the user was still looking. While busy it is undismissable, so a
  stray click or Escape cannot orphan an in-flight request.
- **`deleteResource` in `kubernetes-react`** — the first Kubernetes `DELETE` in the
  app. Same proxy, same trailing-slash stripping, same `ForbiddenError`/`NotFoundError`
  naming as `patchResource`, so an idempotent caller can treat a 404 as success. The
  details both verbs need moved into a shared `k8sMutation` module;
  `BACKSTAGE_FIELD_MANAGER` keeps its existing import path.

An agent can be deleted from its details page, by users whose cluster RBAC allows it.
On success they land back on the agents list with a toast that says "Deleting", not
"Deleted": the `HelmRelease` has a finalizer, so all that is certain is that the
apiserver accepted the request, and the agent can still be in the list for a few
seconds. On failure the dialog stays open with the message inline, and no toast —
the user is still looking at the modal they pressed Delete in.

The confirmation modal says exactly one thing, because it is the only thing the
person clicking cannot work out for themselves:

> This ends any session currently running with this agent — including sessions started
> by other people, which are not shown to you.

kagent scopes its session list to the caller, so a quiet sessions list is not evidence
that an agent is idle. Everything mechanical is deliberately kept out of the modal
(which `HelmRelease` goes, what happens to the shared chart source, that a _suspended_
release is not uninstalled at all) — true, and all of it noise at the moment of
deciding. It is documented in `docs/agent-platform.md` instead.

Toasts use `toastApiRef` from `@backstage/frontend-plugin-api` rather than the
deprecated `alertApi`, which now emits a runtime deprecation warning and is scheduled
for removal upstream.

### How does it look like?

The kebab on the agent details page, and the confirmation:

```
Delete agent "Issue Tracker"?

This ends any session currently running with this agent — including sessions
started by other people, which are not shown to you.

                                          [ Cancel ]  [ Delete agent ]
```

The constraint that shaped this is documented in `docs/agent-platform.md` under
"Shared OCIRepository per namespace": one shared `OCIRepository/agent` per namespace,
many `HelmRelease`s. `AgentActionsMenu`'s docstring previously named this as the
reason no delete existed yet.

Two things worth knowing for review:

- **A bug only the running app caught.** `AgentActionsMenu` is rendered into the shared
  plugin header via `useProvidePageHeaderActions`, which is **outside** the plugin's
  `QueryClientProvider` — so calling a react-query hook there throws "No QueryClient
  set" and takes the page down. The page now calls `useDeleteAgent` and passes the
  result down as a prop. Deliberately _not_ fixed by wrapping the actions element in a
  second provider: that component builds a `new QueryClient` per mount, so the delete
  would have got a private cache and its invalidations would never have reached the
  page's reads. There is now a regression test that renders the menu with no client in
  scope.
- **Sessions survive.** An earlier draft of the modal claimed deleting an agent made
  its session history unreachable. It does not: sessions live in kagent's own store
  keyed by `user_id`, not in the `Agent` CR, and both the list and a session's detail
  are fetched by session id — `toSessionRow`'s `decodeAgentIdLabel` fallback exists
  precisely to label a session whose `Agent` is gone. Nor is a re-created agent a clean
  slate, since `agent_id` is derived from `namespace/name` alone. Both claims were
  removed.

Verified end-to-end against a real installation, including that the shared
`OCIRepository` is correctly kept when other agents in the namespace still use it.

Yes, done in this PR — `docs/agent-platform.md` gains a "Deleting an agent" section,
and the passages that described deletion as future work were rewritten.

### DIFF
--- a/plugins/agent-platform/src/hooks/useDeleteAgent.ts
+++ b/plugins/agent-platform/src/hooks/useDeleteAgent.ts
@@ -1,0 +1,280 @@
+import { useCallback, useMemo } from 'react';
+import { useApi } from '@backstage/core-plugin-api';
+import { kubernetesApiRef } from '@backstage/plugin-kubernetes-react';
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+import {
+  Agent,
+  CustomResourceMatcher,
+  deleteResource,
+  HelmRelease,
+  getHelmReleaseName,
+  getHelmReleaseNamespace,
+  isManagedByFlux,
+  OCIRepository,
+  useResource,
+  useResources,
+  useSelfSubjectAccessReview,
+} from '@giantswarm/backstage-plugin-kubernetes-react';
+
+/** A `spec.chartRef` pointing at the chart source a release renders from. */
+type ChartSourceRef = {
+  name: string;
+  namespace: string;
+};
+
+function isSameObject(
+  a: { name?: string; namespace?: string },
+  b: { name?: string; namespace?: string },
+) {
+  return a.name === b.name && a.namespace === b.namespace;
+}
+
+/**
+ * Deleting one agent, and the checks that decide whether to offer it at all.
+ *
+ * An agent is not a thing we can delete directly: it is rendered by the `agent`
+ * chart, so the object that owns its existence is the `HelmRelease`, and removing
+ * that is what makes helm-controller uninstall the release and take the `Agent`
+ * CR with it. The owner is found through the Flux provenance labels rather than
+ * by assuming the release is named after the agent, so this also works for agents
+ * created outside this plugin's wizard.
+ *
+ * The `OCIRepository` the release points at is a different matter: every agent in
+ * a namespace shares one `agent` chart source, so it may only be removed once
+ * nothing else references it. The check is a list of the `HelmRelease`s in the
+ * source's namespace, and every uncertainty resolves to keeping it — an orphaned
+ * chart source is inert and the next agent created in the namespace re-applies an
+ * identical one, while a wrongly deleted one breaks every remaining agent's next
+ * reconciliation.
+ *
+ * Call this from inside the plugin's `QueryClientProvider` — i.e. from the page,
+ * not from the actions element the page hands to `useProvidePageHeaderActions`,
+ * which renders in the shared header outside that provider. `agent` is therefore
+ * optional: the page has none while it is still loading, and a hook cannot be
+ * called conditionally.
+ */
+export function useDeleteAgent(agent: Agent | undefined) {
+  const kubernetesApi = useApi(kubernetesApiRef);
+  const queryClient = useQueryClient();
+
+  const cluster = agent?.cluster ?? '';
+  const helmReleaseName = agent ? getHelmReleaseName(agent) : undefined;
+  const helmReleaseNamespace = agent
+    ? (getHelmReleaseNamespace(agent) ?? agent.getNamespace())
+    : undefined;
+  const hasOwner = Boolean(helmReleaseName);
+
+  // The owning release, for three things: whether Flux owns it declaratively,
+  // which chart source it uses, and the API version to address it at.
+  const { resource: helmRelease, isLoading: isLoadingHelmRelease } =
+    useResource(
+      cluster,
+      HelmRelease,
+      {
+        name: helmReleaseName ?? '',
+        namespace: helmReleaseNamespace,
+        enableDiscovery: false,
+      },
+      { enabled: hasOwner },
+    );
+
+  // A release applied by a Kustomization has its desired state in Git, so
+  // deleting it here would be undone on the next reconciliation — a confusing
+  // no-op dressed up as a deletion. Those agents stay read-only.
+  const isGitOpsOwned = helmRelease ? isManagedByFlux(helmRelease) : false;
+
+  const { allowed: isAllowed, isLoading: isCheckingPermission } =
+    useSelfSubjectAccessReview(
+      cluster,
+      {
+        group: HelmRelease.group,
+        resource: HelmRelease.plural,
+        namespace: helmReleaseNamespace,
+        // Named, so a grant restricted via `resourceNames` answers accurately.
+        name: helmReleaseName,
+        verb: 'delete',
+      },
+      { enabled: hasOwner },
+    );
+
+  const chartRef = helmRelease?.getChartRef();
+  const chartSource: ChartSourceRef | undefined =
+    chartRef?.kind === 'OCIRepository'
+      ? { name: chartRef.name, namespace: chartRef.namespace }
+      : undefined;
+
+  // Everything else in the source's namespace that could be holding on to it.
+  const { resources: namespaceReleases, clustersData } = useResources(
+    cluster,
+    HelmRelease,
+    { [cluster]: { namespace: chartSource?.namespace ?? '' } },
+    { enableDiscovery: false, enabled: Boolean(chartSource) },
+  );
+
+  // `clustersData` only carries clusters whose read succeeded, which is the
+  // distinction that matters here: an empty list because nothing else references
+  // the source is not the same answer as an empty list because the read failed.
+  const hasReadNamespaceReleases = clustersData.some(
+    entry => entry.cluster === cluster,
+  );
+
+  const isChartSourceShared =
+    chartSource !== undefined &&
+    namespaceReleases.some(release => {
+      // Our own release does not count as another user of the source.
+      if (
+        isSameObject(
+          { name: release.getName(), namespace: release.getNamespace() },
+          { name: helmReleaseName, namespace: helmReleaseNamespace },
+        )
+      ) {
+        return false;
+      }
+
+      const ref = release.getChartRef();
+
+      return ref?.kind === 'OCIRepository' && isSameObject(ref, chartSource);
+    });
+
+  // Deliberately not a check on the whole cluster: `HelmRelease`s in other
+  // namespaces can reference this source cross-namespace, but listing them needs
+  // a cluster-wide read a tenant user does not have. Getting this wrong costs a
+  // chart source that the next agent creation re-applies.
+  const willRemoveChartSource =
+    Boolean(chartSource) && hasReadNamespaceReleases && !isChartSourceShared;
+
+  // Read at its served version so the delete addresses a version the cluster
+  // actually has — `OCIRepository` exists as both v1beta2 and v1, so discovery
+  // stays on here. No object means there is nothing left to remove.
+  const { resource: chartSourceResource } = useResource(
+    cluster,
+    OCIRepository,
+    {
+      name: chartSource?.name ?? '',
+      namespace: chartSource?.namespace ?? '',
+    },
+    { enabled: willRemoveChartSource },
+  );
+
+  const invalidateReads = async (gvks: CustomResourceMatcher[]) => {
+    // Prefixes of the keys the read hooks register (see `useListResources` /
+    // `useGetResource`), so one entry per kind covers every list and instance of
+    // it on this cluster. Invalidate rather than editing the cache: the
+    // QueryClient is persisted to localStorage, so a stale pre-deletion object
+    // could otherwise be rehydrated on reload.
+    await Promise.all(
+      gvks.flatMap(gvk =>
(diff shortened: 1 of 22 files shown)


plugins/agent-platform/src/hooks/useDeleteAgent.ts
```diff
@@ -0,0 +1,280 @@
...
+  const chartSource: ChartSourceRef | undefined =
+    chartRef?.kind === 'OCIRepository'
+      ? { name: chartRef.name, namespace: chartRef.namespace }
+      : undefined;
+
+  // Everything else in the source's namespace that could be holding on to it.
+  const { resources: namespaceReleases, clustersData } = useResources(
+    cluster,
+    HelmRelease,
+    { [cluster]: { namespace: chartSource?.namespace ?? '' } },
```
marians: **This sibling list shares a react-query cache key with every other namespace's list, so the "provably unused" check can be answered from the wrong namespace — and then delete a shared `OCIRepository` that other agents still reference.**

`useListResources` builds its key as `['cluster', cluster, 'list', gvk.group, gvk.apiVersion, gvk.plural]` — `options` (the namespace) is not part of it. So this read for `team-a` and the same read for `team-b` on the same installation are *one* query. With the plugin's `staleTime: 60_000` (and a localStorage-persisted cache that restores `dataUpdatedAt`), the second one issues no request at all.

I verified this with a throwaway test: two `useResources('gazelle', HelmRelease, { gazelle: { namespace: 'team-a' | 'team-b' } }, { enableDiscovery: false })` renders produced exactly **one** HTTP request (`/apis/helm.toolkit.fluxcd.io/v2/namespaces/team-a/helmreleases/`), and the `team-b` hook received team-a's items.

Concrete failure: open agent `a1` in `team-a`, go back to the list, open agent `b1` in `team-b` (which still has two agents) within the minute. `namespaceReleases` is then team-a's list; none of those releases' `chartRef` resolves to `{ name: 'agent', namespace: 'team-b' }`, so `isChartSourceShared` is `false`, while `hasReadNamespaceReleases` is `true` because the cached query is `isSuccess`. `willRemoveChartSource` therefore becomes `true`, and deleting `b1` also deletes `OCIRepository/agent` in `team-b` — breaking every remaining agent's next reconciliation there. That is precisely the outcome this hook is built to prevent, and the "every uncertainty resolves to keeping it" invariant does not hold because the read *looks* certain.

The same mechanism makes a milder stale case possible even within one namespace: a sibling agent created 30 s ago in another tab is invisible to a cache entry that `staleTime` still considers fresh.

Worth fixing at the source — adding the namespace/label-selector to `useListResources`' `queryKey` would also remove the same latent hazard from `useNodePoolsForAWSCluster` and `SecretStoreSelector` — or, if that is too broad for this PR, read the sibling list here through a query whose key includes the namespace, and force it fresh before the destructive step rather than trusting a <=60 s cache.

plugins/agent-platform/src/hooks/useDeleteAgent.ts
```diff
@@ -0,0 +1,280 @@
...
+  // Memoized as a whole because the page passes this straight into the element it
+  // registers as the header's actions: a fresh object every render would
+  // re-register (and re-render) that slot on every poll.
+  return useMemo(
+    () => ({
+      /**
+       * Whether to offer the deletion: an owner we can act on, not declaratively
+       * owned by Flux, and the cluster says this user may delete it.
+       */
+      isDeletable: hasOwner && !isGitOpsOwned && isAllowed,
```
marians: `isDeletable` never requires that the owning `HelmRelease` was actually **read** — `hasOwner` is only "the agent carries the `helm.toolkit.fluxcd.io/name` label". The mutation, on the other hand, requires `helmRelease` (line 185).

So when that `get` *errors* rather than being in flight, the two disagree. `isLoadingHelmRelease` is `false`, `helmRelease` is `undefined`, and `isCheckingDeletable` is therefore `false`, so nothing withholds the item; a cached `allowed: true` SSAR makes `isDeletable` `true`. Reachable in practice:

- a transient proxy 5xx on that read — `ServiceUnavailableError` is not retried by the plugin's `QueryClient`, and this query has no `refetchInterval`, so it stays failed until the page remounts;
- RBAC that grants `delete` without `get` on `helmreleases`.

The user is then offered **Delete agent…**, and confirming fails with *"This agent has no HelmRelease to delete. It was applied directly, so it has to be removed the same way."* — which is false, and points them at the wrong remedy.

The same window silently switches off the Kustomization guard: `isGitOpsOwned` (line 84) defaults to `false` when the release could not be read, so a GitOps-owned agent is offered for deletion as well.

Suggest deciding on the object rather than the label, and treating an unreadable release as "cannot decide" instead of "no owner":

```suggestion
      isDeletable: Boolean(helmRelease) && !isGitOpsOwned && isAllowed,
```

plugins/agent-platform/src/hooks/useDeleteAgent.ts
```diff
@@ -0,0 +1,280 @@
...
+      if (!agent || !helmRelease || !helmReleaseName) {
+        throw new Error(
+          'This agent has no HelmRelease to delete. It was applied directly, so it has to be removed the same way.',
+        );
+      }
+
+      const helmReleaseGVK = helmRelease.getResolvedGVK();
+
+      try {
+        await deleteResource({
```
marians, line 213: A **suspended** `HelmRelease` turns this into a no-op that is reported as progress. As `docs/agent-platform.md` now acknowledges, Flux drops the finalizer on a suspended release without running the uninstall — so the `HelmRelease` goes away while the `Agent` CR and everything else the chart rendered stay behind.

The caller can't tell: the mutation resolves, so `AgentActionsMenu` posts `Deleting agent "…"` with "Flux is uninstalling its Helm release, so it may take a moment to disappear from the list" and navigates to the list — where the agent stays indefinitely, now with **no owning release at all**, i.e. no longer deletable through this path either (`hasOwner` still true from the label, `helmRelease` gone → the "applied directly" error from the check above).

The information is already in hand: `helmRelease.isSuspended()` (`FluxObject`). The test helper in `useDeleteAgent.test.tsx` even takes a `suspend` option that no test passes, which suggests a guard was considered and dropped. Cheapest correct behaviour is to refuse (or say so in the dialog) rather than claim an uninstall that will not happen.

plugins/agent-platform/src/hooks/useDeleteAgent.ts
marians: Confirmed and fixed at the source, in 42761038.

`useListResources` now includes the scope in its key. I verified the collision the way you describe, and also that `useGetResource` has always keyed on its namespace and name — so this was an inconsistency rather than a deliberate trade-off, which settles whether fixing it is in scope.

There is a regression test (`useListResources.test.tsx`) that renders two lists differing only by namespace against a shared client with a 60s `staleTime` and asserts two requests, each hook holding its own namespace's items. I checked it fails without the fix by reverting the key change: 1 failed, 1 passed. A second test pins the caching this exists for — same namespace is still one query.

The scope segment is appended **last**, so the existing 6-segment prefix still matches. Both invalidation call sites (`useFluxResourceActions` and this hook) use it as a prefix, so they are unaffected.

On your milder stale case: I took that as the more important half, because it survives the key fix. The sibling check no longer reads through the cache at all — new `fetchResourceList` in `kubernetes-react` lists the namespace *at mutation time*, and the check runs inside the mutation rather than during render. It could afford to be declarative before only because the modal displayed the outcome; after the modal was simplified, nothing needed it at render time.

This also means the "every uncertainty resolves to keeping it" invariant is now actually load-bearing: `fetchResourceList` throws on a failed read, and the caller treats that as "keep the source" rather than "nothing found".

plugins/agent-platform/src/hooks/useDeleteAgent.ts
marians: Right on all three counts — applied your suggestion in 42761038.

`isDeletable` is now `Boolean(helmRelease) && !isGitOpsOwned && isAllowed`. `hasOwner` survives only where it belongs: gating `enabled` on the release read and the access review, where the label genuinely is the right signal for "is there anything to look up".

The knock-on you spotted — the Kustomization guard silently switching off — was the part I had not seen, and it is the worse of the two, since it turns a deliberate safety check into a no-op precisely when we know least about the object.

I also fixed the error message the old path produced. `"This agent has no HelmRelease to delete. It was applied directly…"` was false in exactly this case: the agent carries the provenance label, so the release exists and we simply could not read it. It now says the release could not be read and suggests reloading, so nobody goes off deleting things by hand on its advice.

Test: `refuses when the owning release could not be read`. Writing it turned up a related trap worth mentioning — my `setup({ helmRelease: undefined })` re-triggered the default parameter and quietly gave the test a release, so the case now uses `null` with a comment on the helper explaining why.

plugins/agent-platform/src/hooks/useDeleteAgent.ts
marians: Correct, and your read of the history is right: the `suspend` option in the test helper is left over from a guard I considered and dropped. I had flagged the case to Marian as a possible conditional line in the dialog, he decided the modal should carry only the sessions warning, and I then let the caveat fall through to documentation instead of code — which left the docs describing a hazard the code walked straight into.

Refusing, in 42761038. The mutation now throws before deleting anything:

> HelmRelease `<ns>/<name>` is suspended. Flux would remove it without uninstalling the agent, leaving its resources behind. Resume the release first, then delete the agent.

Refusing in the mutation rather than hiding the menu item, for two reasons. Hiding it would leave someone who suspended the release themselves with no affordance and no explanation, and the message can name the remedy — resume, then delete. It also keeps the dialog body to the single sentence Marian asked for, since the text only appears on the failure path.

Your point about it being unrecoverable afterwards is what settled it. Deleting the release removes the only handle this path has: the `Agent` stays, `getHelmReleaseName` still returns a name, and the release it names is gone — so a second attempt could only produce the "could not be read" error from the thread above. Reported, meanwhile, as an uninstall in progress. That is the worst combination available, so refusing up front is the only defensible behaviour.

Test: `refuses to delete a suspended release rather than claim an uninstall`, and the `suspend` helper option now has a caller.

PR #2031, opened 2026-08-06
feat(agent-platform): delete a session from the details page

marians:
### What does this PR do?

Adds **Delete session…** to the session detail page's kebab menu, mirroring the agent deletion added in #2024. It calls kagent's `DELETE /api/sessions/:id` through the agent-platform proxy — the first write on the kagent REST side, so `agent-platform-backend` gains a `DELETE /kagent/sessions/:sessionId` route and its client a method parameter. Everything else about the transport is the reads' machinery reused: same installation resolution, same forwarded per-installation Dex token, same status mapping.

A few decisions worth reviewing:

**There is no permission gate, because there is nothing to ask.** Unlike the agent delete's `SelfSubjectAccessReview`, a session is not a Kubernetes object — kagent derives the acting user from the forwarded token alone (`GetUserID` → `GetPrincipal`), so the item is simply offered on any session that loaded. The route makes that token **required** for the same reason: without one, a controller running in `unsecure` mode would delete the shared default user's session on behalf of nobody in particular.

**A delete that matched nothing still answers 200.** kagent's statement is an `:exec`, so zero affected rows is not an error — a session that never existed, was already deleted, or belongs to another user all succeed silently. Nothing tries to detect that: a resolved promise means "kagent accepted this", and the refreshed list shows the truth a moment later. There is consequently no 404 path and no "already gone" case on the write side.

**One non-obvious piece of cache handling.** The sessions list key is invalidated normally — it is shared with `useAgentSessions`, so the fleet list and the agent page's recent-sessions card both correct themselves. This session's own two reads are invalidated with `refetchType: 'none'`: the detail page is still mounted at that moment, so refetching would race the navigation with a request that now 404s and flash "Session not found" at someone who just deleted the session deliberately.

**The menu is given an explicit width, and that line is load-bearing.** bui leaves `.bui-MenuContent`'s width to its content above a `min-width: 150px` (its own `width` fallback is the string `"undefined"`, which the browser discards), and `MenuItem` puts a 24px `gap` between label and trailing slot. "Delete session…" with its icon wants ~155px — just over — so the popover rendered at the natural width and then settled back to 150px, and that second layout pass made react-aria's popover resize observer trip the browser's `ResizeObserver loop completed with undelivered notifications`, twice, on every open. Sentry filters that message by default and production has no error overlay, but the dev-server overlay covers the page with it. The agent kebab escapes this only because its items happen to measure just under 150px, so it may be worth an upstream bui issue.

A session can be deleted from its detail page. The confirmation dialog says:

> The session and its conversation stop being listed and cannot be opened again. kagent keeps the record on its own server, but there is no way to restore it from here.

and, only when the `/me` probe reports the deployment is not user-scoped, adds:

> This kagent deployment does not scope sessions to individual users, so this session may have been started by somebody else.

On success the user lands back on the sessions list with a `Session "…" deleted` toast — past tense, unlike the agent's "Deleting", because kagent's delete is synchronous. On failure the dialog stays open and shows the message, with no toast.

Deleting from a list row is deliberately not offered: a destructive action on a row someone is scanning past is easy to hit by accident.

### How does it look like?

The kebab sits in the shared plugin header next to the section tabs, with a single red `Delete session…` item.

Verified end to end against `gazelle`, deleting a real session:

```
"DELETE /api/agent-platform/kagent/sessions/2f6047bb…?installation=gazelle HTTP/1.1" 200 79
```

That closes the one thing the repo could not answer — whether the installation's ingress passes `DELETE` through (no 405 from nginx or agentgateway). The 79-byte body is exactly kagent's `{"error":false,"data":{},"message":"Session deleted successfully"}` envelope. Afterwards the deep link lands on the existing "Session not found" empty state rather than an error alert, and the session is gone from the list. Nothing was logged at `warn`/`error`, so nothing reached Sentry.

Follows #2024, which established the write path for this plugin (kebab → `ConfirmDialog` → toast → navigate back) and left `ui-react`'s `ConfirmDialog` and `useProvidePageHeaderActions` behind to reuse — neither needed a change here.

Behaviour was verified against the kagent source at both `v0.9.9` (what the fleet runs) and `v0.10.0-rc1`: the handler, the SQL (`UPDATE session SET deleted_at = NOW() WHERE id = $1 AND user_id = $2 AND deleted_at IS NULL`, with every read filtering `deleted_at IS NULL`) and the 200-with-envelope response are identical in both.

Both react-query hooks are called by the page rather than the menu, because the menu renders in the shared plugin header — outside the plugin's `QueryClientProvider`, where a mutation or query would throw "No QueryClient set". Same constraint `AgentActionsMenu` documents.

Yes, done in this PR: `docs/agent-platform.md` gains a "Deleting a session" section, and the three stale "read-only" claims plus the "Session management + detail" open-TODO entry are corrected.

### DIFF
--- a/plugins/agent-platform/src/apis/KagentApiClient.ts
+++ b/plugins/agent-platform/src/apis/KagentApiClient.ts
@@ -229,8 +229,13 @@ export class KagentApiClient implements KagentApi {
     // `{error, data, message}` envelope, and the envelope carries nothing worth
     // returning — but a `200 error: true` would be a failure reported in-band, the
     // same shape the session and list readers already refuse to treat as success.
-    // A version that answered 204, or an empty body, still counts as a success:
-    // requiring JSON here would fail a delete that actually happened.
+    //
+    // An empty body still counts as a success: requiring JSON here would fail a
+    // delete that actually happened. That tolerance only reaches all the way
+    // through because `KagentClient.request` in agent-platform-backend has the
+    // matching half — it returns an upstream 204 as an empty success instead of
+    // reporting a missing content-type as a sign-in page, and the route passes it
+    // on as a 204. Both halves are needed; neither is any use alone.
     const body = await response.json().catch(() => undefined);
     if (isErrorEnvelope(body)) {
       throw upstreamError(
--- a/plugins/agent-platform/src/apis/KagentApiClient.ts
+++ b/plugins/agent-platform/src/apis/KagentApiClient.ts
@@ -55,6 +55,21 @@ function upstreamError(message: string): Error {
   return error;
 }
 
+/**
+ * A kagent envelope reporting a failure in-band on an otherwise successful
+ * response.
+ *
+ * Deliberately narrow — `error === true` and nothing else. The reads go through
+ * the zod layer for this; the write path has no payload to parse, so it checks the
+ * one field that can turn a 200 into a failure.
+ */
+function isErrorEnvelope(body: unknown): body is Record<string, unknown> {
+  if (typeof body !== 'object' || body === null) {
+    return false;
+  }
+  return (body as { error?: unknown }).error === true;
+}
+
 /** Which read produced the drift — part of the dedupe key and of the message. */
 type DriftSource = 'sessions' | 'session' | 'session tasks';
 
@@ -198,6 +213,34 @@ export class KagentApiClient implements KagentApi {
     return tasks;
   }
 
+  async deleteSession(installation: string, sessionId: string): Promise<void> {
+    const { url, headers } = await this.prepare(
+      `/kagent/sessions/${encodeURIComponent(sessionId)}`,
+      installation,
+    );
+    const response = await this.fetchApi.fetch(url, {
+      method: 'DELETE',
+      headers,
+    });
+
+    await this.throwIfNotOk(response);
+
+    // The body is read but not required. kagent answers 200 with its usual
+    // `{error, data, message}` envelope, and the envelope carries nothing worth
+    // returning — but a `200 error: true` would be a failure reported in-band, the
+    // same shape the session and list readers already refuse to treat as success.
+    // A version that answered 204, or an empty body, still counts as a success:
+    // requiring JSON here would fail a delete that actually happened.
+    const body = await response.json().catch(() => undefined);
+    if (isErrorEnvelope(body)) {
+      throw upstreamError(
+        typeof body.message === 'string' && body.message
+          ? body.message
+          : 'kagent reported an error while deleting the session, without saying what.',
+      );
+    }
+  }
+
   async getIdentity(installation: string): Promise<KagentIdentity> {
     // Best-effort token: the backend reads it as optional here, so a broker or
     // Dex-session failure must not stop the probe. This is the one installation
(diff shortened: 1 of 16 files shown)


plugins/agent-platform/src/hooks/useDeleteSession.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const mutation = useMutation({
+    mutationFn: async () => {
+      await kagentApi.deleteSession(installation, sessionId);
+
+      // The one key that matters, and it is shared: `SessionsDataProvider` and
+      // `useAgentSessions` read the fleet list under it, so this corrects both the
+      // sessions list the user is about to land on and the agent page's recent
+      // sessions card.
+      await queryClient.invalidateQueries({
```
marians, line 37: The `useAgentSessions` half of this claim cannot hold: the Sessions tab and the Agents tab each mount their **own** `QueryClientProvider` (`components/QueryClientProvider.tsx` builds the client with `useMemo(() => new QueryClient(config), [])`, and `SessionsRouter` and `AgentsRouter` each render one), so this `queryClient` is not the client `useAgentSessions` reads from on the agent detail page. Concretely: delete a session here, switch to Agents, open an agent — its recent-sessions card is served by a different cache that this invalidation never touched.

It happens to look right anyway, because `AgentsRouter` mounts a fresh client and the `sessions` keys are excluded from persistence, so the card refetches from scratch — but that is not the reason given here. The same claim is repeated in `docs/agent-platform.md` and the changeset; worth trimming all three to the sessions list, so nobody later relies on cross-tab invalidation that isn't there.

plugins/agent-platform/src/apis/KagentApiClient.ts
```diff
@@ -198,6 +213,34 @@ export class KagentApiClient implements KagentApi {
...
+
+    await this.throwIfNotOk(response);
+
+    // The body is read but not required. kagent answers 200 with its usual
+    // `{error, data, message}` envelope, and the envelope carries nothing worth
+    // returning — but a `200 error: true` would be a failure reported in-band, the
+    // same shape the session and list readers already refuse to treat as success.
+    // A version that answered 204, or an empty body, still counts as a success:
+    // requiring JSON here would fail a delete that actually happened.
+    const body = await response.json().catch(() => undefined);
```
marians, line 239: The 204 / empty-body tolerance described here is unreachable through our own proxy, so the claim in the comment (and in the changeset) only holds for half the path.

`KagentClient.request` in `agent-platform-backend` rejects **any** 2xx whose content-type isn't `application/json` with an `AuthenticationError` ("returned a non-JSON response … which usually means a sign-in page was served instead") → 401, and otherwise does `await response.json()`, which throws on an empty body and becomes an `UpstreamError` → 500.

So the scenario the comment is written for — a future kagent answering `204 No Content` — plays out as: the session is deleted upstream, the frontend never reaches this line, the dialog stays open showing a sign-in-page error, the sessions list is not invalidated and no navigation happens. If it is worth supporting, the backend client needs the matching half (e.g. treat 204 / `content-length: 0` as an empty success for the DELETE); otherwise the comment here overstates what the stack tolerates.

plugins/agent-platform/src/hooks/useDeleteSession.ts
marians: Confirmed and fixed in 09d8a3aa. `AgentsRouter` and `SessionsRouter` each render their own `<QueryClientProvider>`, and the client is `useMemo(() => new QueryClient(config), [])` — per mount — so `useAgentSessions` on the agent detail page provably reads a different cache from the one this invalidation touches. The comment was describing a shared *key* as if it were a shared *cache*.

Trimmed in all three places (hook, `docs/agent-platform.md`, changeset) to claim only the Sessions tab, plus an explicit note that the card needs no help — fresh client starts empty, and these keys are excluded from persistence — so nobody later builds on cross-tab invalidation that isn't there.

plugins/agent-platform/src/apis/KagentApiClient.ts
marians: Right on both counts, and I went with implementing the missing half rather than trimming the claim — 09d8a3aa.

The failure mode you describe is the bad kind: the deletion has *happened* upstream, and the user is told a sign-in page was served while the dialog sits there on an error. So `KagentClient.request` now returns a `204` as an empty success, ahead of both guards (it carries no content-type, so the sign-in-page check would misread it, and `response.json()` would throw on the empty body), and the delete route passes that on as a `204` rather than `res.json(undefined)`'s empty-body-under-a-JSON-content-type. Two tests: the client resolves undefined on a 204, and the route answers 204 with an empty body.

Nothing kagent serves today hits this — both v0.9.9 and v0.10 answer 200 with the envelope — but now the tolerance the frontend comment describes exists end to end, and that comment says explicitly that both halves are needed and neither is any use alone.

PR #2049, opened 2026-08-06
fix(agent-platform): render the question a session is waiting on

marians:
### What does this PR do?

A session whose last turn ends by asking the user something showed **nothing at all** for that question. The timeline stopped at the agent's previous message, so it read as if the agent had trailed off mid conversation — only the "Waiting for input" badge in the header hinted otherwise.

The question fell between two paths that each assumed the other had it:

- The raw `ask_user` call **is** in `task.history`, but it is deliberately skipped as ADK plumbing (`INTERNAL_TOOL_NAMES` in `kagentParts.ts`) on the grounds that *"`ask_user` is rendered by the approval path rather than as a tool call."*
- The approval path renders `adk_request_confirmation` — but it only ever read `history`, and for an **unanswered** question there is nothing there to read. kagent puts the pending confirmation on `task.status.message` and nowhere else.

`buildTimeline` now appends `status.message` as a final history entry while the task is waiting on it, so the existing approval handling applies unchanged — dedupe, part walking, the `ask_user` → `asks: 'input'` discrimination, and the cross-task verdict tracking — with no second code path. kagent's own UI solves this with two separate passes that are concatenated; keeping one list means the question stays in its chronological place rather than being appended to the end.

It is gated on the state (`input-required` / `auth-required`) rather than merely on the message being present, which is what makes the card self-clearing: once the user answers elsewhere, the task reaches a terminal state, the prompt stops being emitted from `status`, and the now-answered confirmation renders from history with its verdict. A question therefore cannot linger on screen after it has been answered.

Second part: **the question now renders as prose.** Making the row appear was not quite enough — `ask_user` arguments were only reachable by expanding the row, where they showed as JSON in a monospace, single-line-ellipsised slot built for tool payloads. A question is the last thing the agent *said*, so the approval item now carries the extracted `questions` and the entry renders them as markdown (numbered when one `ask_user` asks several). With the questions on the row there is nothing left behind the expander, so it renders as a plain row instead of offering a click that reveals a worse copy of what is already on screen. This applies to already-answered questions too, which had the same problem less visibly.

Extraction is deliberately narrow — `questions[].question` strings and nothing else. A question rendered from a guessed field would put words in the agent's mouth, so an unrecognised payload falls back to showing the raw arguments as before.

A session that ends with the agent asking a question now shows that question in the conversation, marked "User input requested / Awaiting a reply", instead of appearing to stop mid conversation. Questions that were already answered are now readable as text rather than as JSON behind an expander.

### How does it look like?

Verified in the browser against a live gazelle session (`019fd5d5-…`), whose last task is `input-required`. The end of the timeline now reads:

```
6 Aug 2026, 06:49 UTC

You
Please ask me a question using the appropriate tool. Just any question will do.

User input requested   [Awaiting a reply]
What would you like to investigate next on the gazelle MC? For example: dig into
the muster auth config, check the agentic-platform-connectivity Helm release, or
look at something else entirely?
```

Before this change, everything from "User input requested" down was absent.

The shape was confirmed against the live payload rather than inferred. On that session's final task, `status.message` carries `messageId` `019fd5d5-8f0b-…`, which appears in **none** of the task's four history entries, and wraps the question in the same `adk_request_confirmation` envelope an answered one has:

```json
{ "args": { "originalFunctionCall": { "name": "ask_user",
            "args": { "questions": [ { "question": "…" } ] } },
            "toolConfirmation": { "confirmed": false, "hint": "…" } },
  "name": "adk_request_confirmation" }
```

It carries no message-level metadata, so it contributes no token usage — covered by a test, since a prompt that inflated the session's token total would be a quiet regression.

The new fixture (`tasks.ask-user-pending.json`) is shaped from that payload with the wording and ids replaced; real `/tasks` payloads are not committed, as they contain full conversations including tool arguments and results.

One existing test changed contract: it clicked the expander to assert a "Questions" payload, and now asserts the question is visible as prose with no expander.

No.

### DIFF
--- a/plugins/agent-platform/src/lib/kagentTimeline.ts
+++ b/plugins/agent-platform/src/lib/kagentTimeline.ts
@@ -78,6 +78,16 @@ export type TimelineItem =
       asks: 'approval' | 'input';
       /** The tool the agent proposed to run, when the payload named one. */
       toolName?: string;
+      /**
+       * The questions themselves, when this is an `ask_user`.
+       *
+       * Extracted here rather than left for the UI to dig out of {@link args},
+       * because a question is the last thing the agent *said* — it belongs in the
+       * conversation as prose, not behind an expander as JSON. Empty when the
+       * payload used a shape we don't recognise, in which case the raw args remain
+       * the only record and the UI falls back to showing them.
+       */
+      questions?: string[];
       args?: unknown;
       /** Undefined while the request is still unanswered. */
       verdict?: 'approved' | 'rejected';
@@ -102,6 +112,54 @@ export type SessionTimeline = {
 type OpenCall = { callId: string; itemIndex: number };
 
 const EMPTY_USAGE: TokenUsage = { total: 0, prompt: 0, completion: 0 };
+
+/**
+ * The A2A states in which `status.message` is a prompt the task is *waiting on*,
+ * rather than incidental status text.
+ *
+ * The legacy (v0) spellings, which is what this client reads: `listSessionTasks`
+ * deliberately sends no `A2A-Version` header, and kagent treats a missing header
+ * as the legacy wire.
+ */
+const AWAITING_INPUT_STATES = new Set(['input-required', 'auth-required']);
+
+/**
+ * A task's history, plus the question it is currently waiting on.
+ *
+ * kagent puts an *unanswered* confirmation on `task.status.message` and **not** in
+ * `history`, so a session that ends by asking the user something rendered as if
+ * the agent had simply stopped talking. The raw `ask_user` call does appear in
+ * history, but it is deliberately skipped as ADK plumbing (`INTERNAL_TOOL_NAMES`)
+ * because the approval path is supposed to render it — and that path only ever
+ * looked at history. The question fell between the two.
+ *
+ * Verified against a live gazelle session: the pending `status.message` carries a
+ * distinct `messageId` that appears nowhere in `history`, wrapping the question in
+ * the same `adk_request_confirmation` shape an answered one has. So appending it
+ * as a final entry gets the existing approval handling — including the
+ * `ask_user` -> `asks: 'input'` discrimination — for free.
+ *
+ * Gated on the state rather than merely on the message being present. Two reasons:
+ * it is the documented contract (`status.message` "carries the pending prompt
+ * while a task waits for input"), and it makes the item self-clearing — once the
+ * user answers elsewhere and the task reaches a terminal state, the prompt stops
+ * being emitted here and the answered confirmation renders from history instead,
+ * so the card cannot linger as a question that has already been answered.
+ *
+ * kagent's own UI splits the same problem across two passes
+ * (`extractMessagesFromTasks` skips unresolved confirmations in history;
+ * `extractApprovalMessagesFromTasks` reads `status.message`). One list keeps the
+ * question in its chronological place instead of appending it to the end.
+ */
+function historyWithPendingPrompt(task: A2aTaskWire): unknown[] {
+  const history = Array.isArray(task.history) ? task.history : [];
+  const state = task.status?.state?.toLowerCase();
+  const pending = task.status?.message;
+  if (!pending || !state || !AWAITING_INPUT_STATES.has(state)) {
+    return history;
+  }
+  return [...history, pending];
+}
 
 /**
  * Turn kagent's A2A tasks into a flat, renderable timeline.
@@ -166,14 +224,14 @@ export function buildTimeline(tasks: A2aTaskWire[]): SessionTimeline {
     // because this is now the *only* timestamp source — it becomes `at` for every
     // item in the task.
     const taskTimestamp = normalizeTimestamp(task.status?.timestamp);
-    const history = Array.isArray(task.history) ? task.history : [];
+    const entries = historyWithPendingPrompt(task);
 
     // Open calls are per task: a response never answers a call from another turn,
     // and letting them match across tasks would attach a result to the wrong call
     // when a tool is used repeatedly.
     const openCalls: OpenCall[] = [];
 
-    history.forEach((entry, entryIndex) => {
+    entries.forEach((entry, entryIndex) => {
       const parsed = parseHistoryEntry(entry);
       if (parsed.kind === 'unparseable') {
         skippedMessages += 1;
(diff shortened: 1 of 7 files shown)


plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -498,6 +560,39 @@ function makeCallItem(input: {
  * kagent wraps it as `args.originalFunctionCall` on the `adk_request_confirmation`
  * call (`buildApprovalMessage` in kagent's UI reads the same field).
  */
+/**
```
marians, line 593: The new docblock is inserted *between* `readProposedCall`'s docblock ("The tool an approval request is asking about… `args.originalFunctionCall`") and the function it documents, so two comment blocks now stack here and the first one describes `readAskUserQuestions`, which is not what it is about. `readProposedCall` is left undocumented and the first block reads as a wrong description of the new helper (in IDE hover, only the *last* block attaches — so the `originalFunctionCall` doc is now effectively lost). Moving `readAskUserQuestions` below `readProposedCall` (or moving the orphaned block down onto it) fixes it.

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -103,6 +113,54 @@ type OpenCall = { callId: string; itemIndex: number };
...
+ * question in its chronological place instead of appending it to the end.
+ */
+function historyWithPendingPrompt(task: A2aTaskWire): unknown[] {
+  const history = Array.isArray(task.history) ? task.history : [];
+  const state = task.status?.state?.toLowerCase();
+  const pending = task.status?.message;
+  if (!pending || !state || !AWAITING_INPUT_STATES.has(state)) {
+    return history;
+  }
+  return [...history, pending];
```
marians, line 173: The self-clearing argument assumes the answered task leaves `input-required`. But this file's own test `attaches a verdict recorded in a later task` documents that "kagent can record the decision in a new task" — in that shape the *asking* task is not the one that gets resumed, so its `status.state`/`status.message` may well stay at `input-required` with the prompt still attached. Then, on the next poll, the answered `adk_request_confirmation` renders from the later task's history **and** this appends the still-present pending prompt: the same question shows twice, once as "Awaiting a reply" forever. The `seenMessageIds` dedupe only saves us if the history copy reuses the pending message's `messageId`, which the PR's live evidence (a pre-answer snapshot) cannot confirm. Worth checking against a session that has actually been answered; if the ids differ, a guard like "skip when a confirmation with this proposed-call id already rendered" would be needed on top of the state gate.

plugins/agent-platform/src/lib/kagentTimeline.ts
```diff
@@ -103,6 +113,54 @@ type OpenCall = { callId: string; itemIndex: number };
...
+ *
+ * kagent's own UI splits the same problem across two passes
+ * (`extractMessagesFromTasks` skips unresolved confirmations in history;
+ * `extractApprovalMessagesFromTasks` reads `status.message`). One list keeps the
+ * question in its chronological place instead of appending it to the end.
+ */
+function historyWithPendingPrompt(task: A2aTaskWire): unknown[] {
+  const history = Array.isArray(task.history) ? task.history : [];
+  const state = task.status?.state?.toLowerCase();
+  const pending = task.status?.message;
```
marians, line 164: `status.message` is typed `z.unknown()` in `a2aTaskStatusWireSchema`, and it is now fed straight into `parseHistoryEntry`. Anything that is not object-shaped (a kagent version that puts a plain string there, e.g. an `auth-required` hint) fails `a2aMessageWireSchema.safeParse` and lands in the `unparseable` branch, incrementing `skippedMessages` — which surfaces to the user as the "1 message could not be read" alert in `SessionTimeline.tsx` on an otherwise healthy waiting session. Cheap fix: only append when it is a non-null object (`typeof pending === 'object'`), so a shape we can't read stays invisible rather than being reported as data loss.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Correct, and the IDE-hover consequence is the part that makes it more than cosmetic. Fixed in 170f9787 by moving `readAskUserQuestions` below `readProposedCall`, so each docblock sits on the function it describes and the `originalFunctionCall` doc is attached again.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Confirmed and fixed in 170f9787. `historyWithPendingPrompt` now also requires `typeof pending === 'object'`, so a shape we cannot render stays invisible instead of being reported as data loss.

Added a test for it (`does not report an unreadable status message as data loss`) asserting `skippedMessages` stays 0 and no approval renders when `status.message` is a bare string.

plugins/agent-platform/src/lib/kagentTimeline.ts
marians: Good catch to push on — you're right that the live evidence was a pre-answer snapshot and couldn't settle it. I went to kagent's source rather than reasoning from our fixtures, and the resume path rules the scenario out.

`go/adk/pkg/a2a/executor.go:225-238`:

```go
if reqCtx.StoredTask == nil {
    // New task — emit submitted with the user's message
} else if ExtractDecisionFromMessage(reqCtx.Message) != "" {
    dropPreAppendedDecisionFromHistory(reqCtx.StoredTask, reqCtx.Message)
    decision := a2atype.NewStatusUpdateEvent(reqCtx, a2atype.TaskStateWorking, reqCtx.Message)
```

A HITL decision resumes the **stored** task: it emits `working` on that same task and appends the decision to *its* history — it does not open a new one. `BuildResumeHITLMessage` (`hitl.go:375`) reinforces it from the other side, refusing to build a resume at all unless `storedTask.Status.State == TaskStateInputRequired` and `Status.Message != nil`. So the asking task cannot both be answered and stay `input-required`; the state gate clears the prompt on the same poll that reveals the answered copy.

Corroborated by `tasks.ask-user.json`, whose shapes came from a real *answered* `ask_user` session: state `completed`, no `status.message`, confirmation in history with the decision after it. That is exactly the post-answer shape the source predicts.

There is a related case the gate handles correctly rather than by luck: if the user starts a fresh turn *without* a decision payload, `BuildResumeHITLMessage` returns nil and a new task is created while the old one stays `input-required` with its prompt. Rendering it as still awaiting a reply is right there — ADK genuinely still has that confirmation pending — and no duplicate arises, because the confirmation is never written to history with a verdict.

Rather than leave this as an argument in a PR thread, 170f9787 pins it: `renders one approval, not two, once the question is answered` builds the post-answer shape (task completed, confirmation + decision appended to history) and asserts exactly one approval item, carrying the verdict. So if kagent ever does start leaving the asking task in `input-required`, that test fails rather than the UI quietly showing the question twice.

I did not take the extra guard (skip when a confirmation with this proposed-call id already rendered) — with the overlap ruled out at the source it would be dead code, and the test is what protects us if that changes. Happy to add it if you'd rather be belt-and-braces.

PR #2063, opened 2026-08-07
feat(agent-platform): rename a session from its detail page

marians:
### What does this PR do?

Lets a session be renamed from its detail page. Two ways in: a `Rename session…` item in the kebab, and the page title itself — a real `<button>` stripped of its chrome rather than a click handler on the heading, so it is keyboard-reachable and announced as operable. Both open the same dialog, which is why its open state sits on the page rather than inside the actions menu the way the delete's does.

**kagent's rename endpoint does not rename on the version the fleet runs**, and that shapes the whole implementation. `PUT /api/sessions/{session_id}` is registered on v0.9.x, but `HandleUpdateSession` there:

- requires **both** `name` and `agent_ref`, rejecting either omission with a 400;
- never reads the `{session_id}` path param — it looks the session up by `*sessionRequest.Name`, i.e. it treats the new name as the id;
- assigns only `session.AgentID`. `session.Name` is never written.

Fixed in `v0.10.0-rc1`. GS pins **0.9.9** (`agentic-platform`'s `values.yaml`), so on every installation we run today the correct endpoint is inert.

The write therefore falls back to `POST /api/sessions` with the existing `id`, whose `StoreSession` is an upsert on `(id, user_id)` that does write `name` — identical SQL in v0.9.9 and v0.10.0-rc1. Echoing the session's own `agent_id` back as `agent_ref` round-trips exactly, since kagent's `ConvertToPythonIdentifier` only rewrites `-` and `/`, neither of which survives in an already-encoded id.

**Only a 400 enters that fallback, and it means "this kagent predates the fix" — nothing more.** The PUT's status says nothing about whether the session still exists: v0.9.x rejects the missing `agent_ref` *before* it looks anything up, so a live session and a deleted one both answer 400, and the 404 that would separate them is reachable only on v0.10+. Everything else (401, 403, 404, 5xx) is surfaced as the failure it is.

**A read-back, not the status, is what enforces "never create".** Before writing, the fallback fetches the session (`GET /sessions/{id}?limit=1`) and gives up if it is gone — otherwise the upsert, which *inserts* when nothing conflicts, would resurrect a session someone had just deleted under its old id. That read also makes the echoed `agent_id`/`source` authoritative: both are overwritten by the write, so they come from kagent rather than from the browser, where a stale or unparsed value would blank a column nobody asked to touch.

Every remaining way the fallback can fail, it fails before writing, and each is a **4xx** rather than a server error: the session is gone (404), it has no agent (409), kagent cannot resolve that agent (409), or a sandbox-workload agent already holds a session (409). None is actionable, and on a fleet where every installation takes this branch a 5xx would mean a standing Sentry issue per case.

All of the fallback is marked `TODO(kagent-0.9)` and comes out when no installation runs kagent v0.9.x, leaving the plain PUT. Our own route takes nothing but the name, so it needs no change then.

A session can be given a meaningful name. Worth having because kagent derives titles from the first message and truncates them to 20 characters, so a session that mattered ends up filed under half a sentence.

Smaller decisions:

- The invalidations **refetch**, unlike the delete's `refetchType: 'none'` — nothing navigates away, so the page has to show the new name. Both are awaited inside the mutation, so the dialog closes onto data that has caught up.
- The name is trimmed, required, and capped at 255 characters. That bound is ours, not kagent's (`session.name` is unbounded `TEXT`), and the backend route enforces it as well as the dialog.
- Confirming does not close the dialog, nor can it be dismissed while the request is in flight — including via the close button bui's `DialogHeader` always renders, which `isDismissable` does not reach.
- The dialog seeds its field on the open transition only, never on a `title` change: the session read polls, so re-seeding would wipe an edit in progress when the session is renamed elsewhere.
- `updated_at` moves, so a renamed session rises to the top of the list — correct for an edit.

### How does it look like?

Verified live against gazelle (kagent 0.9.9), which exercises the **fallback** — the path that matters most. The PUT response came back:

```json
{
  "error": false,
  "data": {
    "id": "46d0ad4d…",
    "name": "Renovate PR reduction (grilling)",
    "created_at": "2026-07-30T13:05:03.529932Z",
    "updated_at": "2026-08-07T08:36:49.20525Z",
    "agent_id": "kagent__NS__grill_master"
  },
  "message": "Successfully created session"
}
```

`"Successfully created session"` is `HandleCreateSession`'s envelope, so kagent rejected the PUT and the upsert did the work. `created_at` is unchanged (the upsert conflicted and updated rather than inserting) and `agent_id` came back identical, proving the `agent_ref` round trip is lossless.

Re-verified after the review rework, which added the read-back: on 0.9.9 the title can only change via the upsert, and the upsert is now unreachable except through a successful read-back — so a working rename exercises all three legs. The session was renamed back afterwards both times.

`docs/agent-platform.md` previously recorded rename as "just wire up kagent's `PUT /api/sessions/:id`". That is wrong for the pinned version, and this PR rewrites that claim along with the "Renaming and continuing a session" gap entry.

The first revision of this PR claimed the 400-vs-404 split alone made creation impossible. It did not — see the review thread on `updateSessionName`: on v0.9.x the PUT 400s at validation before any lookup, so a deleted session never produces the 404 the guard depended on. The read-back replaces that reasoning.

**Not verified live:** the "session already deleted" race itself. It is covered by unit tests (the read-back's 404 and its no-session-in-body variant both assert that no POST is issued), but reproducing it end-to-end needs either raw API calls with a Dex token or deleting a real session.

Yes — done in this PR. `docs/agent-platform.md` gains a "Renaming a session" section beside "Deleting a session", recording the v0.9.9 handler bug, why the read-back rather than the status is the safety property, and the removal criteria for the fallback.


plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -322,6 +365,98 @@ export class KagentClient {
...
+      `kagent rejected the session rename on installation '${this.installation.name}'; retrying via the session upsert`,
+    );
+
+    if (!fallback.agentRef) {
+      throw upstreamError(
+        `The kagent API for installation '${this.installation.name}' is running a version that cannot rename a session, and the workaround needs the session's agent, which this session does not have.`,
+      );
+    }
+
+    return this.request(`${this.installation.apiBaseUrl}/sessions`, options, {
```
marians: **The 404 guard is inert on exactly the versions that reach this fallback.**

As this method's own doc states, v0.9.x `HandleUpdateSession` rejects the request at *validation* — it requires both `name` and `agent_ref`, and we deliberately send neither `agent_ref` nor anything else — which happens **before** it looks anything up. So on v0.9.x the PUT answers 400 identically whether or not the session exists; a missing session never produces the 404 that the design relies on to block the fallback.

Concrete scenario, on the version the whole fleet runs today: the session is deleted from another tab, from kagent's own UI, or from another device, and the detail page has not polled yet (baseline interval is 60 s). The user presses Save. The PUT 400s (missing `agent_ref`, as always). This `POST /sessions` then upserts on `(id, user_id)` with nothing to conflict with — inserting a brand-new empty session under the old id, or resurrecting the soft-deleted row. That is precisely the outcome the PR description, the docs table and the `does NOT fall back on a 404` test say cannot happen.

A cheap fix that also removes the need to trust the client's `agentRef`/`source`: on the fallback path, read the session back first (`GET /sessions/{id}`) and only upsert if it is still there, echoing its own `agent_id`/`source` from that read.

plugins/agent-platform/src/components/SessionDetailPage/SessionRenameDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+    event.preventDefault();
+    if (isValid && !isRenaming) {
+      onConfirm(trimmed);
+    }
+  };
+
+  return (
+    <Dialog
+      isOpen={isOpen}
+      onOpenChange={onOpenChange}
```
marians: **The dialog can still be dismissed while the rename is in flight — by the header's own close button.**

`isDismissable`/`isKeyboardDismissDisabled` only block the outside click and Escape. `DialogHeader` from `@backstage/ui` unconditionally renders its own close control (`<Button name="close" aria-label="Close" slot="close">`, see `node_modules/@backstage/ui/dist/components/Dialog/Dialog.esm.js`), and nothing here gates it on `isRenaming`.

Scenario: the user presses Save, the request is slow (the mutation also awaits an invalidation that refetches and retries up to twice with backoff), and they click the X. The dialog closes, the mutation keeps running, and when it fails the page's `onConfirm` catch swallows the rejection while the dialog — the only surface that renders `rename.error` — is gone. The user is left on the old title believing the rename succeeded. The `locks itself while the rename is in flight` test only asserts the *Cancel* button.

The close button routes through the dialog's own close, so guarding `onOpenChange` covers both it and any future affordance:

```suggestion
      onOpenChange={next => {
        // Also covers DialogHeader's own close button, which neither
        // `isDismissable` nor `isKeyboardDismissDisabled` reaches.
        if (!isRenaming) {
          onOpenChange(next);
        }
      }}
```

plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -322,6 +365,98 @@ export class KagentClient {
...
+      }
+    }
+
+    // TODO(kagent-0.9): remove from here to the end of the method.
+    this.logger.debug(
+      `kagent rejected the session rename on installation '${this.installation.name}'; retrying via the session upsert`,
+    );
+
+    if (!fallback.agentRef) {
+      throw upstreamError(
```
marians: **This turns an expected outcome into a 5xx, and therefore into recurring Sentry issues.**

`upstreamError` is documented as surfacing a 500 so it *reaches Sentry*, which is right for "kagent is deployed but unwell". A session with no agent is not that: `KagentSession.agentId` is optional (`plugins/agent-platform/src/lib/kagentSessions.ts:27`), kagent's `agent_id` is nullable, and today **every** installation runs v0.9.x — so every rename attempt on an agent-less session is an unavoidable, non-actionable 500 plus a Sentry event, for what is really "this session cannot be renamed on this kagent".

The same applies to the fallback's own documented failure modes below: the 409 from a sandbox-workload agent that already holds a session, and a 400 from an agent kagent cannot resolve, both land on the generic `!response.ok` branch and become 500s too.

CLAUDE.md is explicit ("Never return a 5xx for an expected outcome" / `MiddlewareFactory` logs >= 500 to Sentry regardless of your own log level). A 4xx from `@backstage/errors` — `ConflictError` or `NotAllowedError` — reports the same thing to the dialog without the Sentry fan-out.

plugins/agent-platform/src/apis/KagentApiClient.ts
```diff
@@ -246,6 +246,43 @@ export class KagentApiClient implements KagentApi {
...
+      body: JSON.stringify({
+        name,
+        // Read only by the backend's v0.9.x workaround; see
+        // `KagentApi.renameSession`.
+        ...(fallback.agentRef && { agentRef: fallback.agentRef }),
+        ...(fallback.source && { source: fallback.source }),
+      }),
+    });
+
+    await this.throwIfNotOk(response);
```
marians: **A name-validation 400 from the new route is reclassified as `NotFoundError`.**

`throwIfNotOk` (line 360) renames *any* 400 to `NotFoundError`, on the documented assumption that a 400 from this proxy only ever means "unknown installation", i.e. the silent "kagent isn't available here" path the reads take. The PUT route added in this PR breaks that assumption: it answers 400 for `name must be a string`, `name must not be empty` and `name must be at most N characters`.

Today the impact is contained — the dialog only renders `error.message`, which survives — but the error is now labelled as the one name the plugin treats as "kagent is absent", so any future handling that branches on the name (or a caller that reuses `renameSession`) will read a rejected name as a missing kagent. Worth restricting the 400 → `NotFoundError` mapping to the read path, or discriminating on the backend error's own name.

plugins/agent-platform/src/components/SessionDetailPage/SessionRenameDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+  isOpen,
+  onOpenChange,
+  isRenaming,
+  error,
+  onConfirm,
+  isUserScoped,
+}: SessionRenameDialogProps) {
+  const [value, setValue] = useState(title);
+
+  useEffect(() => {
```
marians, line 77: **This re-seeds on every `title` change, not only on the open transition, so it can clobber an edit in progress.**

`title` is live data: the session read polls on the baseline interval, so it changes under the dialog whenever the session is renamed elsewhere (kagent's own UI, another tab, another device). While the dialog is open the effect fires and replaces whatever the user has typed with the new upstream title — mid-sentence, with no indication anything happened.

The intent (documented above, and covered by the "forgets an abandoned edit when reopened" test) is only to reset on open. Seeding on the `false → true` transition preserves that without touching an active edit — e.g. keep the previous `isOpen` in a ref and re-seed only when it flips, and drop `title` from the dependency list.

plugins/agent-platform-backend/src/KagentClient.ts
```diff
@@ -322,6 +365,98 @@ export class KagentClient {
...
+    return this.request(`${this.installation.apiBaseUrl}/sessions`, options, {
+      method: 'POST',
+      // `source` is echoed back because the upsert overwrites it from what we
+      // send. v0.9.9 omits it from its reads, in which case it is already null
+      // and sending nothing changes nothing.
+      body: {
+        id: sessionId,
+        name,
+        agent_ref: fallback.agentRef,
+        ...(fallback.source && { source: fallback.source }),
```
marians: **"v0.9.9 omits `source` from its reads" does not imply the column is null — it is the case where we cannot know.**

The comment's inference runs the wrong way: if the read never returns `source`, then `fallback.source` is always `undefined` on v0.9.x and this spread always omits it, whatever is actually stored. And the doc states the upsert SQL is identical in v0.9.9 and v0.10.0-rc1, which means `StoreSession` on v0.9.9 *does* write the column — so if only the response type drops it, every rename on the fleet's current version overwrites a populated `source` with the zero value. That is a silent field wipe on an operation the user thinks only touches the name.

Worth confirming against the v0.9.9 session response struct. Reading the session back from kagent on the fallback path (see the other comment on the 404 guard) would settle it for good, since the echoed values would then come from kagent rather than from what the browser happened to be able to parse.

plugins/agent-platform-backend/src/KagentClient.ts
marians: You're right, and this was the load-bearing mistake in the whole design. I read the 400 as "old kagent" and *also* as "but a missing session would have 404'd" — the second half never holds on v0.9.x, because the `agent_ref` check runs before the lookup. So the guard was inert on precisely the versions that reach it, and the scenario you describe (delete elsewhere, 60 s poll window, press Save) would have inserted the session back under its old id.

Taken your fix in 4b07fec: the fallback now reads the session back (`GET /sessions/{id}?limit=1`, trimmed like `getSession`) and gives up if it is gone. The status is demoted to what it can actually tell us — "this kagent predates the fix" — and the read is what enforces "never create".

Two tests pin it: `never writes when the session is already gone` (404 on the read) and `never writes when the read comes back without a session` (readable 200, no session — the same condition, since kagent scopes by user id). Both assert no POST is issued, not just that it throws.

As you note, this also removes the need to trust the client: `agent_ref` and `source` now come from that read, and the route no longer accepts them at all.

plugins/agent-platform/src/components/SessionDetailPage/SessionRenameDialog.tsx
marians: Confirmed — the X is there in the live a11y tree alongside Cancel and Save, and I had only disabled Cancel. Applied your suggestion in 4b07fec, gating `onOpenChange` so it covers the header button and anything added later.

Two tests, since the guard has to fail in both directions: `cannot be closed by the header's own X while the rename is in flight`, and `can be closed by the header's X when nothing is in flight` — otherwise a too-eager gate would make the dialog permanently sticky and nothing would catch it.

Worth noting `ConfirmDialog` in `ui-react` has the same gap (`isBusy` disables its own two buttons but not the header's X), so the agent and session delete dialogs can be closed mid-delete too. Out of scope here, but I can follow up.

plugins/agent-platform-backend/src/KagentClient.ts
marians: Agreed, and the "every installation takes this branch today" point is what makes it bite — this would have been a permanent Sentry issue per case, not a rare one.

Fixed in 4b07fec. All three are now `ConflictError` (409):

- **no agent on the session** — after the read-back, since `agent_id` is nullable and kagent needs an `agent_ref` to accept the upsert;
- **409 from the upsert** — sandbox-workload agent that already holds a session, via a new `conflict` opt-in on `request`, placed above the generic `!response.ok` branch alongside the existing `badRequest` one;
- **400 from the upsert** — kagent could not resolve the agent the session itself names, mapped at the call site.

Tests: `reports a session with no agent as a conflict, not a server error`, plus an `it.each` over the 400 and 409 upsert failures asserting `ConflictError`. The `does not enter the fallback on a %s` case keeps 500 → `UpstreamError`, so a genuinely unwell kagent still reaches Sentry.

plugins/agent-platform/src/apis/KagentApiClient.ts
marians: Good catch — I had reused `throwIfNotOk` without noticing it encodes an assumption that only holds for the reads.

Fixed in 4b07fec by making the mapping opt-out: `throwIfNotOk(response, { badRequestIsMissing })`, defaulting to the reads' behaviour so nothing else changes, with `renameSession` passing `false`. Discriminating on the backend error's own name wouldn't have worked — the unknown-installation case is an `InputError` too (`resolveInstallation` throws one), so both 400s serialize under the same name.

Two tests, one per direction: `does not label a rejected name as a missing kagent` and `still labels a read's 400 as a missing kagent`. The second matters because the reads' mapping is load-bearing in two places — the retry predicate in `QueryClientProvider` and the sessions provider's "kagent isn't deployed here" classification — and silently losing it would be worse than the bug being fixed.

plugins/agent-platform/src/components/SessionDetailPage/SessionRenameDialog.tsx
marians: Right — I wrote the effect for the open transition and then depended on `title`, which makes it fire on every upstream change too.

Fixed in 4b07fec with a `wasOpen` ref, re-seeding only on the `false → true` flip. `title` stays in the dependency list (it is read inside, and dropping it would just be an exhaustive-deps lie) but no longer triggers a reset on its own.

New test `does not overwrite an edit in progress when the title changes upstream`, and the existing `forgets an abandoned edit when reopened` still passes, so both halves are pinned. The test drives the title change through a captured setter rather than a button — the dialog is modal, so anything rendered beside it is unreachable from the DOM while it is open, which is exactly the situation being reproduced.

plugins/agent-platform-backend/src/KagentClient.ts
marians: You're right that my reasoning ran backwards — I inferred "the column is null" from "the read omits it", when the omission is the *consequence* of it being null.

Checked the v0.9.9 struct, and the good news is the conclusion was accidentally safe:

```go
// go/api/database/models.go:76 @ v0.9.9
Source *SessionSource `json:"source,omitempty"`
```

So v0.9.9 *does* serialize `source` when it is set. Its absence from our fixtures means those sessions have it null, not that the version cannot report it — the wipe was a latent risk (a value we failed to parse or forgot to forward), not a certainty.

Either way, taken your suggestion: the fallback now reads the session back and echoes kagent's own `agent_id` and `source` from that read, so the question is closed by construction rather than by an argument about serialization. The route no longer accepts either field from the caller. Test `echoes the session's own source, so the upsert cannot blank it` covers it, and the comment now states the actual reason rather than the inverted one.

PR #2077, opened 2026-08-10
feat(app): apply Giant Swarm logo and Roboto typeface

Rotfuks:
### What does this PR do?

Gives the app Giant Swarm's identity in place of Backstage's.

- **Logo** — the horizontal lockup from [`giantswarm/brand`](https://github.com/giantswarm/brand) (`logo/standard`), vendored as React components in `packages/app/src/assets/logo/`. It renders its wordmark in `currentColor`, so a single asset works on both the dark sidebar and light content surfaces. The collapsed sidebar rail uses the single-color ant mark, which `GiantSwarmIcon` in the app's icon bundle now shares instead of duplicating the paths.
- **Favicons and app icons** — regenerated from the full-color ant on a Giant Swarm Blue tile, scaled to stay legible at 16px. `theme-color`, `mask-icon` and `manifest.json` updated to match.
- **Typography** — Roboto, self-hosted via `@fontsource/roboto`, because the app's CSP has no `font-src` allowlist for external hosts. It is set in two places, since the component libraries are themed independently: `fontFamily` in `createUnifiedTheme` covers Material UI and `core-components`, and `--bui-font-regular` in `bui-overrides.css` covers `@backstage/ui`, which reads none of the Material UI palette and otherwise falls back to bare `system-ui`.

The white-label path is preserved throughout: every logo component keeps its `useBranding()` custom-asset branch first, so a mounted branding ConfigMap still overrides the Giant Swarm defaults. `app.branding.theme` also gains `secondaryTextColor`, `linkColor`, `accentColor`, `accentTextColor`, `border1` and `border2`, closing gaps in the override surface.

The portal shows the Giant Swarm logo instead of the Backstage logo — in the sidebar (expanded and collapsed), on the home page, and in the browser tab. Text renders in Roboto instead of the system font. Layout, density and colors are unchanged.

Customer portals that mount their own branding assets are unaffected — their assets still win.

### How does it look like?

Verified in the running app against a live backend, in both light and dark themes:

- Sidebar shows the Giant Swarm lockup expanded and the ant mark collapsed.
- Home page shows the lockup at 96px; the wordmark follows `text.primary`, so it is white on dark and near-black on light, while the ant keeps its brand gradient.

Favicon corner pixels were verified programmatically as `#002645` — `qlmanage` rasterises onto an opaque white background, so a rounded tile bakes white corners into every size.

Yes — done in this PR. `docs/branding.md` gains a **Default branding** section covering the logo, the favicon generation and the typography wiring, and its config reference is extended with the six new keys. `README.md` and `docs/ui.md` link to it.

No changeset: this only touches `packages/app` (private, not published to npm) and docs. A `CHANGELOG.md` entry is included instead.


marians:
Superseded by https://github.com/giantswarm/backstage/pull/2081

PR #2081, opened 2026-08-11
feat(app): ship the Giant Swarm logo as the default branding

marians:
### What does this PR do?

Makes the Giant Swarm logo the app's default branding, replacing the Backstage defaults:

- **Expanded sidebar and home page** — the horizontal lockup from [`giantswarm/brand`](https://github.com/giantswarm/brand) (`logo/standard`), vendored as React components under `packages/app/src/assets/logo/`.
- **Collapsed sidebar rail** — the ant on its own, cropped from the *same* artwork as the lockup so it does not move when the sidebar expands.
- **Favicons and app icons** — the full-color ant on a full-bleed Giant Swarm Blue (`#002645`) tile, plus the Safari pinned-tab mask in Giant Swarm Orange.

Two details worth a reviewer's attention:

- `GiantSwarmAntPaths` (`GiantSwarmAnt.tsx`) is shared by the lockup and the standalone mark. `ANT_VIEW_BOX` deliberately keeps the lockup's vertical framing rather than cropping tight to the ant — both forms render at a fixed height, so a shared viewBox height means a shared scale and the ant lands on identical pixels in both states. Cropping it tighter reintroduces a visible ~1px jump on expand/collapse; there is a test locking the invariant.
- `GiantSwarmIconPaths` stays a separate single-color ant in `currentColor`. That one is the *UI glyph* (auth provider list in settings, the app's icon bundle), not the logo, so it should keep following the surrounding text color.

Deployments that mount their own branding assets are unaffected — custom assets still take precedence, and the override list is unchanged.

Two page-metadata fixes came along, because both shipped Giant-Swarm-or-Backstage-specific text to *every* deployment:

- `<meta name="description">` was hardcoded; it now comes from a new `app.description` config key, matching how `<title>` already reads `app.title`.
- `manifest.json`'s PWA `name`/`short_name` said `Backstage` on `main` and would have said `Giant Swarm Developer Portal`; they are now generic, consistent with `app.title: Dev Portal` in the base config. The manifest is a static asset and cannot read app config, so a deployment needing its own PWA identity has to ship a patched file — noted in the docs. This also fixes `favicon.ico` being declared as `type: image/png`.

The portal now carries Giant Swarm's identity instead of the Backstage default logo and favicon — in the sidebar (expanded and collapsed), on the home page, in the browser tab, and as the home-screen/PWA icon. Customer portals that supply their own branding assets look exactly as before.

### How does it look like?

Screenshots of the collapsed and hover-expanded sidebar are attached below.

Verified in the running app: the ant renders at an identical position and size in both sidebar states (`x 24.96, y 34.34, 26.7 × 23.8` in each), the two simultaneously-mounted lockups get distinct gradient ids that each resolve within their own SVG, and `app.description` reaches the served page end-to-end.

Supersedes #2077, which bundled this together with a Roboto typeface and a set of `app.branding.theme` color additions. This PR keeps only the logo/icon work; the typeface is dropped, and the color-config work is left for a separate PR after some review findings are resolved (`accentColor` maps to `--bui-accent-bg`, which only `Switch` reads, while every primary action reads `--bui-bg-solid`; and `linkColor` maps to `--bui-ring`, the focus ring, which no link reads at rest).

`theme.palette.navigation.color` was considered for the sidebar wordmark and deliberately rejected: it is `#b5b5b5` in both Backstage default palettes — the muted color for *unselected* nav labels — so it renders the logo in inactive-menu-item gray, and recoloring the lockup is against the styleguide anyway. The wordmark is white on the dark sidebar, and a deployment with a light `navigation.background` should supply its own `logo-full.svg`.

Yes — done in this PR. `docs/branding.md` gains a "Default branding" section covering the logo components, the icon glyph, and the deployment-specific text keys, and `app.description` is in the configuration reference.

- [x] `CHANGELOG.md` updated under `### Changed`.

No changeset needed: this only touches `packages/app`, which is released as the image + Helm chart from the conventional-commit PR title, not published to npm.


packages/app/public/index.html
```diff
@@ -3,10 +3,10 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <meta name="theme-color" content="#000000" />
+    <meta name="theme-color" content="#002645" />
     <meta
       name="description"
-      content="Backstage is an open source framework for building developer portals"
+      content="<%= config.getOptionalString('app.description') ?? 'A developer portal built on Backstage' %>"
```
marians: **medium — `app.description` is interpolated unescaped into an HTML attribute**

`<%= %>` is lodash's *unescaped* interpolate delimiter, and here it lands inside a double-quoted attribute. `injectConfigIntoHtml` (`@backstage/plugin-app-backend`) compiles this template with `{ interpolate: /<%=([\s\S]+?)%>/g }` and no escaping, so:

```yaml
app:
  description: 'The "Acme" portal'
```

renders

```html
content="The "Acme" portal"
```

The attribute truncates at `The` and the rest becomes stray attributes on the `<meta>`. A value containing `"><script>` injects markup into `<head>` at serve time.

Since only `interpolate` is overridden, lodash's default *escape* delimiter still works — `<%- ... %>` fixes it:

```html
content="<%- config.getOptionalString('app.description') ?? 'A developer portal built on Backstage' %>"
```

The `app.title` interpolation further down has the same class of exposure, but in text context a bare quote is harmless, whereas descriptions routinely contain quotes.

packages/app/src/modules/nav/LogoFull.tsx
```diff
@@ -1,19 +1,19 @@
...
 import { useBranding } from '../branding';
+import { GiantSwarmLogoFull } from '../../assets/logo';
 
 const useStyles = makeStyles({
   svg: {
     width: 'auto',
-    height: 30,
+    height: 26,
+    // The lockup's wordmark is `currentColor`; the ant keeps its brand gradient.
+    color: '#ffffff',
```
marians, line 11: **low — hardcoded white wordmark ignores a configurable sidebar background**

`app.branding.theme.<variant>.navigation.background` is a supported config key (declared in `packages/app/config.d.ts`), so a deployment can legitimately set a light sidebar. With the wordmark pinned to `#ffffff` it then renders white-on-light and disappears entirely — only the ant survives. Before this PR the whole logo was `#7df3e1`, which at least stayed visible on both.

`docs/branding.md` acknowledges this and tells such deployments to ship their own `logo-full.svg`, so it's documented rather than unknown — but deriving the color from `theme.palette.navigation.background` (e.g. via MUI's `getContrastText`) would turn a silent failure into a correct default.

packages/app/public/index.html
```diff
@@ -39,7 +39,7 @@
     <link
       rel="mask-icon"
       href="<%= publicPath %>/safari-pinned-tab.svg"
-      color="#5bbad5"
+      color="#e86d00"
```
marians, line 50: **low — the mask-icon `color` is not part of the branding override path**

`BrandingFavicon` rewrites only the `href` of `link[rel="mask-icon"]`, never `color`. So a customer portal that mounts its own `safari-pinned-tab.svg` still gets its mask filled in Giant Swarm Orange in Safari's pinned tabs — which contradicts the PR body's "custom assets still take precedence".

Either extend `BrandingFavicon` to set `color` from the branding theme, or pick a neutral value here.

packages/app/public/index.html
```diff
@@ -3,10 +3,10 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <meta name="theme-color" content="#000000" />
+    <meta name="theme-color" content="#002645" />
```
marians, line 6: **low — `theme-color` hardcodes Giant Swarm Blue with no override path**

Same shape as the `mask-icon` `color` below: `BrandingFavicon` doesn't touch this meta, so mobile browser chrome on *every* customer portal is now Giant Swarm Blue. The previous `#000000` was at least brand-neutral.

Consider reading it from the branding theme in `BrandingFavicon`, or reverting to a neutral value.

packages/app/public/manifest.json
```diff
@@ -1,15 +1,25 @@
...
+      "sizes": "192x192",
       "type": "image/png"
     }
   ],
   "start_url": "./index.html",
   "display": "standalone",
-  "theme_color": "#000000",
-  "background_color": "#ffffff"
+  "theme_color": "#002645",
+  "background_color": "#002645"
```
marians, line 24: **low — the PWA manifest is outside the branding override path entirely**

The manifest now points at Giant-Swarm-branded icons and sets `theme_color`/`background_color: #002645`. The browser fetches `manifest.json` directly, so `BrandingFavicon` can't rewrite any of it — a customer portal installed as a PWA on Android or desktop shows the Giant Swarm ant and a blue splash screen.

`docs/branding.md` notes the `name`/`short_name` limitation but not the icons or colors, and the PR body's "look exactly as before" isn't accurate for installed portals. Worth either extending the docs' limitation note to cover icons + colors, or serving the manifest through a route that applies branding config.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Package specific changes (for packages from `packages/*` and `plugins/*`) can be
...
 - AI chat: let users ask the AI to explain Flux error messages. The Flux resource card's "Troubleshoot with AI" button now sends the actual failing condition message (instead of asking the AI to look the resource up), and the HelmRelease conditions card on the deployment details page gets an "Explain this error" button on failing conditions. The buttons render nothing on installations without ai-chat enabled (#1935).
 
+### Changed
+
+- Branding: the app now ships Giant Swarm's logo as its default identity — the horizontal lockup in the expanded sidebar and on the home page, and the ant mark in the collapsed rail, the favicons and the app icons. Deployments that mount their own branding assets are unaffected: custom assets still take precedence. The page description now comes from the new `app.description` config key instead of being hardcoded, and `manifest.json`'s PWA name is generic so no deployment ships another company's name (#2077).
```
marians: **low — links the superseded PR**

The entry ends with `(#2077)`, which is the superseded PR, so the changelog points readers at a closed PR with no merged changes. Should be `(#2081)`.

packages/app/src/assets/icons/CustomIcons.tsx
```diff
@@ -1,28 +1,9 @@
 import { SvgIcon, SvgIconProps } from '@material-ui/core';
+import { GiantSwarmIconPaths } from '../logo/GiantSwarmMark';
```
marians: **low — deep import instead of the new barrel**

This reaches into `'../logo/GiantSwarmMark'` rather than the `'../logo'` barrel added in this PR, against the documented convention ("Always import from directory").

It also reads oddly: the UI glyph isn't part of `GiantSwarmMark` conceptually, so a reader following the import lands in a file about the logo mark. Importing from `'../logo'` — or splitting `GiantSwarmIconPaths` into its own module under `logo/` — keeps the boundary the docs describe.

marians (commented):
## Code review (medium effort)

7 findings inline, plus one below that can't be anchored (binary file). One `medium`, the rest `low`.

### `packages/app/public/favicon.ico` — low: the new `.ico` dropped its 32px and 16px entries

The regenerated file contains a single 48×48 (PNG-encoded) entry; the previous one carried three (48, 32, 16). Anywhere the `.ico` is consumed at 16px — bookmark lists, Windows pinned sites, clients that prefer `rel="shortcut icon"` over the sized PNGs — the 48px artwork now gets downscaled by the browser. That's exactly the legibility case `docs/branding.md` says needed a hand-tuned 92% scale at 16px, so the tuning is bypassed on those surfaces. Regenerating the `.ico` as a multi-size container from the existing `favicon-16x16.png` / `favicon-32x32.png` preserves it.

### Verified, not assumed

The SVG work checks out numerically:

- **No-jump invariant holds.** The ant's real bbox in the lockup's user space is x 99.5–169.2, y 95.0–156.9, so `ANT_VIEW_BOX = '97.5 93 73.7 67'` clips nothing and yields identical padding (2.48 units) and identical scale in both sidebar states.
- **Nothing is cut off.** The wordmark bbox ends at x≈450 against a viewBox right edge of 453.5.
- **Gradient parity is real.** The gradient's `userSpaceOnUse` + `gradientTransform` maps to exactly y 95.29–156.74, so the standalone mark and the lockup get the same gradient.
- **All four new PNGs** have `#002645` in all four corners.
- **The new tests should pass as written** — jsdom matches the case-sensitive `linearGradient` selector, and `stopColor` renders as `stop-color`.

Only the `index.html` interpolation is worth fixing before merge. The rest are branding-override gaps that need either a docs correction or an explicit decision that Giant Swarm defaults are acceptable for customer portals.

packages/app/public/index.html
marians: Fixed in 9e739a3a. Confirmed the mechanism against the installed code rather than assuming it: `injectConfigIntoHtml` does compile with `{ interpolate: /<%=([\s\S]+?)%>/g }` and no `escape` option, and lodash fills unspecified delimiters from `templateSettings`, so `<%- %>` is available. Verified both directions:

```
input:      The "Acme" portal"><script>alert(1)</script>
with <%= :  content="The "Acme" portal"><script>alert(1)</script>"
with <%- :  content="The &quot;Acme&quot; portal&quot;&gt;&lt;script&gt;alert(1)&lt;/script&gt;"
```

Then end-to-end with that value actually in `app-config.local.yaml`: the attribute stays intact, the `<meta>` carries exactly 2 attributes (no stray ones), and nothing executes.

I also switched `app.title`, and it's worth noting the exposure there is more than quotes: `<title>` is RCDATA, so a value containing `</title><script>` breaks out and executes. Confirmed it does with `<%=` and doesn't with `<%-`. Escaping is transparent for ordinary values — a title with an apostrophe serves as `&#39;` and still renders as `'`.

Left `<%= publicPath %>` alone deliberately: it's `new URL(app.baseUrl, …).pathname`, which percent-encodes quotes and angle brackets (`/a%22%3E%3Cscript%3E…`), so it can't break out — and the bundler emits its own `<%= publicPath %>` lines into the template, so changing ours would diverge from upstream.

One trap worth recording, since it cost a build: `html-webpack-plugin`'s loader lodash-compiles this file at build time, so *any* literal delimiter in it is compiled — including inside HTML comments. My first attempt documented the fix in a comment containing a literal empty `<%= %>`, which compiles to `((__t = ( )) == null …)` and fails the build with `SyntaxError: Unexpected token ')'`. The comment now describes the delimiters in prose and warns about this.

Unrelated to this file, but found while testing: in **dev only**, config strings are injected raw into the `<script type="backstage.io/config">` block, so a hostile `app.description` executes there. Production is not affected — `injectConfigIntoHtml` strips `</script` from that JSON. It's upstream behaviour on a value the operator authors, so out of scope here, but noting it.

CHANGELOG.md
marians: Fixed in 32def5c9 — now `(#2081)`.

packages/app/src/assets/icons/CustomIcons.tsx
marians: Fixed in 32def5c9 — imports from `'../logo'` now.

Took the barrel option rather than splitting `GiantSwarmIconPaths` into its own module. The doc comments on both sides already draw the boundary you're pointing at (the glyph is documented as the UI icon, the mark as the logo), so the extra file didn't seem to buy much beyond the corrected import. Happy to split it if you'd rather the file layout carried that distinction too.

marians (commented):
Fixed in 32def5c9. `favicon.ico` is a 48/32/16 container again, rebuilt from the existing `favicon-32x32.png` and `favicon-16x16.png` plus the 48px entry extracted from the previous file — so all three sizes are the hand-tuned artwork, and the 92% scaling is no longer bypassed at 16px.

```
container: type=1 (1=icon), reserved=0, entries=3
  entry 0: declared 48x48 32bpp PNG 3840B | actual 48x48 | all corners #002645 ✓
  entry 1: declared 32x32 32bpp PNG 2484B | actual 32x32 | all corners #002645 ✓
  entry 2: declared 16x16 32bpp PNG 1289B | actual 16x16 | all corners #002645 ✓
```

Each entry is verified to decode at the size its directory entry declares, and `file` agrees: `MS Windows icon resource - 3 icons`. The served file from the dev server is the 3-entry version.

Entries are PNG-encoded rather than BMP/DIB like the pre-PR file. That's supported by every browser and Windows Vista onward, and the 48px entry this PR shipped was already PNG, so it keeps the file smaller (7,667 bytes vs 15,086 for main's BMP version).

Worth noting for the record: `https://www.giantswarm.io/favicon.ico` is not a usable source here — it responds `Content-Type: image/png` with a single 64×64 PNG, so it carries no embedded resolutions either. Everything needed was already in the branch.

